PR #4, opened 2016-08-17
Consolidate props `blockedDates`, `unblockedDates`, `blockedByDefault` into prop `isDayBlocked`

Corei13:


majapw:
I think it's a great idea to accept a function for blockedDates but (1) I think it's maybe a bit problematic to overload the prop in this way and (2) having a method kind of makes the `blockedDates`, `unblockedDates`, `blockedByDefault` props unnecessary as you can just pass in `(day) => blockedDatesArray.contains(day)` or `(day) => !blockedDatesArray.contains(day)`. 

It might be better to consolidate this all into one prop, a la `isDayBlocked: PropTypes.func`. What do you think?

Corei13:
@majapw Agreed, that was my first intention actually. But I didn't want to make any breaking changes. I'll go ahead and consolidate the props into one.

stories/DateRangePicker.js
```diff
@@ -83,12 +83,11 @@ storiesOf('DateRangePicker', module)
   ))
   .add('with some blocked dates', () => (
     <DateRangePickerWrapper
-      blockedDates={datesList}
+      isDayBlocked={day1 => datesList.some(day2 => day2.isSame(day1, 'day'))}
```
majapw: We could use the `isSameDay` util method here, but that's somewhat of a nit

majapw:
This looks great. Happy to merge once it is rebased and commits are squashed. I'd like to get this out in this major version bump. :)

majapw:
LGTM! Thanks!

PR #7, opened 2016-08-18
Adds a travis file

goatslacker:
Let's add a travis file so we can get CI running on each PR.

@majapw @ljharb


.travis.yml
```diff
@@ -0,0 +1,18 @@
+language: node_js
+node_js:
+  - "5.1"
```
ljharb: we should be testing node 6 as well

.travis.yml
```diff
@@ -0,0 +1,18 @@
...
+  - "5.0"
+  - "4.2"
```
ljharb: and also latest node 4

.travis.yml
```diff
@@ -0,0 +1,18 @@
...
+  - "4.1"
+  - "4.0"
+script: npm run lint && npm test
+env:
+  - REACT=0.14
```
ljharb: since our dev deps have react 15, we should be testing on that also

PR #8, opened 2016-08-18
Include a link to gh-pages in README

aray12:
Since storybook is already deployed to github pages, might as well provide a link to that so potential users can find a live demo without cloning the repo. Might recommend putting it up higher. For me, the first thing I love for is a live demo. But, with current README structure, this made the most sense. 

BTW really happy you guys OS'ed this. Very good chance I'll be using it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -127,6 +127,10 @@ The `phrases` prop is an object that contains all the English language phrases c
 ## Live Playground
 For examples of the datepicker in action,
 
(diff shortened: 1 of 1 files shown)


majapw:
We'll be adding more to the README in terms of screenshots, the other components' APIs and whatnot like tonight, so we can hoist this up when we do that. Thanks for this!

PR #10, opened 2016-08-18
update scss variables to be overridden

travisbloom:
PR for https://github.com/airbnb/react-dates/issues/11
set all SCSS variables to !default to enable easy theming.

### DIFF
--- a/README.md
+++ b/README.md
@@ -149,3 +149,14 @@ The `phrases` prop is an object that contains all the English language phrases c
     clearDates: PropTypes.node,
   }),
 ```
+
+## Theming
+
+react-dates comes with a set of SCSS variables that can be overridden to add your own project-specific theming. Override any variables found in `css/variables.scss` with your own and then import `~react-dates/css/styles.scss` (and `~react-dates/css/variables.scss` if you're only overriding a few). If you were using [sass-loader](https://github.com/jtangelder/sass-loader) with webpack, the code below would properly override the selected variables:
+```sass
+//overriding default sass variables with my project's colors
+$react-dates-color-primary: $some-color-specific-to-my-project;
+$react-dates-color-primary-dark: $some-other-color-specific-to-my-project;
+@import '~react-dates/css/variables';
+@import '~react-dates/css/styles.scss';
+```
--- a/css/variables.scss
+++ b/css/variables.scss
@@ -1,32 +1,23 @@
-$input-width: 130px;
+$react-dates-width-input: 130px !default;
+$react-dates-width-day-picker: 300px !default;
 
-$color-babu: #00a699;
-$color-babu-dark: #00514a;
-$color-white: #fff;
+$react-dates-color-primary: #00a699 !default;
+$react-dates-color-primary-dark: #00514a !default;
+$react-dates-color-primary-shade-1: #33dacd !default;
+$react-dates-color-primary-shade-2: #66e2da !default;
+$react-dates-color-primary-shade-3: #80e8e0 !default;
+$react-dates-color-primary-shade-4: #b2f1ec !default;
+$react-dates-color-secondary: #007a87 !default;
+$react-dates-color-white: #fff !default;
+$react-dates-color-gray: #565a5c !default;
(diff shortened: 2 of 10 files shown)


majapw:
@lencioni can you take a look at this?

ljharb:
Can you elaborate a bit more on how "!default" works?

travisbloom:
Sure, see this article: https://robots.thoughtbot.com/sass-default

From the SASS docs:

> You can assign to variables if they aren’t already assigned by adding the !default flag to the end of the value. This means that if the variable has already been assigned to, it won’t be re-assigned, but if it doesn’t have a value yet, it will be given one.

travisbloom:
basically I want to do something like this in my teams project:

```
$color-babu: $some-color-specific-to-my-company;
@import '~react-dates/css/styles';
```

lencioni:
The `!default` makes sense. We should probably also namespace these variables better.

README.md
```diff
@@ -124,6 +124,9 @@ The `phrases` prop is an object that contains all the English language phrases c
   }),
 ```
 
+**Theming**
+react-dates comes with a set of scss variables that can be overridden to add your own project-specific theming. Override any variables found in `css/variables.scss` with your own and then import `~react-dates/css/styles.scss`.
```
lencioni: nit: scss -> SCSS. Also, it would be nice to provide a full example here to help people understand how to do this.

travisbloom:
@lencioni if you like i can modify all the vars from `$color-babu` to `$react-dates-color-babu`? That could be considered a breaking change tho?

README.md
travisbloom: yup, will expand

lencioni:
@travisbloom I think that makes sense, but I'd like @majapw to make the call.

majapw:
`babu` as the name of the color is not super meaningful in term of what its function is in the repo. I think these color names could be cleaned up to be more semantic as to what they actually do here, but I think prefixing them with `$react-dates-` is a good first step. 👍 

Let's do that and write up a clearer explanation of how to override the `!default` and I think that would be good.

css/variables.scss
```diff
@@ -1,32 +1,32 @@
...
+$color-babu: #00a699 !default;
+$color-babu-dark: #00514a !default;
+$color-white: #fff !default;
 
-$border-color: #dbdbdb;
-$placeholder-color: #757575;
-$text-color: #484848;
-$focus-color: #99ede6;
-$focus-text-color: #007a87;
+$border-color: #dbdbdb !default;
```
lencioni: If we are renaming variables, it would also be nice to apply a consistent naming scheme. I like names that go from most general on the left to most specific on the right. e.g. `$namespace-type-name` e.g. `$react-dates-color-babu` is good. In this case, that would look like `$react-dates-color-border`, `$react-dates-color-placeholder`, `$react-dates-color-text-focus`, `$react-dates-border-chevron` etc.

lencioni:
Yeah, instead of words like "babu", maybe words like "primary", "secondary", etc would work?

travisbloom:
@majapw @lencioni updated the PR with a first pass at having more generic, namespaced sass variables. Figured renaming css classes would go under a separate PR.

Let me know your thoughts.

lencioni:
LGTM

README.md
```diff
@@ -140,6 +140,16 @@ The `phrases` prop is an object that contains all the English language phrases c
   }),
 ```
 
+**Theming:**
```
majapw: Sorry! One thing on the README... this doesn't really fall under the `## API` category imo as that related more to the actual props that you can pass into the components. I think we're going to be doing a lot of improvements on Theming and the styles in general in the future, so maybe we could have this as its own top-level section in the README?

majapw:
This looks great! Thanks @travisbloom!

README.md
travisbloom: Oops, good catch. updated!

majapw:
Wonderful! Is there any way you could squash this into 1 or 2 commits before we merge in? Thank you!

travisbloom:
@majapw squashed and rebased :)

PR #12, opened 2016-08-18
Add prop `isOutsideRange` and remove prop `allowPastDates`

Corei13:


majapw:
QQ: would you not be able to do this with the `isDayBlocked` prop in https://github.com/airbnb/react-dates/pull/4? Are you trying to style `out-of-range` differently from `calendar-blocked`? I guess I'm looking for what your usecase is for this change. 

Thanks!

Corei13:
> Are you trying to style `out-of-range` differently from `calendar-blocked`?
@majapw Yes. Apart from that, there's also a small change in the behaviour. If no day of the current month is within the allowed range, the date picker shows the closest month with at least one allowed day. This behaviour is not possible by `isDayBlocked`.

jptaylor:
+1

majapw:
Sorry for the delayed response! Um, so I think that `minimumAllowedDate` and `maximumAllowedDate` make sense as props, but I'd like to decouple it from what month is visible when the datepicker opens a la discussion on https://github.com/airbnb/react-dates/issues/17. Would it be possible to get that out there so that we can merge in the props and cut v2.0.0 to include this change?

Corei13:
@majapw So the current month will be visible by default like now?

majapw:
@Corei13 That would be my thought. We can then we can add the visible month prop as a non-breaking change in a separate PR. Would that be okay?

Corei13:
@majapw Cool! I'll push the changes.

Corei13:
@majapw done

ljharb:
What about making these props be functions that return a moment object, instead of a moment object?

That would allow you to customize the allowed dates if, say, the clock went from 11:59PM to 12:01AM and you wanted the allowable dates to follow suit.

Corei13:
@ljharb: Good idea. Also, if we go for a function, these two props can be merged into one like `isOutsideRange: day => boolean`. What do you think?
cc @majapw

ljharb:
The downside of `isOutsideRange` is that you have to do the moment comparison yourself. I'm not sure which would be better.

majapw:
I kind of think that's fine. We already expect a lot of moment comparison from users of the API so this is probably also okay. 

Let's go with the `isOutsideRange: day => boolean`.

Corei13:
On second thought, if we use `isOutSideRange`, it won't be possible later to show initial month based on actual minimum and maximum allowed dates.

majapw:
@Corei13 As discussed before, I think we want the initial visible month to be controllable by a prop so that parent would have the responsibility to do that, not the `DateRangePicker` component itself. Since the parent component ostensibly has the information to create the `isOutsideRange` method, it should also have the information to figure out what the appropriate first visible month is.

src/components/DateRangePicker.jsx
```diff
@@ -228,9 +228,9 @@ export default class DateRangePicker extends React.Component {
 
   doesNotMeetMinimumNights(day) {
     const { startDate, focusedInput, minimumNights } = this.props;
-    const today = moment().startOf('day');
-    const dayDiff = startDate ? day.diff(startDate, 'days') : day.diff(today, 'days');
+    if (!startDate) return false;
 
+    const dayDiff = day.diff(startDate, 'days');
```
majapw: Will this work when `startDate` is `null`? What's the behavior then?

Also this is an interesting side-effect that I had not thought about.

majapw:
@ljharb can you take a look at this PR as well?

One thing of note is that this changes the `doesNotMeetMinimumNights` behavior. Namely we can't block out dates that wouldn't meet minimum nights if we don't have have a start date selected... (ie today cannot be an end date). Maybe this is bad. Thoughts?

src/components/DateRangePicker.jsx
ljharb: I think the previous behavior of defaulting to `today` makes sense here.

majapw: wait actually, let's just do this:
```
doesNotMeetMinimumNights(day) {
  const { startDate, focusedInput, minimumNights } = this.props;
  if (focusedInput !== END_DATE) return false;

  let doesNotMeetMinimumNights;
  if (startDate) {
    const dayDiff = day.diff(startDate, 'days');
    doesNotMeetMinimumNights = dayDiff < minimumNights && dayDiff >= 0;
  } else {
    doesNotMeetMinimumNights = this.isBlocked(moment(day).subtract(minimumNights, 'days'));
  }

  return doesNotMeetMinimumNights;
}
```

Would that work?

ljharb: that seems reasonable

majapw:
This looks good once the `doesNotMeetMinimumNights` issue is fixed!

src/components/DateRangePicker.jsx
Corei13: `doesNotMeetMinimumNights = this.isBlocked(moment(day).subtract(minimumNights, 'days'));`

@majapw did you mean `doesNotMeetMinimumNights = this.props.isOutsideRange(moment(day).subtract(minimumNights, 'days'));`? Because `moment(day).subtract(minimumNights, 'days')` can be blocked for other reasons, in those case it is possible to select `day` as end date and some other non-blocked day before `moment(day).subtract(minimumNights, 'days')` as start date.

majapw: Mmm fair point. Given that we allow blocked days inside ranges right now,
what you're saying is true. So yes, probably we'd want
`this.props.isOutsideRange` there.

src/components/DateRangePicker.jsx
```diff
@@ -227,11 +227,14 @@ export default class DateRangePicker extends React.Component {
...
+    const { startDate, isOutsideRange, focusedInput, minimumNights } = this.props;
+    if (focusedInput !== END_DATE) return false;
 
-    const dayDiff = day.diff(startDate, 'days');
-    return focusedInput === END_DATE && dayDiff < minimumNights && dayDiff >= 0;
+    if (startDate) {
+      const dayDiff = day.diff(startDate, 'days');
+      return dayDiff < minimumNights && dayDiff >= 0;
+    }
+    return isOutsideRange(moment(day).subtract(minimumNights, 'days'));
```
majapw: Nice! You're better at our style guide than I am. :p

majapw:
Nice. I'm about to be on a flight, but assuming tests pass, I'll merge this in and cut a release when I laaand. Thanks so much!

README.md
```diff
@@ -84,9 +84,10 @@ To indicate which days are blocked from selection, you may provide a function to
   isDayBlocked: PropTypes.func
 ```
 
-If you would like to allow the user to selected days in the past, you may set `allowPastDates` to true.
+`isOutsideRange` indicates which days are out of selectable range.
+Past dates out of range by default. If you would like to allow the user to select days in the past, you may set `isOutsideRange` to `() => true`.
```
majapw: Might be good to add a comment about the relationship between `isOutsideOfRange` and `doesNotMeetMinimumNights`. This is not blocking and I can add this after.

test/components/DateRangePicker_spec.jsx
```diff
@@ -837,25 +861,29 @@ describe('DateRangePicker', () => {
...
                 minimumNights={MIN_NIGHTS}
+                isOutsideRange={isOutsideRange}
               />
             );
             expect(wrapper.instance().doesNotMeetMinimumNights(testDate)).to.equal(true);
           });
 
-          it('returns false if arg is > props.minimumNights after today', () => {
-            const testDate = moment(today).add(MIN_NIGHTS + 1, 'days');
+          it('returns true if arg - props.minimumNights is inside allowed range', () => {
```
majapw: This should say returns false.

README.md
Corei13: Not sure what to write here, can you please add that? I'm not good with words.

majapw: 👍 I'll do it in a follow up.

PR #16, opened 2016-08-18
Remove jQuery as a dependency

milesj:
This PR removes jQuery as a dependency and replaces the outer/inner width/height methods with a single function that handles all scenarios. Will need to do further testing in browser and within unit tests to verify the numbers are correct.

### DIFF
--- a/package.json
+++ b/package.json
@@ -86,7 +86,6 @@
     "react-portal": "^2.2.1"
   },
   "peerDependencies": {
-    "jquery": ">=2.1",
     "moment": ">=2.10",
     "react": ">=0.14",
     "react-dom": ">=0.14",
--- a/package.json
+++ b/package.json
@@ -60,7 +60,6 @@
     "git-directory-deploy": "^1.5.1",
     "imports-loader": "^0.6.5",
     "in-publish": "^2.0.0",
-    "jquery": "^3.1.0",
     "mocha": "^3.0.2",
     "mocha-wrap": "^2.0.5",
     "moment": "^2.14.1",
--- a/src/components/DayPicker.jsx
+++ b/src/components/DayPicker.jsx
@@ -2,7 +2,6 @@ import React, { PropTypes } from 'react';
 import shallowCompare from 'react-addons-shallow-compare';
 import ReactDOM from 'react-dom';
 import moment from 'moment';
-import $ from 'jquery';
 import cx from 'classnames';
 
 import OutsideClickHandler from './OutsideClickHandler';
@@ -22,9 +21,10 @@ import { HORIZONTAL_ORIENTATION, VERTICAL_ORIENTATION } from '../constants';
 const CALENDAR_MONTH_WIDTH = 300;
 const DAY_PICKER_PADDING = 9;
 const MONTH_PADDING = 23;
-
 const PREV_TRANSITION = 'prev';
 const NEXT_TRANSITION = 'next';
+
+const slice = Array.prototype.slice;
 
 const propTypes = {
   enableOutsideDays: PropTypes.bool,
@@ -104,14 +104,63 @@ export default class DayPicker extends React.Component {
   }
 
   getMonthHeightByIndex(i) {
-    const $transitionContainer = $(ReactDOM.findDOMNode(this.refs.transitionContainer));
-    return this.getMonthHeight($transitionContainer.find('.CalendarMonth').eq(i));
+    return this.getMonthHeight(
+      ReactDOM.findDOMNode(this.refs.transitionContainer).querySelectorAll('.CalendarMonth')[i]
+    );
   }
 
-  getMonthHeight($el) {
+  getMonthHeight(el) {
+    const caption = el.querySelector('.js-CalendarMonth__caption');
+    const grid = el.querySelector('.js-CalendarMonth__grid');
+
     // Need to separate out table children for FF
-    return $el.find('.js-CalendarMonth__caption').outerHeight(true) +
-      $el.find('.js-CalendarMonth__grid').height() + 1; // for border
+    return (
+      this.calculateDimension(caption, 'height', 'outer', true) +
+      this.calculateDimension(grid, 'height') +
+      1 // for border
+    );
+  }
+
+  applyTransformStyles(el, transform, opacity = '') {
+    const transformStyles = getTransformStyles(transform);
+    transformStyles.opacity = opacity;
+
+    Object.keys(transformStyles).forEach((styleKey) => {
+      // eslint-disable-next-line no-param-reassign
+      el.style[styleKey] = transformStyles[styleKey];
+    });
+  }
+
+  calculateDimension(el, axis, range = 'inner', withMargin = false) {
+    if (!el) {
+      return 0;
+    }
+
+    const style = window.getComputedStyle(el);
+    const axisStart = (axis === 'width') ? 'Left' : 'Top';
+    const axisEnd = (axis === 'width') ? 'Right' : 'Bottom';
+    const extraSize = (
+      parseFloat(style[`padding${axisStart}`]) +
+      parseFloat(style[`padding${axisEnd}`]) +
+      parseFloat(style[`border${axisStart}Width`]) +
+      parseFloat(style[`border${axisEnd}Width`])
+    );
+    let size = parseFloat(style[axis]);
+
+    if (range === 'outer' && style.boxSizing === 'content-box') {
+      size += extraSize;
+    } else if (range === 'inner' && style.boxSizing === 'border-box') {
+      size -= extraSize;
+    }
+
+    if (withMargin) {
+      size += (
+        parseFloat(style[`margin${axisStart}`]) +
+        parseFloat(style[`margin${axisEnd}`])
+      );
+    }
+
+    return size;
   }
 
   isHorizontal() {
@@ -123,8 +172,11 @@ export default class DayPicker extends React.Component {
   }
 
   initializeDayPickerWidth() {
-    const $calendarMonthGrid = $(ReactDOM.findDOMNode(this.refs.calendarMonthGrid));
-    this.dayPickerWidth = $calendarMonthGrid.find('.CalendarMonth').outerWidth();
+    this.dayPickerWidth = this.calculateDimension(
+      ReactDOM.findDOMNode(this.refs.calendarMonthGrid).querySelector('.CalendarMonth'),
+      'width',
+      'outer'
+    );
   }
 
   updateStateAfterMonthTransition() {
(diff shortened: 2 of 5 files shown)


src/components/DayPicker.jsx
```diff
@@ -22,10 +21,11 @@ import { HORIZONTAL_ORIENTATION, VERTICAL_ORIENTATION } from '../constants';
 const CALENDAR_MONTH_WIDTH = 300;
 const DAY_PICKER_PADDING = 9;
 const MONTH_PADDING = 23;
-
 const PREV_TRANSITION = 'prev';
 const NEXT_TRANSITION = 'next';
 
+const slice = Array.prototype.slice;
```
majapw: This is a really weird thing to pull out, I feel.

src/components/DayPicker.jsx
```diff
@@ -190,27 +242,31 @@ export default class DayPicker extends React.Component {
...
 
   adjustDayPickerHeight() {
-    const $transitionContainer = $(ReactDOM.findDOMNode(this.refs.transitionContainer));
-    const heights = $transitionContainer.find('.CalendarMonth[data-visible="true"]').get().map(
-      (month) => this.getMonthHeight($(month))
-    );
+    const transitionContainer = ReactDOM.findDOMNode(this.refs.transitionContainer);
+    const heights = [];
+
+    slice.call(transitionContainer.querySelectorAll('.CalendarMonth')).forEach((el) => {
```
majapw: I feel like this is not the most readable way of doing this?

src/components/DayPicker.jsx
```diff
@@ -2,7 +2,6 @@ import React, { PropTypes } from 'react';
 import shallowCompare from 'react-addons-shallow-compare';
 import ReactDOM from 'react-dom';
 import moment from 'moment';
-import $ from 'jquery';
```
kesne: I love you.

src/components/DayPicker.jsx
```diff
@@ -104,14 +104,63 @@ export default class DayPicker extends React.Component {
   }
 
   getMonthHeightByIndex(i) {
-    const $transitionContainer = $(ReactDOM.findDOMNode(this.refs.transitionContainer));
-    return this.getMonthHeight($transitionContainer.find('.CalendarMonth').eq(i));
+    return this.getMonthHeight(
+      ReactDOM.findDOMNode(this.refs.transitionContainer).querySelectorAll('.CalendarMonth')[i]
```
kesne: This looks like it's a dom ref, so you don't need to do findDOMNode.

src/components/DayPicker.jsx
```diff
@@ -104,14 +104,63 @@ export default class DayPicker extends React.Component {
...
+    const caption = el.querySelector('.js-CalendarMonth__caption');
+    const grid = el.querySelector('.js-CalendarMonth__grid');
+
     // Need to separate out table children for FF
-    return $el.find('.js-CalendarMonth__caption').outerHeight(true) +
-      $el.find('.js-CalendarMonth__grid').height() + 1; // for border
+    return (
+      this.calculateDimension(caption, 'height', 'outer', true) +
+      this.calculateDimension(grid, 'height') +
+      1 // for border
```
kesne: Comments should go before lines. https://github.com/airbnb/javascript#comments--singleline

src/components/DayPicker.jsx
kesne: You could use `Array.from` and then force people to polyfill.

kesne: Or pull a module to do this! 😀

majapw: I'm pretty against forcing people to polyfill. The whole point is for this to Just Work.

src/components/DayPicker.jsx
```diff
@@ -190,27 +242,31 @@ export default class DayPicker extends React.Component {
...
+      }
+    });
 
     const newMonthHeight = Math.max(...heights) + MONTH_PADDING;
-    if (newMonthHeight !== $transitionContainer.height()) {
+
+    if (newMonthHeight !== this.calculateDimension(transitionContainer, 'height')) {
       this.monthHeight = newMonthHeight;
-      $transitionContainer.css('height', newMonthHeight);
+      transitionContainer.style.height = newMonthHeight;
```
razh: I think this needs a `px`.

src/components/DayPicker.jsx
milesj: `Array.prototype.slice` is the way to do it. Thought I was going to be using it more, hence the reference at the top.

src/components/DayPicker.jsx
milesj: This is necessary for React 0.14.

src/components/DayPicker.jsx
milesj: Right, nice catch!

src/components/DayPicker.jsx
```diff
@@ -190,27 +240,33 @@ export default class DayPicker extends React.Component {
...
-    const $transitionContainer = $(ReactDOM.findDOMNode(this.refs.transitionContainer));
-    const heights = $transitionContainer.find('.CalendarMonth[data-visible="true"]').get().map(
-      (month) => this.getMonthHeight($(month))
-    );
+    const transitionContainer = ReactDOM.findDOMNode(this.refs.transitionContainer);
+    const calendarMonths = transitionContainer.querySelectorAll('.CalendarMonth');
+    const heights = [];
+
+    // convert node list to array
+    Array.prototype.slice.call(calendarMonths).forEach((el) => {
```
ljharb: `Array.prototype.forEach.call` - no need to create the intermediate array.

Alternatively, `[...calendarMonths].forEach` but the former is faster.

package.json
```diff
@@ -80,7 +81,6 @@
     "react-portal": "^2.2.1"
   },
   "peerDependencies": {
-    "jquery": ">=2.1",
```
ljharb: please also update the readme to remove jquery from the example generated `npm install` command

src/components/DayPicker.jsx
milesj: Totally forgot about spreading. Will give that a shot (since it will probably compile down to Array prototype anyways).

ljharb: spreading will compile down to effectively a slice + forEach - forEach.call will just be the forEach (so i'd recommend doing that)

milesj:
Some screenshots of the calculations in action. Compare the numbers in the console to the dimensions of the hovered element.

screen shot 2016-08-19 at 10 19 17 am
screen shot 2016-08-19 at 10 19 25 am
screen shot 2016-08-19 at 10 20 02 am

I'm not sure how to write unit tests for this as their is no DOM. There were no tests before for it anyways.

majapw:
Yeah of my TODOS is to enable dom testing at some point which would help.
You could stub some things but it might be somewhat meaningless.

Do the height animations still work?

majapw:
Also can we test cross-browser before we merge this in?

milesj:
I've tested it on Mac FF/Chrome/Safari with no issues. Trying to test IE and mobile.

package.json
```diff
@@ -70,6 +69,7 @@
     "react-addons-test-utils": "^15.2.1",
     "react-dom": "^15.2.1",
     "react-svg-loader": "^1.1.1",
+    "rimraf": "^2.5.4",
```
ianks: Why was this added?

Daniel15:
> Trying to test IE and mobile.

I can test in IE + Edge for you if you provide a public URL where I can hit your fork! :smile:

package.json
milesj: Because it's required to run `npm build`. It was missing.

milesj:
@Daniel15 I won't be using a public URL. Was just going to be testing locally.

src/components/DayPicker.jsx
kesne: Not quite: https://facebook.github.io/react/blog/2015/10/07/react-v0.14.html#dom-node-refs

milesj:
I tested in IE11 and Edge with no issues. Storybook doesn't work in IE9/10, so can't verify.

Also, note to self, update this code. (I can't because of Windows line endings).

```
  calculateDimension(el, axis, borderBox = false, withMargin = false) {
    if (!el) {
      return 0;
    }

    const style = window.getComputedStyle(el);
    const axisStart = (axis === 'width') ? 'Left' : 'Top';
    const axisEnd = (axis === 'width') ? 'Right' : 'Bottom';

    // Includes border and padding
    let size = (axis === 'width') ? el.offsetWidth : el.offsetHeight;

    // Get the inner size
    if (!borderBox) {
      size -= (
        parseFloat(style[`padding${axisStart}`]) +
        parseFloat(style[`padding${axisEnd}`]) +
        parseFloat(style[`border${axisStart}Width`]) +
        parseFloat(style[`border${axisEnd}Width`])
      );
    }

    // Apply margin
    if (withMargin) {
      size += (
        parseFloat(style[`margin${axisStart}`]) +
        parseFloat(style[`margin${axisEnd}`])
      );
    }

    return size;
  }
```

src/components/DayPicker.jsx
```diff
@@ -104,14 +102,63 @@ export default class DayPicker extends React.Component {
...
+      el.style[styleKey] = transformStyles[styleKey];
+    });
+  }
+
+  calculateDimension(el, axis, range = 'inner', withMargin = false) {
+    if (!el) {
+      return 0;
+    }
+
+    const style = window.getComputedStyle(el);
```
Rendez: Would it be an option to provide padding and border as inline styles to the component to avoid getting computed style?

ljharb: Computed style is about its actual size - padding and border just contribute to its desired size. User stylesheets, different font sizes, etc, could all effect the rendered dimensions.

Rendez: Sure, but if border and padding and margin are passed as props and used for inline css, then getBoundingClientRect() could be used instead, which is x100 faster

ljharb: Ah, I see what you mean. That's not a bad idea, but might require a bit too much coupling.

milesj:
@ljharb @majapw Updated the code. Added some stub tests as well.

All I really need to do is test IE8/9/10, but not sure how, because storybook does not work in it.

majapw:
@milesj So the best bet is probably to do this using airbnb.com at this point. We can `npm link` to this repo locally and specifically to this branch, and check to see if our homepage works as expected. I can probably do this later today or tomorrow, although it will probably require another PR first. Are you okay with holding off on this for a few more days? We can have this in a minor release then. 

I really don't want to merge this in and then after the fact discover that things are broken in those browsers.

milesj:
Yeah, no problem. No rush to get this out right now.

majapw:
So I tested on IE9, and it works as expected (no animations, but no errors). Same as before.

IE10 works well. :) Have you looked at mobile safari and mobile chrome?

milesj:
Haven't tested mobile, but I don't foresee any issues. It's pretty straight forward.

I'll rebase tomorrow morning.

majapw:
Awesome. :+1:

milesj:
@majapw Rebased. Passed linting and tests locally.

PR #19, opened 2016-08-19
Adds webpack and babel-plugin-syntax-jsx as dev dependencies

majapw:
We were noticing some issues with (a) starting storybook and (b) getting the `react-svg-loader` to work properly with node 4.2.4. This adds those missing dependencies to hopefully fix those issues.


ljharb:
LGTM

PR #20, opened 2016-08-19
pass tabIndex down as a prop to input components

travisbloom:


test/components/DateInput_spec.jsx
```diff
@@ -12,6 +12,11 @@ describe('DateInput', () => {
       expect(wrapper.is('.DateInput')).to.equal(true);
     });
 
+    it('will apply tabIndex', () => {
+      const wrapper = shallow(<DateInput tabIndex={1} id="date" />);
+      expect(wrapper.find('input').is('[tabIndex]')).to.equal(true);
```
ljharb: should this assert that the tabIndex is actually 1?

travisbloom: good catch, fixed

lencioni:
@travisbloom why do you want to be able to do this? Can you help me understand your use-case?

travisbloom:
@lencioni we have a react form being rendered on a legacy page. We want to ensure that form inputs get tabbed over before other tabbable fields that exist on the page. We pass a tabIndex default of 1 to all form inputs on that page.

lencioni:
Thanks for the explanation! Any chance you can include this in your commit message to help future explorers understand why this exists? 

https://medium.com/brigade-engineering/the-secrets-to-great-commit-messages-106fc0a92a25

backwardok:
@travisbloom - adding in `tabIndex` in this way has bad accessibility ramifications. If you want to have a user's focus start in a particular place, it's better to just focus the first element you want to be interacted with and then have the user be able to go through the natural flow of the page. By setting the value to anything other than 0 or -1, you are forcing a particular tab flow, which results in odd keyboard navigation. See more about that here: http://webaim.org/techniques/keyboard/tabindex

travisbloom:
@backwardok interesting article! Do we want to handle the use case from the article below or should I go ahead and close this PR?

> - The default tab order is not ideal, AND
> - The tab order cannot be changed by rearranging items in the content and/or by altering the style sheet to reflect the best visual arrangement

backwardok:
@travisbloom I'd say we should probably close this PR. Probably 99% of the time, adding an explicit tab order is not the best way to solve the problem, and adding in the ability to do it here encourages people to use it.

travisbloom:
sounds good

backwardok:
@bernatfortet can you clarify why you would want users to avoid tabbing into the calendar? Seems like that would break behavior for keyboard users.

bernatfortet:
You're right @backwardok .
The Date component was visually hidden and tabbing was brining it into view without me wanting that.
I just stopped rendering the component until needed. 
So it's not an issue for me anymore. Thanks for the quick reply

PR #32, opened 2016-08-21
Run test suite in Firefox via Karma

iancmyers:
Since the primary use-case for react-dates is in a browser environment it seems appropriate to run the test suite in a browser.

I've added configuration to run the test suite in Firefox via Karma with the necessary adjustments to the Travis config. I've commented the two places in the Karma config where I'm doing something odd in order to make things work.


test/components/CalendarDay_spec.jsx
```diff
@@ -44,6 +44,10 @@ describe('CalendarDay', () => {
       handleDayClickSpy = sinon.spy(CalendarDay.prototype, 'handleDayClick');
     });
 
+    afterEach(() => {
+      CalendarDay.prototype.handleDayClick.restore();
```
ljharb: we should probably use sinon-sandbox instead of having to restore spies after every test manually

karma.conf.js
```diff
@@ -0,0 +1,88 @@
...
+module.exports = (config) => {
+  config.set({
+    basePath: '',
+
+    frameworks: ['mocha', 'sinon', 'chai'],
+
+    files: [
+      'test/_helpers/*.js',
+      'test/utils/*.js',
+      'test/components/*.jsx',
```
ljharb: do we have to specify the file extensions here? it seems more robust to just have "*"

karma.conf.js
```diff
@@ -0,0 +1,88 @@
...
+      ],
+      module: {
+        loaders: [
+          {
+            test: /\.jsx?/,
+            loader: 'babel-loader',
+            include: [
+              path.join(__dirname, 'src'),
+              path.join(__dirname, 'test'),
+              path.join(__dirname, 'node_modules', 'airbnb-js-shims'),
```
ljharb: `require.resolve('airbnb-js-shims')`

in fact, i think this whole array could be `['./src', './test', 'airbnb-js-shims'].map(require.resolve)`?

iancmyers:
@ljharb Issues addressed.

karma.conf.js
```diff
@@ -0,0 +1,88 @@
...
+        new webpack.NormalModuleReplacementPlugin(/^\.\/package$/, (result) => {
+          if (/cheerio/.test(result.context)) {
+            result.request = './package.json';
+          }
+        }),
+      ],
+      module: {
+        loaders: [
+          {
+            test: /\.jsx?/,
```
lencioni: This regex should probably be `/\.jsx?$/`

test/components/CalendarDay_spec.jsx
```diff
@@ -44,6 +44,10 @@ describe('CalendarDay', () => {
       handleDayClickSpy = sinon.spy(CalendarDay.prototype, 'handleDayClick');
     });
 
+    afterEach(() => {
```
goatslacker: Does [this file](https://github.com/airbnb/react-dates/blob/master/test/_helpers/restoreSinonStubs.js) not run in karma?

test/components/CalendarDay_spec.jsx
```diff
@@ -84,6 +92,10 @@ describe('CalendarDay', () => {
       handleDayMouseUpSpy = sinon.spy(CalendarDay.prototype, 'handleDayMouseUp');
     });
 
+    afterEach(() => {
+      sinon.restore();
```
spikebrehm: Perhaps you could do a single, top-level `afterEach()` that calls this?

majapw:
@iancmyers can you address comments and rebase?

test/components/CalendarDay_spec.jsx
iancmyers: @goatslacker It gets included in the Webpack build as far as I can tell, but I still got double-wrap errors from sinon that necessitated this change.

test/components/CalendarDay_spec.jsx
iancmyers: @spikebrehm There's one here: https://github.com/airbnb/react-dates/blob/master/test/_helpers/restoreSinonStubs.js, but it didn't seem to do the trick.

ljharb: https://github.com/airbnb/react-dates/blob/master/package.json#L15 needs to be updated to explicitly pass the _helpers paths to mocha first, before the rest of the tests.

majapw:
So after a lot of digging, it seems that the reason that the `restoreSinonStubs` helper module is not functioning as expected with karma is that we end up with duplicate instances of `sinon` when using `karma-sinon`. This is at least partly due to `karma-sinon`'s implicit dependency on an older version of `sinon` (https://github.com/yanoosh/karma-sinon/blob/master/index.js#L8). 

We can merge this in as is with the `afterEach` restores, which is gross but I feel like valid while we figure out this issue for ourselves.

PR #35, opened 2016-08-23
First try at month/year menus

rodryquintero:
what do you think?

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "build": "npm run clean && npm run build:js && npm run build:css && npm run build:svg",
     "build:js": "babel src/ -d lib/ --ignore src/components",
     "build:css": "node-sass css/styles.scss lib/css/_datepicker.css",
-    "build:svg": "webpack",
+    "build:svg": "webpack --watch",
     "clean": "rimraf lib",
     "lint": "eslint --ext .js,.jsx src test",
     "mocha": "mocha --compilers js:babel-register,jsx:babel-register --require airbnb-js-shims --recursive ./test/_helpers",
@@ -87,5 +87,8 @@
     "react": ">=0.14",
     "react-dom": ">=0.14",
     "react-addons-shallow-compare": ">=0.14"
+  },
+  "jshintConfig": {
+    "esnext": true
   }
 }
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -87,5 +87,8 @@
     "react": ">=0.14",
     "react-dom": ">=0.14",
     "react-addons-shallow-compare": ">=0.14"
+  },
+  "jshintConfig": {
```
ljharb: please remove this; we don't use jshint.

src/components/DateRangePicker.jsx
```diff
@@ -25,7 +25,7 @@ import {
   START_DATE,
   END_DATE,
   HORIZONTAL_ORIENTATION,
-  VERTICAL_ORIENTATION,
+  VERTICAL_ORIENTATION
```
ljharb: trailing commas are required by our linter everywhere they are supported.

ljharb:
Can you include a screenshot and/or a link to a demo page?

rodryquintero:
Sure. Here are some.
image

image
image

ljharb:
What happens if you select a month/year that falls outside of the selectable range?

rodryquintero:
It displays the calendar but the days are grayed out. You cannot select them.

image

This, of course, is not ready to be merged as all tests have failed. I do not have experience using the test tools so I did not check before doing the pull request.

Feel free to delete this pull request. Just wanted to do a "proof of concept" and show you.

ljharb:
It seems like if you can navigate to a useless month, there should be an easy way to jump back to the closest month with selectable days. Alternatively, based on the "year" setting, months with no selectable days should not be selectable.

rodryquintero:
Will get to it. Thanks for the help.

arnars:
How is this coming along? We could really use this feature in our current project.

selvagsz:
Any progress in this feature ?

majapw:
This particular implementation is going to be really hard to resolve with the structural changes that have been included in react-dates since this PR was opened. I'm going to be investigating this conceptually pretty soon tho and trying to push something through the door.

PR #43, opened 2016-08-26
Fix demo GIF

lencioni:
I originally added this to the gh-pages branch so that the GIF wouldn't
bloat the repo or the package that people install. However, I didn't
realize that we have a script that blows all of those files away
whenever we publish, so it quickly disappeared. We've decided that repo
bloat is okay, and we can .npmignore this file for the sake of
installers.

While I was at it, I sorted some things alphabetically in the .npmignore
file.


majapw:
wondrous. lgtm.

PR #44, opened 2016-08-26
Export date comparison methods

majapw:
This PR exports the following date comparison methods as utility classes:
```
  isInclusivelyAfterDay
  isInclusivelyBeforeDay
  isNextDay
  isSameDay
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -371,6 +371,20 @@ The `monthFormat` prop abides by [moment's date formatting rules](http://momentj
   monthFormat: PropTypes.string,
 ```
 
+## Utility Methods
+
+### Date Comparison
+
+We provide four utility methods for date comparison:
+```
+  isInclusivelyAfterDay
+  isInclusivelyBeforeDay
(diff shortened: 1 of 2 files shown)


ljharb:
Assuming these methods are unit-tested, should we also have a test that verifies the things that `index.js` exports?

majapw:
That is probably a good idea. Can you write a test for this file in the same way we test our components?

majapw:
(also these methods are definitely unit-tested, who are you even talking to @ljharb :P)

ljharb:
We can, but we'd need to do `npm run build` on each test run in order to get the lib components. I can add a commit to your branch if you like?

ljharb:
Actually now that I think about it, that test would literally just be duplicating the imports/requires in the index file, and verifying that they were exported - that probably won't add much value.

majapw:
Okay! If you can think of a good way to do this, I would not mind the
commit on this branch!

Issue #17, opened 2016-08-18
Cannot indicate which month is visible on open


majapw:
Ideally this should be a controlled prop that would default to the current month. That way you can close the `<DayPicker />` and reopen to where you were previously

My vision is that this would be a `currentMonth` prop on the `SingleDatePicker` and on the `DateRangePicker` that would take in a moment object. If a parent component wanted to keep track of this in the state then, they could do so using the `onPrevMonthClick` and `onNextMonthClick` callbacks.

travisbloom:
I can take a stab at this. You still want DayPicker to own the current month in its state after mounting right? Or should currentMonth be able to dynamically change and rerender the view?

majapw:
I think that the `DayPicker` should still own its state, so this would not be fully controlled I guess. This would just be what month is visible when you first open the datepicker. So maybe we could name it something like `visibleMonthOnOpen` or something? What do you think?

travisbloom:
Ya, I was concerned that if currentMonth was a prop the parent owned after showing it would mess up all the animation goodness.

Agree with calling it visibleMonthOnOpen. If there is a date passed to the component, do you want to use that as the default month as a backup?

Order I'm envisioning: month showing on mount goes `visibleMonthOnOpen || date || moment().startOf('month')`?

travisbloom:
Trying to think if it's ever expected behavior for the month to be the current month if an initial date prop was passed...

majapw:
Hmmm, I feel like it should be either explicitly defined or it should be the present month. I think that maybe it's not transparent enough behavior if we fall back to the selected date (plus with the `DateRangePicker`, do you use `startDate`? `endDate`? I think we probably want `visibleMonthOnOpen || moment().startOf('month')`

travisbloom:
sounds good!

Corei13:
@majapw @travisbloom fyi: https://github.com/airbnb/react-dates/pull/12/files#diff-38206f500edecedcd970973ba637f407R12

travisbloom:
@Corei13 does that handle this use case? 

If a user selects a `visibleMonthOnOpen` that is before `minimumAllowedDate`, I'm assuming they should still be taken to that month. I could add a warning to the prop type of visibleMonthOnOpen if it detects its before minimumAllowedDate?

majapw:
I feel like that should be on the parent component to guarantee, not on the
DRP/SDP itself and so these should maybe be two separate changes.

jcreamer898:
[...]

PR #47, opened 2016-08-30
Add initialVisibleMonth fn prop to DayPicker

travisbloom:
Related to https://github.com/airbnb/react-dates/issues/17
Adds a new prop to DateRangePicker, SingleDatePicker, and DayPicker
that passes a fn to DayPicker. On the initial focus, that fn will be called
and set the initial month to the returned moment object.


travisbloom:
this PR is on hold until https://github.com/airbnb/react-dates/pull/32

travisbloom:
There's some larger issues about how focused is passed to the different components I need to understand better, closing for now.

Issue #26, opened 2016-08-19
Would love to be able to change the format of the inputs


jcreamer898:
I noticed that the input values are always in mm/dd/yyyy format, I think because of https://github.com/airbnb/react-dates/blob/04e2840fe6181dadcb5edd695a0550b46ea9b054/src/components/DateRangePicker.jsx#L395.

It'd be nice to be able to format that date to a custom format.

majapw:
Yeah! This is def a thing that I've kind of started thinking about in the utility classes (namely the `toMomentObject` method takes in a custom format as an optional argument). 

My thought is that the `DateRangePicker` and `SingleDatePicker` would have a `customDateFormat` prop and then we could have `toLocalizedDateString` use that in the render method and `toMomentObject` use that `onStartDateChange` and `onEndDateChange` methods. If it's falsey, then we just fall back to the `L` format. :)

jcreamer898:
Sweet. I can try to PR for it.

majapw:
Sounds great!

PR #52, opened 2016-09-01
Allows for a custom display format 

majapw:
Fix for https://github.com/airbnb/react-dates/issues/26

@jcreamer898

### DIFF
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -54,6 +54,7 @@ const defaultProps = {
   onNextMonthClick() {},
 
   // i18n
+  displayFormat: moment.localeData().longDateFormat('L'),
   monthFormat: 'MMMM YYYY',
   phrases: {
     closeDatePicker: 'Close',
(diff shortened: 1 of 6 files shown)


jcreamer898:
Well dang! Haha, that's awesome, you rock @majapw.

👍 👍

src/components/DateRangePicker.jsx
```diff
@@ -198,6 +199,14 @@ export default class DateRangePicker extends React.Component {
     }
   }
 
+  getDateString(date) {
+    const { displayFormat } = this.props;
+    if (displayFormat) {
+      return date && date.format(displayFormat);
```
ljharb: when `date` is falsy, can you remind me what `toLocalizedDateString` will return?

majapw: `toLocalizedString` returns `null` for falsey dates. I'm going to go back and add some tests for that I think.

majapw: Ha! I already wrote a test - https://github.com/airbnb/react-dates/blob/master/test/utils/toLocalizedDateString_spec.js#L7:L9

majapw:
Hmm, I just realized that this still expects `L` format dates when typing... I think that this prop should supercede that in that case as well.

src/components/DateRangePicker.jsx
ljharb: would it be better to do `if (date && displayFormat)` so that falsy dates always go through `toLocalizedDateString`? With the current approach, passing `NaN` with a format will return `NaN`, for example.

majapw:
Hey @airbnb/webinfra can someone gimme a stamp on this?

lencioni:
LGTM

PR #53, opened 2016-09-01
Move constants to top level directory

majapw:
This is to allow for `import { START_DATE END_DATE } from 'react-dates/constants';` and it makes sense for it not to need transpilation since the file is literally a collection of strings. However, it is breaking.


ljharb:
LGTM - looks like you'll need to update some paths in tests too

majapw:
Oops! I totally thought I had.

ljharb:
LGTM

Issue #51, opened 2016-09-01
CSS issues


dipetersen:
I am trying to use react-dates (SingleDatePicker) in one of my projects.  I did the normal `import {SingleDatePicker} from 'react-dates'`, which enabled me to use the control in my code.  The problem is that the CSS isn't picked up.  I forced import the CSS by using `import '../node_modules/react-dates/lib/css/_datepicker.css';`

This does render the date picker with CSS but the Month caption is hidden behind the left selector arrow.  I'm pretty new to React so I am allowing that I don't know what I'm doing.  Any suggestions?

majapw:
Hi @dipetersen! Can you post a screenshot?

As for the `import`, you're totally right! We don't package our css with the components... yet! In our own system, we're sprocket-requiring the css file from node_modules (ick), but we want to move to CSS in JS as well... soon! PRs most definitely welcome.

dipetersen:
After a bit of debugging, I found the culprit.  It seems as though the Bootstrap definition of _caption_ superseded the _caption_ definition in the CSS for react-dates.  I'm going to play around with it for a while to see if I can come up with a solution but CSS isn't my strong suit.  Here's what it looks like on my screen at the moment.

![react-dates-css](https://cloud.githubusercontent.com/assets/2683871/18206352/defb4a46-70eb-11e6-9b75-c7ec949b94db.png)

dipetersen:
I think if we just add `text-align: center` to the _CalendarMonth__caption_ class, it will solve the problem without causing any problems.  It worked in my environment.

dipetersen:
PR submitted.

ljharb:
Issues should remain open until relevant PRs are merged.

PR #54, opened 2016-09-02
Update CalendarMonth__caption class

dipetersen:
Bootstrap caused the caption to align left.  Adding a text-align:center makes sure that the Month caption is always aligned in the center.


ljharb:
This will fix #51?

dipetersen:
It fixed it in my environment.  It shouldn't conflict with anything else since the scope is limited to that class.

majapw:
LGTM. If you rebase your change, we can merge it in. :)

Issue #55, opened 2016-09-02
Firefox displaying October 15th two times


wmartins:
I was just testing this using the github pages and realized that when the calendar is opened using Firefox (48.0.2), it is displaying two times the "October 15th" day. Here is a screenshot:

![2016-09-02_14-37-24](https://cloud.githubusercontent.com/assets/1560801/18212868/d61a3524-711b-11e6-8748-4c1b377126f3.png)

I don't know if it has something to do with my timezone but I'm running this at BRT Timezone (GMT-03:00).

I'm running this using a windows machine, but I tested that in Firefox developer edition (in a Mint VM) and I can reproduce the problem.

If you need more info, just tell me! :)

ljharb:
This is likely because in Brazil, daylight savings time hits at midnight instead of 2AM - the fix is likely to use a time on our moment objects of `2:01 AM` rather than midnight?

wmartins:
Yep, daylight saving time will start at October 16th (http://www.timeanddate.com/time/change/brazil/brasilia). I've done some tests, I don't know if you can use `moment.utc` in function `getCalendarMonthWeeks`, because I think that solves this issue:

![2016-09-02_15-16-52](https://cloud.githubusercontent.com/assets/1560801/18213890/b9dbdd7c-7120-11e6-8cd1-ae421bb99cd5.png)

In the screenshot, you can see that `Array[8]` is the "wrong" output. You can see that when using `moment.utc`, the result is correct (`Array[7]`). What do you think about it?

majapw:
Nooooooooo. Okay. We will fix this asap.

wmartins:
@majapw, sure! I tested that on Chrome and everything is ok, I will try to install Safari here to see if it is a problem, but I think it is FF specific.

wmartins:
@majapw, it also happens on Safari (`Version 9.1.2 (11601.7.7)`) and IE 11 (`Version 11.0.9600.18376`).

majapw:
Sweet I can repro by changing my time zone. I will dig into it. :)

Also:
https://github.com/moment/moment/issues/2353
https://github.com/moment/moment/issues/1440

majapw:
Courtesy of my coworker, @mstorus:
Mark: How many datepicker issues would you say you've been dealing with today?
Me: I don't know, a lot?
Mark: Would you say it's been... a B(r)azillion?

wmartins:
@majapw hahahahahahaha

felquis:
[...]

PR #56, opened 2016-09-02
Fixes Brazilian Daylights Savings Time duplication bug in Firefox

majapw:
Fix for https://github.com/airbnb/react-dates/issues/55

@ljharb @wmartins 

Also, for a fun time, take a look at https://en.wikipedia.org/wiki/Daylight_saving_time_by_country

Did you know that Greenland does DST at 10PM in the fall and 11PM in the spring? Did you know that not all of Brazil uses Daylight Savings Time? Did you know that Arizona does not have Daylight Savings Time, even though the rest of the USA does, but that the Navajo Nation Community inside of Arizona does??? U.U

### DIFF
--- a/src/utils/getCalendarMonthWeeks.js
+++ b/src/utils/getCalendarMonthWeeks.js
@@ -1,6 +1,6 @@
 export default function getCalendarMonthWeeks(month, enableOutsideDays) {
-  const firstOfMonth = month.clone().startOf('month');
-  const lastOfMonth = month.clone().endOf('month');
+  const firstOfMonth = month.clone().startOf('month').utc();
+  const lastOfMonth = month.clone().endOf('month').utc();
 
   const currentDay = firstOfMonth.clone();
   let currentWeek = [];


src/utils/getCalendarMonthWeeks.js
```diff
@@ -1,6 +1,6 @@
 export default function getCalendarMonthWeeks(month, enableOutsideDays) {
-  const firstOfMonth = month.clone().startOf('month');
-  const lastOfMonth = month.clone().endOf('month');
+  const firstOfMonth = month.clone().startOf('month').add(2, 'hours').add(1, 'minute');
```
ljharb: it would be great to add a test with the brazil moment locale :-)

ljharb:
LGTM

wmartins:
Hey @majapw, the code fixes it, but I don't know if it is the best solution for that.

For example, if you set your timezone to "Europe/Athens" you end up having things like that:



(Displaying 1 extra day, and with external days prop  as false)

ljharb:
hmm, there's probably other logic that needs adjusting to take the extra 121 minutes into account.

majapw:
Yeah. Good catch! I will dig around more into this. :)

wmartins:
@ljharb @majapw use utc in this case is an option?

Setting `firstOfMonth` and `lastOfMonth` to utc fixes it. The downside is that the following tests will break:

```
1) getCalendarMonthWeeks enableOutsideDays arg is false first non-null element is first of the month
2) getCalendarMonthWeeks enableOutsideDays arg is false last non-null element is last of the month
3) getCalendarMonthWeeks enableOutsideDays arg is true contains last of the month
```

I will check it out if this is something specific to tests or if it is really breaking things.

majapw:
Those tests should only be comparing by day though... that seems concerning.

wmartins:
@majapw, sorry, totally my bad, I was just testing that using your branch (I modified your modified code), let me try that in `master` branch

wmartins:
@majapw, tested the `utc` solution on `master` (now I did it correctly) and everything works fine (with `America/Sao_Paulo` and `Europe/Athens` TZs). Also, the tests are passing. :)

majapw:
So I changed `getCalendarMonthWeeks` to use utc locally as you suggested @wmartins, but my tests don't seem to have been affected.

wmartins:
I'll run more tests here to see if there is nothing wrong

wmartins:
@majapw, I tested here and it is showing a an extra day in month, I think you need to change the order of `utc` calling. Because your code is setting it to utc after getting the start of month.

Change those:

```javascript
const firstOfMonth = month.clone().startOf('month').utc();
const lastOfMonth = month.clone().endOf('month').utc();
```

to:

```javascript
const firstOfMonth = month.clone().utc().startOf('month');
const lastOfMonth = month.clone().utc().endOf('month');
```

Here is the problem occurring:

majapw:
I think the edge case I'm concerned about is... let's say I am in PT, and it is 5:01PM on August 31st.

`moment()` returns Aug 31st. Does `moment.utc()` return September 1st? If so, would calling `getCalendarMonthWeeks` at that time actually return the weeks for September?

wmartins:
Yeah, this could be a problem, depending on how this method is being
called. We can try to "build" a new "base date" (an utc one) based on the
month (the current user date, non utc). And then get the start of month and
end of month. It will guarantee that the month is correct.

What do you think about it?

Em sex, 2 de set de 2016 20:08, Maja Wichrowska <notifications@github.com>
escreveu:

majapw:
Yeah, I was thinking that the best bet is probably to extract the month information from the passed in object and then construct an entirely new moment object.... and then apply the utc method to that. Yuck!

wmartins:
Yep. Another option is to check if moment has some "set to utc without
modifying the date" method. If so, it would preserve the date, then it
would be safe to use it.

Em sex, 2 de set de 2016 20:44, Maja Wichrowska <notifications@github.com>
escreveu:

wmartins:
Hey @majapw, I think I've found a better solution for this!

First, the problem is occurring because we are building today (under GMT-03:00) all the dates in the "future", but, `moment#add` is messing things up when the date is not an utc date. So, it adds one more day when the timezone is changed through the year (in this case, at October 15th). What we can do to mitigate this is to use utc dates in `firstOfMonth` and `lastOfMonth`, but then we would end up with the problems talked before (changing the month based on the current date and screwing up everything).

I have a better solution that doesn't screw up everything:

1. Create a `baseDate`, that is a clone of the `month` object, but with the [`utcOffset`](http://momentjs.com/docs/#/manipulating/utc-offset/) **fixed**
2. Now, when calling `moment#add`, moment knows how to do the calculation (https://github.com/moment/moment/issues/2353#issuecomment-97470780). Using `utcOffset` will "flag" this date as utc

Here is the code I used:

```javascript

const baseDate = month.clone().utcOffset(month.utcOffset());
const firstOfMonth = baseDate.clone().startOf('month');
const lastOfMonth = baseDate.clone().endOf('month');
```

It seems more reliable, and it is not that hard to understand, because you're basically saying "ok moment, use that timezone, do all the math you need and display things based on that timezone".

What do you think?

majapw:
This sounds amazing! If you want to put up a PR for that (maybe with a comment explaining the reasoning), I would be happy to close this one and merge that in. :)

wmartins:
Oh, it would be a pleasure, here is the PR @majapw: #58!

PR #57, opened 2016-09-02
Fixes edge-case `CalendarMonth` height issue 

majapw:
I recently noticed that sometimes the height of the datepicker was not rendering correctly, specifically that the table seemed to be an extra row representing the full first week of the following month:
screen shot 2016-09-02 at 4 10 45 pm

Turns out, when the next month starts at the beginning of the week (according to whatever locale you're in), we still render that full week as outside days. 

I also discovered that this was a ticking time bomb as related to the tests for `getCalendarMonthWeeks` that would have started breaking in December. As a result I've added some tests to cover this edgecase as well.

### DIFF
--- a/src/utils/getCalendarMonthWeeks.js
+++ b/src/utils/getCalendarMonthWeeks.js
@@ -24,13 +24,17 @@ export default function getCalendarMonthWeeks(month, enableOutsideDays) {
     }
   }
 
-  // days belonging to the next month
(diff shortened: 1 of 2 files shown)


ljharb:
LGTM

ljharb:
LGTM

ljharb:
LGTM

ljharb:
:shipit:

ljharb:
LGTM

ljharb:
LGTM

PR #65, opened 2016-09-06
Custom display formatting

averted:
I have a use-case where I need to set a custom label (e.g. “Now”) if the selected date is set to Today. This is very similar to #52 except `displayFormat` is expecting a string, so you don’t get a lot of flexibility.

This PR gives you a bit more control over display label formatting.

I think the ultimate solution would be merging `displayLabel` and `displayFormat` (from #52) into a single prop, which would expect either a string or a function. Something in the lines of:

```
...

getDateString(date) {

  if (typeof displayFormat === 'function') {
    return displayFormat(date);
  } else if (typeof displayFormat === ‘string’) {
    return date && date.format(displayFormat);
  }

  return toLocalizedDateString(date);
}

...
```

Thoughts?

Example:
airbnb-datepicker

### DIFF
--- a/src/components/SingleDatePicker.jsx
+++ b/src/components/SingleDatePicker.jsx
@@ -43,6 +43,7 @@ const defaultProps = {
   onNextMonthClick() {},
 
   // i18n
+  displayLabel: null,
   monthFormat: 'MMMM YYYY',
   phrases: {
     closeDatePicker: 'Close',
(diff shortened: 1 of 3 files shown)


majapw:
So I've thought about this a bit and while I def understand the need, I am also somewhat loath to overload the `displayFormat` method or add another method. Given that this is a controlled component, I can think of a way to fill your use case with just the string. For the SDP, it'd be something like this:

xuexiang0304:
Hi, Just tried the displayFormat feature, it works very well, Exactly what I wanted. However, somehow, the allowPastDates feature doesn't work anymore after I updated to V2.2.0.

ljharb:
@xuexiang0304 the place to post this isn't on this issue, but in a new one; however, the [changelog](https://github.com/airbnb/react-dates/blob/master/CHANGELOG.md#v200) tells you that in v2.0, that prop was removed.

PR #70, opened 2016-09-09
Add initialVisibleMonth fn prop to DayPicker

travisbloom:
Adds a new prop to DateRangePicker, SingleDatePicker, and DayPicker
that passes a fn to DayPicker. On the initial focus, that fn will be called
and set the initial month to the returned moment object.

solves issue https://github.com/airbnb/react-dates/issues/17

### DIFF
--- a/.istanbul.yml
+++ b/.istanbul.yml
@@ -11,6 +11,6 @@ instrumentation:
 check:
   global:
     statements: 89
-    branches: 70
+    branches: 69
     lines: 91
     functions: 79
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -44,6 +44,7 @@ const defaultProps = {
   showClearDates: false,
   disabled: false,
   reopenPickerOnClearDates: false,
+  initialVisibleMonth: () => moment(),
 
   orientation: HORIZONTAL_ORIENTATION,
   withPortal: false,
@@ -330,6 +331,8 @@ export default class DateRangePicker extends React.Component {
       withPortal,
       withFullScreenPortal,
       enableOutsideDays,
+      initialVisibleMonth,
+      focusedInput,
     } = this.props;
 
     const modifiers = {
@@ -370,6 +373,8 @@ export default class DateRangePicker extends React.Component {
           onNextMonthClick={onNextMonthClick}
           monthFormat={monthFormat}
           withPortal={withPortal || withFullScreenPortal}
+          hidden={!focusedInput}
+          initialVisibleMonth={initialVisibleMonth}
           onOutsideClick={onOutsideClick}
         />
 
--- a/src/components/DayPicker.jsx
+++ b/src/components/DayPicker.jsx
@@ -30,6 +30,8 @@ const propTypes = {
   modifiers: PropTypes.object,
   orientation: OrientationShape,
   withPortal: PropTypes.bool,
+  hidden: PropTypes.bool,
+  initialVisibleMonth: PropTypes.func,
   onDayClick: PropTypes.func,
   onDayMouseDown: PropTypes.func,
   onDayMouseUp: PropTypes.func,
@@ -52,6 +54,8 @@ const defaultProps = {
   modifiers: {},
   orientation: HORIZONTAL_ORIENTATION,
   withPortal: false,
+  hidden: false,
+  initialVisibleMonth: () => moment(),
   onDayClick() {},
   onDayMouseDown() {},
   onDayMouseUp() {},
@@ -71,8 +75,10 @@ const defaultProps = {
 export default class DayPicker extends React.Component {
   constructor(props) {
     super(props);
+
+    this.hasSetInitialVisibleMonth = !props.hidden;
     this.state = {
-      currentMonth: moment(),
+      currentMonth: props.hidden ? moment() : props.initialVisibleMonth(),
       monthTransition: null,
       translationValue: 0,
     };
@@ -86,6 +92,15 @@ export default class DayPicker extends React.Component {
     if (this.isHorizontal()) {
       this.adjustDayPickerHeight();
       this.initializeDayPickerWidth();
+    }
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (!this.hasSetInitialVisibleMonth && !nextProps.hidden) {
+      this.hasSetInitialVisibleMonth = true;
+      this.setState({
+        currentMonth: nextProps.initialVisibleMonth(),
+      });
     }
   }
 
--- a/src/components/SingleDatePicker.jsx
+++ b/src/components/SingleDatePicker.jsx
@@ -38,6 +38,8 @@ const defaultProps = {
   orientation: HORIZONTAL_ORIENTATION,
   withPortal: false,
   withFullScreenPortal: false,
+  initialVisibleMonth: () => moment(),
+
 
   onPrevMonthClick() {},
   onNextMonthClick() {},
@@ -181,6 +183,8 @@ export default class SingleDatePicker extends React.Component {
       onNextMonthClick,
       withPortal,
       withFullScreenPortal,
+      focused,
+      initialVisibleMonth,
     } = this.props;
 
     const modifiers = {
@@ -209,6 +213,8 @@ export default class SingleDatePicker extends React.Component {
           onNextMonthClick={onNextMonthClick}
           monthFormat={monthFormat}
           withPortal={withPortal || withFullScreenPortal}
+          hidden={!focused}
+          initialVisibleMonth={initialVisibleMonth}
           onOutsideClick={onOutsideClick}
         />
 
--- a/README.md
+++ b/README.md
@@ -108,6 +108,11 @@ By default, we do not show days from the previous month and the next month in th
   enableOutsideDays: PropTypes.bool
 ```
 
+`initialVisibleMonth` indicates the month that should be displayed initially when the calendar is first opened. The prop is a function that must return a Moment.js object. This function will be called the first time the user focuses on the `DateRangePicker`/`SingleDatePicker` inputs or when the `focused` prop is passed to the `DayPicker` component.
+```
+   initialVisibleMonth: PropTypes.func
+```
+
 **DayPicker presentation:**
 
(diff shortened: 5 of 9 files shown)


test/components/DayPicker_spec.jsx
```diff
@@ -260,6 +261,20 @@ describe('DayPicker', () => {
       });
     });
 
+    describe('#componentWillReceiveProps', () => {
```
travisbloom: This test can't run currently because of https://github.com/airbnb/react-dates/blob/master/test/components/DayPicker_spec.jsx#L225

Is there an open issue/discussion for how to solve? getting `It looks like you called `mount()` without a global document being loaded.` when i remove `skip`

src/components/DayPicker.jsx
```diff
@@ -52,6 +54,8 @@ const defaultProps = {
   modifiers: {},
   orientation: HORIZONTAL_ORIENTATION,
   withPortal: false,
+  focused: true,
```
ljharb: absence and `false` should be the same - instead of a default of true, could this either default false, or use an inverted name such that it can default false?

src/components/DayPicker.jsx
```diff
@@ -71,8 +75,10 @@ const defaultProps = {
 export default class DayPicker extends React.Component {
   constructor(props) {
     super(props);
+
+    this.hasSetInitialVisibleMonth = !!props.focused;
     this.state = {
-      currentMonth: moment(),
+      currentMonth: props.focused ? props.initialVisibleMonth() : moment(),
```
ljharb: this logic seems useful to encapsulate in a function (not necessarily a class method)

src/components/SingleDatePicker.jsx
```diff
@@ -167,6 +167,8 @@ export default class SingleDatePicker extends React.Component {
       onNextMonthClick,
       withPortal,
       withFullScreenPortal,
+      focused,
```
ljharb: propType and defaultProp for this?

src/components/DayPicker.jsx
travisbloom: something like `const triggerInitialVisibleMonth = (props) => props.focused ? props.initialVisibleMonth() : moment();` for clarity?

ljharb: sure, that works fine! just so it's harder to screw up the logic later, when it's in two places

src/components/DayPicker.jsx
travisbloom: i agree that defaulting to true is confusing, but in this case it might make sense. It seems unintuitive for people to have to specify `focused` when rendering the DayPicker, the default behavior should be "if you render this component, it will show".

I could change the name, but then it will deviate from the prop names used in `SDP` and `DRP`... if you think that it would still be best i can make it `hidden`?

travisbloom:
@ljharb fixes added, let me know if you have more insight into the `It looks like you called mount() without a global document being loaded` issue

travisbloom:
also the linter is failing because of https://github.com/airbnb/react-dates/pull/70/files#diff-883056b4b11263241c339f927e93411bR75

the issue is that if I make it one line I get `exceeds max length of 100` lint error.

Is the expected style below?
```javascript
const triggerInitialVisibleMonth = (props) => 
  props.hidden ? moment() : 
  props.initialVisibleMonth();
```

src/components/DayPicker.jsx
ljharb: I don't mind if it deviates from the other components. `hidden` is fine, but the opposite of "focused" is "blurred", and that's not a great name either :-/

ljharb:
```js
const triggerInitialVisibleMonth = ({ hidden, initialVisibleMonth }) => (
  hidden ? moment() : initialVisibleMonth()
);
```

travisbloom:
@ljharb whoops, totally spaced on that, good call :)

travisbloom:
Any reason the tests that are failing wouldn't be running/failing on my local machine?

ljharb:
@travisbloom yes - rebase on latest master; one of the dependencies had to be locked down to avoid a bug.

src/components/DayPicker.jsx
```diff
@@ -68,11 +72,17 @@ const defaultProps = {
   monthFormat: 'MMMM YYYY',
 };
 
+const triggerInitialVisibleMonth = (props) => (
+  props.hidden ? moment() : props.initialVisibleMonth()
```
majapw: I'm not sure that we need this logic to be perfectly honest. The constructor is only called once and even if the datepicker is visible at that point, it will be the first time its ever rendered so I think we can just set `state.currentMonth` to `props.initialVisibleMonth()`. What is the edge case this is trying to solve?

src/components/DayPicker.jsx
```diff
@@ -68,11 +72,17 @@ const defaultProps = {
...
+);
+
 export default class DayPicker extends React.Component {
   constructor(props) {
     super(props);
+
+    this.hasSetInitialVisibleMonth = !props.hidden;
```
majapw: Similarly I don't think we need the `hasSetInitialVisibleMonth` logic either.

test/components/DayPicker_spec.jsx
majapw: Yeah, at Airbnb we use jsdom afaict and `mocha-wrap` (https://github.com/ljharb/mocha-wrap) plugin, but I think there was some discussion on what we plan to do for our open-source repos. @ljharb do you have thoughts on getting enzym's `mount` to work in this repo? I haven't really thought about it, but we should really open an issue I guess.

majapw:
Hey @travisbloom, this is looking good, but I'm a bit confused about the constructor logic you've included.

test/components/DayPicker_spec.jsx
ljharb: In order to do that, we'd need to run the test suite twice - once with jsdom, and once without it - and be able to mark tests that require the DOM such that they'd be skipped when running without jsdom.

I don't yet have a clean solution for this; i think it's OK to skip those tests for now.

travisbloom:
@majapw my understanding was we only want `initialVisibleMonth` fn to trigger once, when the calendar opens initially. The constructor logic ensures that the fn will be called on mount, or if the component is not focused yet, it will call the first time the nextProps has `focused === true`...

Now that I think about it, it would prob be better to trigger it every time it opens...is that the expected behavior you were thinking?

travisbloom:
if thats the case, I'd propose changing `initialVisibleMonth` to `visibleMonthOnFocus`

majapw:
Hmm, let me review again with that in mind. I guess the `initialVisibleMonth` does make sense... I was just misunderstanding the implementation I am pretty sure.

We can maybe augment it in the future to do kind of a fully controlled `visibleMonthOnFocus` situation, but for now I think this is sufficient.

src/components/DateRangePicker.jsx
```diff
@@ -361,6 +364,8 @@ export default class DateRangePicker extends React.Component {
           onNextMonthClick={onNextMonthClick}
           monthFormat={monthFormat}
           withPortal={withPortal || withFullScreenPortal}
+          hidden={!focusedInput}
```
majapw: Can we change this name to `isVisible` instead of `hidden`? I think that is more in line with the other nomenclature in these files.

I think once we move to `react-with-styles`, we hopefully won't need this anyways. since the component will be unmounted/remounted every time.

src/components/DayPicker.jsx
majapw: Can we pull this as an instance method onto the class, named as `getInitialVisibleMonth` or something? That way we just have access to the props straight up.

Actually we don't even call this more than once. Why don't we just have this logic directly in the constructor?

src/components/SingleDatePicker.jsx
majapw: This is a pre-existing prop on the SDP @ljharb

src/components/SingleDatePicker.jsx
```diff
@@ -204,6 +208,8 @@ export default class SingleDatePicker extends React.Component {
           onNextMonthClick={onNextMonthClick}
           monthFormat={monthFormat}
           withPortal={withPortal || withFullScreenPortal}
+          hidden={!focused}
```
majapw: same, let's name this as `isVisible`

test/components/DayPicker_spec.jsx
majapw: Yeah let's skip them.

majapw:
We'll also need to rebase before getting this out! :D Thanks so much @travisbloom

src/components/DayPicker.jsx
travisbloom: @ljharb asked for it to be broken out in to a seperate function in case we need the logic elsewhere, not 100% where that would be.

@majapw let me know which you prefer, getting conflicting suggestions :)

majapw: Oh sorry about that! I think if it's only used once, we should not pull it out. :) Let's go with that.

ljharb: Since it's used in two separate components, my hope was that we'd define it in one place and import it in both.

src/components/DateRangePicker.jsx
travisbloom: :( unfortunately I had it named as visible before but @ljharb asked to change it to be hidden so we didn't have a boolean prop that defaulted to true.

Let me know what you would prefer now.

majapw: Man, I should really have read @ljharb's comments more carefully before. :) `hidden` sounds good to me

travisbloom:
@majapw @ljharb a lot of back and forth on what changes should be made :) rebased and made the requested fixes, let me know if thats inline with what we ended up wanting to do.

.istanbul.yml
```diff
@@ -11,6 +11,6 @@ instrumentation:
 check:
   global:
     statements: 89
-    branches: 70
+    branches: 69
```
travisbloom: inline with https://github.com/airbnb/react-dates/pull/70#discussion_r81176064, updating coverage requirements

majapw:
Great! Thank you for all of this @travisbloom.

Issue #71, opened 2016-09-09
Anchor DateRangePicker and SingleDatePicker to right


travisbloom:
Currently you can not anchor `DRP` and `SDP` to the right of a div.

PR #72, opened 2016-09-09
Add directionAnchor prop to allow right-aligned Components

travisbloom:
DateRangePicker and SingleDatePicker have a new prop, directionAnchor
that will align the components to the left/right side of the page.
This prop has two new constants, DIRECTION_ANCHOR_LEFT (default)
and DIRECTION_ANCHOR_RIGHT.

solves issue https://github.com/airbnb/react-dates/issues/71

### DIFF
--- a/css/DateRangePicker.scss
+++ b/css/DateRangePicker.scss
@@ -21,7 +21,6 @@
   z-index: 1;
   position: absolute;
   top: 145%;
-  left: 0;
 
   &::before,
   &::after {
@@ -30,6 +29,14 @@
     position: absolute;
     bottom: auto;
   }
+}
+
+.DateRangePicker__picker--direction-left {
+    left: 0;
+}
+
+.DateRangePicker__picker--direction-right {
+    right: 0;
 }
 
 .DateRangePicker__picker--show {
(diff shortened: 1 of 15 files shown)


css/DateRangePicker.scss
```diff
@@ -61,13 +68,19 @@
...
+.DateRangePicker__picker--direction-right.DateRangePicker__picker--start::before {
+  right: 239px;
+}
 
-.DateRangePicker__picker--end::before {
+.DateRangePicker__picker--direction-left.DateRangePicker__picker--end::before {
   left: 162px;
 }
+.DateRangePicker__picker--direction-right.DateRangePicker__picker--end::before {
+  right: 101px;
```
travisbloom: unsure if there's a measurement strategy that should be used to calculate these offsets, just tried to eye them. Let me know if there's a calculation you were relying on (and the we can add a `$sass-variable`)

travisbloom:
@majapw fixes added

Olliepop:
What's necessary to see this merged? This would be huge for us. Do you need any help?

majapw:
Looking into it today! I think it's basically ready. :)

travisbloom:
@majapw just fixed a rebase issue with my other PR, let me know if there are any concerns :)

majapw:
I don't really know why the approval check wasn't passing. :/

Olliepop:
Thanks guys, this is awesome! Is this worthy of a version bump?

majapw:
Just released v3.3.0 which should include this change! :)

PR #73, opened 2016-09-10
Disable user-select on week header

judikdavid:
Disabled user-select to prevent text-selection when clicking rapidly.


majapw:
So I checked this locally, and the month headers still get selected with this change when you click on the navigation really quickly. However, https://github.com/airbnb/react-dates/pull/74 seems to fix the problem (at least in Chrome and FF), so I will merge in that PR in lieu of this one. Thank you!

Issue #69, opened 2016-09-09
Very unintuitive text selection happening


isaachinman:
See attached screenshot for an example of what I mean. This behaviour can be seen directly on your [sandbox example](http://airbnb.io/react-dates). Open the picker, click either the forward or backward arrow a few times in quick succession. You'll see either the weekday abbreviations or the months get selected/highlighted.

This remedied by a few lines of CSS, at least for the "default" picker:

    .DayPicker__week-header, .CalendarMonth__caption {
      -webkit-user-select: none; /* Chrome/Safari */
      -moz-user-select: none; /* Firefox */
      -ms-user-select: none; /* IE10+ */
    }

I don't see any advantage to not disabling selection on these kind of "read only" values, and I have already done so in my own projects. Thoughts?

![image](https://cloud.githubusercontent.com/assets/10575782/18390644/50dc8732-76ab-11e6-847f-2a444b5a6cbf.png)

majapw:
Huh, this should have been fixed by https://github.com/airbnb/react-dates/pull/50. Maybe we need the prefixed css selectors for this or it needs to be not on the nav itself but on the week-headers/caption. Would you be open to opening a PR for this?

isaachinman:
> Would you be open to opening a PR for this?

Yeah, definitely. It's a quite easy fix, luckily.

majapw:
Great! This should be fixed on master now.

As for the spec, we don't... really test the CSS right now, so I don't know how we'd go about that, but maybe in the future.

ljharb:
The idea was that people would use autoprefixer in their apps to add the vendor prefix versions automatically. Regardless, we do need the standard one for future compatibility.

isaachinman:
@ljharb Okay, didn't realise that. If autoprefixing is going to happen, it should happen at a project level, not at the end-user level (in my opinion). Anyways, great project!

PR #74, opened 2016-09-10
Disable user-select on arrow buttons (#69)

isaachinman:


majapw:
LGTM

css/DayPicker.scss
```diff
@@ -29,7 +29,9 @@
   color: $react-dates-color-placeholder-text;
   cursor: pointer;
   line-height: 0.78;
-  user-select: none;
+  -webkit-user-select: none; /* Chrome/Safari */
```
ljharb: This removes the standard one in favor the vendor prefixes, which ideally should be handled by something like autoprefixer. At the very least, we must keep the standard one.

majapw: You're totally right. I def overlooked that. Just fixed it on master

Issue #68, opened 2016-09-07
Option to not re-open the date picker when the clear dates button is clicked.


jamhall:
Currently when clicking on the clear button to clear the dates, the dates are cleared and the date picker is re-opened. It would be nice to allow an option for when the close button is clicked, the picker is not re-opened. The property could be called 'reopenPickerOnClear'.

Thank you for a great component!

majapw:
Yeah I think this is a good idea, and something we've actually talked about for certain implementations on the site. This is fairly low on my priority list, but we'd definitely be super happy to take a PR!

agupta93:
@majapw we can close this issue i guess

PR #75, opened 2016-09-11
added hidePickerOnClear prop which when true hides the date picker

agupta93:
This PR is in response to the issue #68 . As requested a prop hidePickerOnClear has been added which governs whether the picker will be shown when clear is clicked. 

By default the value is set to false.

### DIFF
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -43,6 +43,7 @@ const defaultProps = {
   numberOfMonths: 2,
   showClearDates: false,
   disabled: false,
+  reopenPickerOnClearDates: false,
 
   orientation: HORIZONTAL_ORIENTATION,
   withPortal: false,
@@ -233,8 +234,11 @@ export default class DateRangePicker extends React.Component {
   }
 
   clearDates() {
-    this.props.onDatesChange({ startDate: null, endDate: null });
-    this.props.onFocusChange(START_DATE);
+    const { onDatesChange, reopenPickerOnClearDates, onFocusChange } = this.props;
+    onDatesChange({ startDate: null, endDate: null });
+    if (reopenPickerOnClearDates) {
+      onFocusChange(START_DATE);
+    }
   }
 
   doesNotMeetMinimumNights(day) {
--- a/test/components/DateRangePicker_spec.jsx
+++ b/test/components/DateRangePicker_spec.jsx
@@ -154,19 +154,40 @@ describe('DateRangePicker', () => {
   });
 
   describe('#clearDates', () => {
-    describe('props.onFocusChange', () => {
-      it('is called once', () => {
-        const onFocusChangeStub = sinon.stub();
-        const wrapper = shallow(<DateRangePicker onFocusChange={onFocusChangeStub} />);
-        wrapper.instance().clearDates();
-        expect(onFocusChangeStub.callCount).to.equal(1);
-      });
+    describe('props.reopenPickerOnClearDates is truthy', () => {
+      describe('props.onFocusChange', () => {
+        it('is called once', () => {
+          const onFocusChangeStub = sinon.stub();
+          const wrapper = shallow(
+            <DateRangePicker
+              onFocusChange={onFocusChangeStub}
+              reopenPickerOnClearDates
+            />);
+          wrapper.instance().clearDates();
+          expect(onFocusChangeStub.callCount).to.equal(1);
+        });
 
-      it('is called with arg START_DATE', () => {
-        const onFocusChangeStub = sinon.stub();
-        const wrapper = shallow(<DateRangePicker onFocusChange={onFocusChangeStub} />);
-        wrapper.instance().clearDates();
-        expect(onFocusChangeStub.getCall(0).args[0]).to.equal(START_DATE);
+        it('is called with arg START_DATE', () => {
+          const onFocusChangeStub = sinon.stub();
+          const wrapper = shallow(
+            <DateRangePicker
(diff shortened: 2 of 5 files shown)


ljharb:
Naming the prop in this way makes it a breaking change - it's be preferred to find a way to name the default false prop so that the reopen behavior is the default.

agupta93:
@ljharb okay will do the same

agupta93:
@ljharb PR updated please check

src/components/DateRangePicker.jsx
```diff
@@ -223,8 +224,11 @@ export default class DateRangePicker extends React.Component {
   }
 
   clearDates() {
-    this.props.onDatesChange({ startDate: null, endDate: null });
-    this.props.onFocusChange(START_DATE);
+    const {onDatesChange, reopenPickerOnClear, onFocusChange} = this.props;
```
ljharb: curly braces need spaces inside them

src/components/DateRangePicker.jsx
```diff
@@ -43,6 +43,7 @@ const defaultProps = {
   numberOfMonths: 2,
   showClearDates: false,
   disabled: false,
+  reopenPickerOnClear: true,
```
ljharb: This won't work either - that's why i said "default false prop" in my comment.

Absence and `false` should always be the same - so we'll have to pick a name for this prop like "don't reopen picker on clear" (but something much better) and leave the default as `false`.

agupta93:
@ljharb made all the changes recommended by you please check

ljharb:
Thanks, LGTM! It'd be great to also add an example for it in the storybook.

agupta93:
@ljharb there is already a case for clear button would modify the same right according to the new feature added.

agupta93:
added the example in storybook also

src/components/DateRangePicker.jsx
```diff
@@ -43,6 +43,7 @@ const defaultProps = {
   numberOfMonths: 2,
   showClearDates: false,
   disabled: false,
+  hidePickerOnClear: false,
```
majapw: If you click the clear button now when the datepicker is open, I don't think this will actually hide the picker so this prop name is a misnomer. Something like `dontReopenPickerOnClear` maybe makes more sense... or even in this case, it might be worth it to have the semver breaking change of `reopenPickerOnClear`. What do you think @ljharb?

majapw:
Thanks for the example. In addition to my comment about the naming, we'll definitely need to add a bit to the `README` about this prop.

We don't have a channel for discussing this code yet, but opening up an issue is probs the easiest way to get in touch.

agupta93:
@majapw yaa i saw your comment, whatever name you guys finalize wont be a major change so once it finalized will change the name and add the corresponding readme regarding it. This would be my first commit to the library and would like to solve other open issues as well. 

I asked for some channel so as to ficilitate some doubts that i might have in the code :)

src/components/DateRangePicker.jsx
ljharb: If we're comfortable with a breaking change, then sure - that might be better.

majapw: I think that would be fine. We can couple the release with https://github.com/airbnb/react-dates/pull/53 for 3.0.0. I think, semantically, `reopenPickerOnClearDates` seems like a better name

agupta93: @ljharb and @majapw so is it final then the prop name would be `reopenPickerOnClearDates ` whose default value would be false?

agupta93:
@majapw and @ljharb as discussed added a prop `reopenPickerOnClearDates` which defaults to false which means that no picker is shown by default on clear

Added a corresponding example in the storybook and also added the README for the same

majapw:
Hi @agupta93! It looks like this is breaking the `onFocusChange` test now (which makes sense). If you could address that and maybe even add a test for this new prop, that would be awesome. :) We can merge it in after the tests change.

Can you also rebase locally and push it up instead of hitting the update branch button? That would help immensely in keeping our tree clean.

agupta93:
@majapw okay will do the same, can you please help me in telling why is it breaking in OnFocusChange test and as i am a bit new to this how to add test for this new prop. 

am new to this so require help in these small things :)

agupta93:
@majapw corrected the build error and added the test for the same please check :)

test/components/DateRangePicker_spec.jsx
```diff
@@ -154,17 +154,32 @@ describe('DateRangePicker', () => {
   });
 
   describe('#clearDates', () => {
+    describe('props.reopenPickerOnClearDates', () => {
```
majapw: So probably the structure of these tests should look something like:
```
describe('#clearDates', () => {
  describe('props.reopenPickerOnClearDates is truthy', () => {
    describe('props.onFocusChange', () => {
      it('is called once', () => {
        ...
      });
      
      it('is called with arg START_DATE', () => {
        ...
      });
  });

  describe('props.reopenPickerOnClearDates is falsey', () => {
    describe('props.onFocusChange is not called', () => {
      ...
    });
  });
});
```

Some of these tests are duplicates right now.

majapw:
Right now the travis build is failing on some linting rules! You should address those and also run `git rebase origin/master` before you push back up to your remote branch. ✌️

agupta93:
@majapw all changes done as asked hope everything looks fine now ✌️

agupta93:
@majapw commits squashed !!

jamhall:
@agupta93 ! Thank you!

PR #76, opened 2016-09-12
Add week number support

judikdavid:
Adds week number in front of every week line in the first column.
It uses a new withWeekNumbers prop on every component, that toggles the
feature.
It supports both horizontal and vertical datepickers.

### DIFF
--- a/css/CalendarMonth.scss
+++ b/css/CalendarMonth.scss
@@ -43,11 +43,6 @@
   cursor: default;
   width: 39px;
   height: 38px;
-
-  user-select: none;
-  -moz-user-select: none;
-  -webkit-user-select: none;
-  -ms-user-select: none;
 }
 
 // This order is important.
--- a/css/CalendarMonth.scss
+++ b/css/CalendarMonth.scss
@@ -41,8 +41,8 @@
   box-sizing: border-box;
   color: $react-dates-color-gray;
   cursor: default;
-  width: 39px;
-  height: 38px;
+  width: $react-dates-width-day;
+  height: $react-dates-height-day;;
 }
 
 // This order is important.
@@ -52,8 +52,8 @@
   box-sizing: border-box;
   color: $react-dates-color-gray;
   cursor: pointer;
-  width: 39px;
-  height: 38px;
+  width: $react-dates-width-day;
+  height: $react-dates-height-day;;
 
   &:active {
     background: darken($react-dates-color-white, 5%);
--- a/css/CalendarMonth.scss
+++ b/css/CalendarMonth.scss
@@ -36,6 +36,20 @@
   margin-bottom: 2px;
 }
 
+.CalendarMonth__weeknumber {
+  padding: 0;
+  box-sizing: border-box;
+  color: $react-dates-color-gray;
+  cursor: default;
+  width: 39px;
+  height: 38px;
+
+  user-select: none;
+  -moz-user-select: none;
+  -webkit-user-select: none;
+  -ms-user-select: none;
+}
+
 // This order is important.
 .CalendarMonth__day {
   border: 1px solid lighten($react-dates-color-border-light, 3);
--- a/src/components/CalendarMonth.jsx
+++ b/src/components/CalendarMonth.jsx
@@ -9,7 +9,7 @@ import getCalendarMonthWeeks from '../utils/getCalendarMonthWeeks';
 
 import OrientationShape from '../shapes/OrientationShape';
 
-import { HORIZONTAL_ORIENTATION, VERTICAL_ORIENTATION } from '../constants';
+import { HORIZONTAL_ORIENTATION, VERTICAL_ORIENTATION } from '../../constants';
 
 const propTypes = {
   month: momentPropTypes.momentObj,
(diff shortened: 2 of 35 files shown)


majapw:
Can we see some screenshots?

judikdavid:
Of course, see below:
screen shot 2016-09-12 at 22 11 19
screen shot 2016-09-12 at 22 11 38

css/CalendarMonth.scss
```diff
@@ -36,6 +36,20 @@
...
 
+.CalendarMonth__weeknumber {
+  padding: 0;
+  box-sizing: border-box;
+  color: $react-dates-color-gray;
+  cursor: default;
+  width: 39px;
+  height: 38px;
+
+  user-select: none;
```
majapw: Can you test that this `user-select` is still necessary given https://github.com/airbnb/react-dates/pull/74 and follow up commit?

src/components/CalendarMonth.jsx
```diff
@@ -53,11 +55,26 @@ export function getModifiersForDay(modifiers, day) {
   return day ? Object.keys(modifiers).filter(key => modifiers[key](day)) : [];
 }
 
+export function getWeekNumberForWeek(week) {
```
majapw: Can we pull this into its own `utils` file a la `getsCalendarMonthWeeks`?

src/components/CalendarMonth.jsx
```diff
@@ -53,11 +55,26 @@ export function getModifiersForDay(modifiers, day) {
...
+  if (week && Array.isArray(week)) {
+    const validDays = week.filter((d) => (d));
+    if (validDays[0] && moment.isMoment(validDays[0])) {
+      const firstDayOfWeek = validDays[0].clone().startOf('week');
+
+      return firstDayOfWeek.week();
+    }
+    return null;
```
majapw: probs we just need the return at the end and not this one, right?

src/components/CalendarMonth.jsx
```diff
@@ -53,11 +55,26 @@ export function getModifiersForDay(modifiers, day) {
   return day ? Object.keys(modifiers).filter(key => modifiers[key](day)) : [];
 }
 
+export function getWeekNumberForWeek(week) {
+  if (week && Array.isArray(week)) {
```
majapw: You seem to be doing a whole lot of defensive checks here that I'm not sure are necessary. `week` is always going to be of a v. specific form and I also don't know that you need to clone the moment objects or call startOf('week') in this way.

src/components/CalendarMonthGrid.jsx
```diff
@@ -60,6 +64,19 @@ const defaultProps = {
   monthFormat: 'MMMM YYYY', // english locale
 };
 
+function getCalendarMonthGridWidthStyle(orientation, withWeekNumbers) {
```
majapw: :| I really want to work on the with-styles change more now cos this sucks that we have to do this.

css/CalendarMonth.scss
judikdavid: Yes, you're right, it's not necessary, deleting it in a next commit.

src/components/CalendarMonth.jsx
judikdavid: Fixing it in a next commit.

src/components/CalendarMonth.jsx
judikdavid: Yes, I overcomplicated things, fixing it.

src/components/CalendarMonth.jsx
judikdavid: Right.

src/components/CalendarMonthGrid.jsx
judikdavid: Yes, I also think that this is not the best solution, but at least the width value is not duplicated in css and js this way.

css/CalendarMonth.scss
```diff
@@ -36,6 +36,15 @@
   margin-bottom: 2px;
 }
 
+.CalendarMonth__weeknumber {
+  padding: 0;
+  box-sizing: border-box;
+  color: $react-dates-color-gray;
+  cursor: default;
+  width: 39px;
```
majapw: Maybe we should pull these values out into a variable so we can share between the `.CalendarMonth__day` class and this one

src/components/CalendarMonthGrid.jsx
majapw: Yeah that totally makes sense! :) Can we pull this out as well into a util method? I'm very much about small utility modules wherever possible. It makes testing a dream.

majapw:
Whoa this new github review flow is fantastic.

jontelm:
Any updates on this one? @judikdavid I want this stuff :)

majapw:
I'm going to close this as this still had some issues with animations at the time of last commit and now will be in conflict hell if we try to reopen it. We can re-entertain this issue in the future if desired.

grenhall:
I'd say this feature is highly requested from everyone in the Nordic region.

Issue #80, opened 2016-09-13
DateRangePicker initialises itself outside viewport width


isaachinman:
As you can seen in the screenshot below, the DateRangePicker component seems to (sometimes) initalise itself without regard to viewport width. In my case, it is actually placing itself in such a way that my entire page breaks (the dark edge is the edge of my viewport, were it not to be expanded by the DateRangePicker).

No, I am not initialising it in a hidden component or anything like that - I am using it completely as intended/expected.

This happens even at screen widths as high as 1000px, if the input is placed close enough to the edge.

In fact, this impacts my entire page _even when the datepicker is closed_, because of the use of `visibility: hidden` instead of `display: none`.

Sorry for not including a live example, but I haven't yet found a decent sandbox to use for npm/es6 stuff.

![screen shot 2016-09-13 at 13 28 52](https://cloud.githubusercontent.com/assets/10575782/18472219/5dd050dc-79b6-11e6-94f9-c9e1793ba1a6.png)

majapw:
This is related to https://github.com/airbnb/react-dates/issues/71 (although that solution I think is only a stopgap).

Ideally we should have a totes responsive component that can see if it is overflowing and adjust itself accordingly. That's not there yet (soon! well, relatively). In the mean time you can override the position on the `.DateRangePicker__picker` (originally set at https://github.com/airbnb/react-dates/blob/master/css/DateRangePicker.scss#L24) in your own stylesheet. You'll probably want to override the `:before` and `:after` positioning styles as well (https://github.com/airbnb/react-dates/blob/master/css/DateRangePicker.scss#L56:L86).

https://github.com/airbnb/react-dates/blob/master/css/DateRangePicker.scss#L56:L86).

This is a bummer:
> In fact, this impacts my entire page even when the datepicker is closed, because of the use of `visibility: hidden` instead of `display: none`.

We use `visibility` over `display` right now so that we can do a height calculation while the datepicker is still closed but the DOM has been rendered.  I want to move away from this... but I think that will have to wait until we manage to get a `react-with-styles` implementation out (which is tricky, but in the works).

isaachinman:
I don't think this the sort of thing one could fix with CSS overrides, and to be honest this library is not production ready until this is addressed.

Lots and lots of similar kinds of components exist out in the wild that set themselves appropriately!

majapw:
Fixed in v3.3.2

PR #83, opened 2016-09-13
Responsivizes the DayPicker container

majapw:
This PR addresses https://github.com/airbnb/react-dates/issues/80.

This pulls in http://tether.io/ which is a sweet library (thanks for the rec, @lencioni!). Basically it pins the DayPicker component to its scrollable parents on the right side and updates this tether if you resize the window.

There are a few things left to do:
- [X] get the `OutsideClickHandler` working properly again. The tether restructures the DOM in such a way that the picker itself no longer is a child of the OutsideClickHandler.
- [X] Fix the nubs so that they align with the inputs again. 🎉 (https://github.com/airbnb/react-dates/pull/118)
- [X] Apply the same changes to the SDP

to: @lencioni @isaachinman

### DIFF
--- a/css/DateInput.scss
+++ b/css/DateInput.scss
@@ -14,8 +14,8 @@
   vertical-align: middle;
 }
 
-.DateInput--focused::before,
-.DateInput--focused::after {
+.DateInput--with-caret::before,
+.DateInput--with-caret::after {
   content: "";
   display: inline-block;
   position: absolute;
@@ -26,12 +26,12 @@
   z-index: 2;
 }
 
-.DateInput--focused::before {
+.DateInput--with-caret::before {
   left: 22px;
   border-bottom-color: rgba(0, 0, 0, 0.1);
 }
 
-.DateInput--focused::after {
+.DateInput--with-caret::after {
   left: 23px;
   border-bottom-color: $react-dates-color-white;
 }
--- a/css/DateRangePicker.scss
+++ b/css/DateRangePicker.scss
@@ -17,10 +17,7 @@
 }
 
 .DateRangePicker__picker {
-  background-color: $react-dates-color-white;
   z-index: 1;
-  position: absolute;
-  top: $react-dates-spacing-vertical-picker;
 }
 
 .DateRangePicker__picker--direction-left {
--- a/css/SingleDatePicker.scss
+++ b/css/SingleDatePicker.scss
@@ -6,52 +6,15 @@
 }
 
 .SingleDatePicker__picker {
-  background-color: $react-dates-color-white;
   z-index: 1;
-  position: absolute;
-  top: 145%;
-
-  &::before,
-  &::after {
-    content: "";
-    display: inline-block;
-    position: absolute;
-    bottom: auto;
-  }
-
-  &::before {
-    top: -10px;
-    border: 10px solid transparent;
-    border-top: 0;
-    border-bottom-color: rgba(0, 0, 0, 0.1);
-  }
-
-  &::after {
-    top: -9px;
-    border: 9px solid transparent;
-    border-top: 0;
-    border-bottom-color: #fff;
-  }
 }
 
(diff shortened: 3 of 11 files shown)


isaachinman:
Wow, that was way quicker than I expected! Are you handling those three remaining tasks yourself, or saying you need help?

majapw:
It'll take a bit for me to fully get to it, but either is good. This step #1 was a proof of concept. @isaachinman if you would be interested in continuing the work, we would def be happy to accept the PR. :) I think the `OutsideClickHandler` work and `nub` functionality might be a bit tricky.

isaachinman:
@majapw Trying to get your branch up and running so I can see how much work this will involve, but having CSS problems. No matter what `npm` script I run, I can't get `storybook.css` inside the `public` dir to be generated. Any advice?

majapw:
@isaachinman What do mean by generating the `storybook.css` file? That file is static and all it really does is set the font to `sans-serif`, change the `box-sizing` and add the background grid to the storybook. Is the css not loading for the datepicker itself? I'm assuming you've done the standard `npm install` and are running storybook using the npm script, yeah? Are you getting any errors?

isaachinman:
Ah, I misunderstood then. Yes I've run `npm i` as well as the build script. The picker seems to be initialising itself upward, and thus mostly not-visible. See screenshot below. I assumed this was a CSS positioning thing, but maybe not?

screen shot 2016-09-15 at 18 07 13

majapw:
Well it's certainly weird that the `storybook.css` file seems not to be loading at all, but the css for the datepicker itself seems to be loading (as indicated by the input styling). This is def an issue with the tether set-up itself WHICH I SWEAR TO GOD WAS WORKING WHEN I PUSHED THIS. :| Let me investigate as I can repro.

majapw:
@isaachinman So in the setup:
```
    this.tether = new Tether({
      element: this.dayPickerContainer,
      target: ReactDOM.findDOMNode(this.input),
      attachment: 'top left',
      targetAttachment: 'bottom left',
      offset: '-20px 0',
      constraints: [{
        to: 'scrollParent',
        attachment: 'together',
        pin: ['right'],
      }],
    });
```

I think the `to: 'scrollParent',` is the relevant part and if we change that to `window`, we get much better behavior. HOWEVER, you don't get like the expected body padding and the like. So `scrollParent` would be ideal. Examining the `<body>` tag that would constitute the datepicker's scroll parent in this scenario, we see that it only has a height of 50px and so the tether is behaving in the way it would if the window was too short. ... of course, it shouldn't be behaving this way because the pinning should only be happening on the right side of the scroll parent. Also maybe more important, I swear that this was working before and idk what the H happened here. :|

thj-dk:
What's the status on this PR?

thj-dk:
@majapw any ideas how to solve the `OutsideClickHandler` related problem?

majapw:
Sorry I have been traveling and it has significantly impacted my work here. 

@thj-dk my initial thought on the OutsideClickHandler issue would be just to wrap the input components and daypicker components separately and then when those are used with the tether the outside click logic is including in the pieces that are moved around in the dom individually.

majapw:
Okay so I think this works! 

I just put the `OutsideClickHandler` on the `DayPicker` div _always_ to solve the close on outside click problem. This has the disadvantage that when you click on an date input, it calls setState on the focus state twice (once in the outside click handler and once in onFocus), but I think that is okay for our usecase.
 
Here are kind of the last bits to do (but they're small):
- [ ] Update tests to look for the Tether and not the OutsideClickHandler
- [ ] Change the edge the datepicker is pinned to based on anchor direction?
- [ ] Double-check all the difference variations of the datepickers (including the portal versions)

And then we'll be good to go!

majapw:
Cool! I think this is ready! :D Just needs to wait on https://github.com/airbnb/react-dates/pull/118 to be merged in.

PR #97, opened 2016-09-26
Updated toMomentObject to take advantage of moment format array

timReynolds:
Moment can take an array of formats as outlined in
http://momentjs.com/docs/#/parsing/special-formats/

Using this makes the toMomentObject util easier to understand as it
reduces if return statements within the method.

All tests still pass with the refactor


majapw:
LGTM as well!

PR #98, opened 2016-09-27
Fixes issues with displayFormat prop's default values by using a thunk

majapw:
to: @ljharb 

When attempting to bump the version of `react-dates` used on Airbnb, we noticed that the display format for the `DateRangePicker` and `SingleDatePicker` components was often returning nonsense because moment's locale was not yet set. Similarly, live right now on the storybook, you can see that the format doesn't work at all for the non-english locale example. This fixes this issue.

### DIFF
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -55,7 +55,7 @@ const defaultProps = {
   onNextMonthClick() {},
 
   // i18n
-  displayFormat: moment.localeData().longDateFormat('L'),
+  displayFormat: () => moment.localeData().longDateFormat('L'),
   monthFormat: 'MMMM YYYY',
   phrases: {
     closeDatePicker: 'Close',
@@ -137,8 +137,9 @@ export default class DateRangePicker extends React.Component {
   }
 
   onEndDateChange(endDateString) {
-    const { startDate, isOutsideRange, onDatesChange, onFocusChange, displayFormat } = this.props;
-    const endDate = toMomentObject(endDateString, displayFormat);
+    const { startDate, isOutsideRange, onDatesChange, onFocusChange } = this.props;
+
+    const endDate = toMomentObject(endDateString, this.getDisplayFormat());
 
     const isEndDateValid = endDate && !isOutsideRange(endDate) &&
       !isInclusivelyBeforeDay(endDate, startDate);
@@ -174,8 +175,7 @@ export default class DateRangePicker extends React.Component {
   }
 
   onStartDateChange(startDateString) {
-    const { displayFormat } = this.props;
-    const startDate = toMomentObject(startDateString, displayFormat);
+    const startDate = toMomentObject(startDateString, this.getDisplayFormat());
 
     let { endDate } = this.props;
     const { isOutsideRange, onDatesChange, onFocusChange } = this.props;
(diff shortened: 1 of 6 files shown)


src/components/DateRangePicker.jsx
```diff
@@ -138,7 +138,9 @@ export default class DateRangePicker extends React.Component {
 
   onEndDateChange(endDateString) {
     const { startDate, isOutsideRange, onDatesChange, onFocusChange, displayFormat } = this.props;
-    const endDate = toMomentObject(endDateString, displayFormat);
+
+    const formatString = typeof displayFormat === 'function' ? displayFormat() : displayFormat;
```
ljharb: the propTypes help ensure it's only a string or function, but perhaps it would be more robust (ie, fail faster when it's not a func or string) if this did `typeof displayFormat === 'string' ? displayFormat : displayFormat();`?

src/components/DateRangePicker.jsx
```diff
@@ -175,7 +177,8 @@ export default class DateRangePicker extends React.Component {
 
   onStartDateChange(startDateString) {
     const { displayFormat } = this.props;
-    const startDate = toMomentObject(startDateString, displayFormat);
+    const formatString = typeof displayFormat === 'function' ? displayFormat() : displayFormat;
```
ljharb: also here. should this logic be consolidated into a helper function?

src/components/DateRangePicker.jsx
```diff
@@ -203,8 +206,9 @@ export default class DateRangePicker extends React.Component {
 
   getDateString(date) {
     const { displayFormat } = this.props;
-    if (date && displayFormat) {
-      return date && date.format(displayFormat);
+    const formatString = typeof displayFormat === 'function' ? displayFormat() : displayFormat;
```
ljharb: `^_^`

src/components/SingleDatePicker.jsx
```diff
@@ -68,7 +68,8 @@ export default class SingleDatePicker extends React.Component {
 
   onChange(dateString) {
     const { displayFormat, isOutsideRange, onDateChange, onFocusChange } = this.props;
-    const date = toMomentObject(dateString, displayFormat);
+    const formatString = typeof displayFormat === 'function' ? displayFormat() : displayFormat;
```
ljharb: ᕕ( ᐛ )ᕗ

src/components/SingleDatePicker.jsx
```diff
@@ -114,8 +115,9 @@ export default class SingleDatePicker extends React.Component {
 
   getDateString(date) {
     const { displayFormat } = this.props;
-    if (date && displayFormat) {
-      return date && date.format(displayFormat);
+    const formatString = typeof displayFormat === 'function' ? displayFormat() : displayFormat;
```
ljharb: there's a lot of these :-p

src/components/DateRangePicker.jsx
majapw: Okay. I wasn't sure if it made more sense to have it fail abruptly in this way or fail silently, but still function, in the other. I can change it to the former.

src/components/DateRangePicker.jsx
majapw: Yes

src/components/SingleDatePicker.jsx
majapw: Lol

src/components/DateRangePicker.jsx
ljharb: i'm a big fan of failing fast and loud

PR #99, opened 2016-09-28
Fix lint errors across the repo

majapw:
This removes the eslint exceptions and updates the repo to follow these new conventions. 

to: @airbnb/webinfra


src/components/CalendarMonthGrid.jsx
```diff
@@ -154,7 +156,7 @@ export default class CalendarMonthGrid extends React.Component {
 
     return (
       <div
-        ref={ref => { this.containerRef = ref; }}
+        ref={ref => { this.container = ref; }}
```
ljharb: the linter should require this arrow arg have parens

src/components/DateInput.jsx
```diff
@@ -106,7 +104,7 @@ export default class DateInput extends React.Component {
           type="text"
           id={id}
           name={id}
-          ref={(ref) => { this.inputRef = ref; }}
+          ref={(ref) => { this.input = ref; }}
```
ljharb: using arrows as ref callbacks (instead of bound instance methods) means it can sometimes get null. I think we should use a bound instance method instead, throughout.

majapw: What does that pattern look like? Something like:
```
constructor(props) {
  super(props);
  this.setInputRef = this.setInputRef.bind(this);
}

setInputRef(ref) {
  this.input = ref;
}

render() {
  return (
    <div ref={this.setInputRef} />
  );
}
```

majapw:
Yeah def, I'll update those deps in this pr as well.

src/components/DateInput.jsx
ljharb: Exactly like that!

.eslintrc
```diff
@@ -5,9 +5,8 @@
     "browser": true,
     "node": true
   },
-
   "rules": {
-    "react/no-find-dom-node": 1,
-    "react/no-string-refs": 1
+    "react/no-unused-prop-types": 1,
+    "jsx-a11y/no-static-element-interactions": 1
```
majapw: So `no-unused-prop-types` is broken which is why that's a warn right now (nested prop types don't check properly). Um, rn I don't know if I should go in and change the interactive elements in the datepicker to be buttons as they should be given that I don't feel like that would be a good keyboard navigable experience yet.

majapw:
PTAL @ljharb I addressed your changes. 

On the other hand, now that we've upgraded I'm ignoring two new rules so uh, basically nothing happened overall.

src/components/CalendarDay.jsx
```diff
@@ -89,13 +89,13 @@ export default class CalendarDay extends React.Component {
...
       <div
         className="CalendarDay"
-        onMouseEnter={(e) => this.handleDayMouseEnter(day, modifiers, e)}
-        onMouseLeave={(e) => this.handleDayMouseLeave(day, modifiers, e)}
-        onMouseDown={(e) => this.handleDayMouseDown(day, modifiers, e)}
-        onMouseUp={(e) => this.handleDayMouseUp(day, modifiers, e)}
-        onClick={(e) => this.handleDayClick(day, modifiers, e)}
-        onTouchStart={(e) => this.handleDayTouchStart(day, modifiers, e)}
-        onTouchEnd={(e) => this.handleDayTouchEnd(day, modifiers, e)}
+        onMouseEnter={e => this.handleDayMouseEnter(day, modifiers, e)}
```
ljharb: all these are event handlers, which shouldn't return a value - can they use `{ }` around the body instead?

src/components/DayPicker.jsx
```diff
@@ -69,6 +68,51 @@ const defaultProps = {
 };
 
 export default class DayPicker extends React.Component {
+  static applyTransformStyles(el, transform, opacity = '') {
+    const transformStyles = getTransformStyles(transform);
+    transformStyles.opacity = opacity;
+
+    Object.keys(transformStyles).forEach((styleKey) => {
```
ljharb: `Object.assign(el.style, transformStyles)`?

src/components/DayPicker.jsx
```diff
@@ -69,6 +68,51 @@ const defaultProps = {
...
+      el.style[styleKey] = transformStyles[styleKey];
+    });
+  }
+
+  static calculateDimension(el, axis, borderBox = false, withMargin = false) {
+    if (!el) {
+      return 0;
+    }
+
+    const axisStart = (axis === 'width') ? 'Left' : 'Top';
```
ljharb: these parens are unnecessary

majapw:
@ljharb PTAL

ljharb:
LGTM

majapw:
Gonna try again in a different PR

PR #101, opened 2016-09-29
Caption mis-positioned with Boostrap4 #100

### LINKED ISSUE #100: Caption mis-positioned with Boostrap4
ss

```scss
/* TEMP FIX */
.CalendarMonth {
  table {
    caption: {
      caption-side: top;
    }
  }
}
.CalendarMonth__caption {
  caption-side: top;
}
```

zeroasterisk:
fixes #100  worked for me locally.

problems occured when combined with bootstrap4:
https://github.com/twbs/bootstrap/blob/a27a488002aafb6725e80f63d7022f9af32df669/dist/css/bootstrap-reboot.css#L386

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Change Log
 
+## v7.0.0
+- [breaking] Simplify `CalendarDay` DOM ([#291](https://github.com/airbnb/react-dates/pull/291))
+
 ## v6.1.0
 - [fix] Revert "Simplify `CalendarDay` DOM ([#291](https://github.com/airbnb/react-dates/pull/291))"
 - [new] Add `renderDay` prop to customize the content of the `CalendarDay` component ([#307](https://github.com/airbnb/react-dates/pull/307))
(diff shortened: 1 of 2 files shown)


css/CalendarMonth.scss
```diff
@@ -14,6 +14,9 @@
   table {
     border-collapse: collapse;
     border-spacing: 0;
+    caption: {
+      caption-side: top;
```
ljharb: should these use `initial` to override bootstrap, rather than specifying "top"?

zeroasterisk: maybe... but wouldn't anything other than `top` break your styling?

ljharb: Since we don't currently specify a caption-top, I'd think we want the browser default.

bsr203: `initial` also worked for me. thanks

majapw:
Hmm, I think I can't push to your master branch I'll reopen these changes with comments addressed in a separate PR.

PR #102, opened 2016-09-29
Sets up JS Code Coverage

majapw:
So it's running now, yay! And the current percentages are set as hard thresholds.

to: @ljharb @airbnb/webinfra


ljharb:
What are the thresholds?

majapw:
Well that's exciting. Um. I don't know how to set thresholds but I will look into that.

lencioni:
LGTM

.travis.yml
```diff
@@ -13,6 +13,9 @@ script:
   - 'if [ -n "${LINT-}" ]; then npm run lint ; fi'
   - 'if [ -n "${TEST-}" ]; then npm run tests-only ; fi'
   - 'if [ -n "${KARMA-}" ]; then npm run tests-karma ; fi'
+  - 'if [ -n "${REACT-}" ]; then npm run cover; fi'
+after_success:
+  - 'if [ "${TRAVIS_NODE_VERSION}" = "4.2" ]; then cat ./coverage/lcov.info | ./node_modules/.bin/coveralls ; fi'
```
ljharb: this seems like it runs coverage and applies thresholds to each react version - rather than merging all coverage results and applying thresholds to *that*. would https://www.npmjs.com/package/istanbul-merge help?

majapw: I'm not sure how you would integrate that into these scripts.

PR #103, opened 2016-09-30
Force minimum dates to affect the current month

benbayard:


majapw:
Can you include a screenshot of before and after so that we can better understand the problem?

majapw:
Also, I'm trying to get this change pushed out ASAP: https://github.com/airbnb/react-dates/pull/70 Do you think that would be able to address your problem without adding knowledge of startDate and min nights to the `DayPicker` component?

benbayard:
I'll absolutely add a screenshot. 

`startDate` as far as I can derive must be required. Because I have to know how far out from the start date I would need to scroll the initial month so-as to not see the right month as empty. This way the user does not have to guess how far they must scroll to find an acceptable day.

majapw:
Yeah, I think I understand the concern. Plus the `initialVisibleMonth` PR is mostly targeted at setting the month when the `DayPicker` is first opened. 

I think my biggest concern about this PR right now is that there's no slick animation for this kind of transition so it would just flash to that month. I def see the need for this but am unsure if this is the best way to implement it.

It's starting to seem like maybe the DayPicker month should be controlled in general. :/ Hmm.

benbayard:
Can we expose the function that does the animation and have initial month
be controlled by props?

benbayard:
Also, I can make this functionality optional until the animation can be
exposed, if that works. I agree that the UX isn't ideal.

benbayard:
Can we expose the function that does the animation and have initial month
be controlled by props? Also, I can make this functionality optional until the animation can be
exposed, if that works. I agree that the UX isn't ideal.

The biggest issue currently is that the UX for minimumNights of 1 "month" makes it look like the dates are selectable and that there is nothing to select. The animation currently does not have an API users can access and since this is controlled by a prop it further complicates the situation. 

Perhaps the best way to handle it might be to say something like:

```jsx
onDatesChange(start, end) {
  const {minimumNights, onDayClick} = this.props;
  if (minimumNightsExceedsCurrentlyVisibleMonths) {
    this.refs.calendarMonthGrid.transitionToDay(start + minimumNights);
  } 
  onDayClick && onDayClick(start, end);
}
```

This requires a few things I do not like. Is there another way to hook in to the animation? Perhaps through firstVisibleMonthIndex?

benbayard:
@majapw I found a way to do this using the existing animation API. If you like how this works and the general logic, I'll add some tests and make changes per your recommendations.

ljharb:
Unfortunately this PR's fork was deleted, so the PR can never be landed.

@benbayard if you're still interested in landing this fix, a new PR would be appreciated.

Issue #110, opened 2016-10-10
Range Picker incorrectly highlights 1st April when hovering over an end date of 31st March


timReynolds:
When using the date range picker if you select any start date and then hover over the 31st March then the 1st April will also be highlighted. 

This behaviour can be seen on any year, but isn't seen for any other month. 

I've tested this on the public React story board were the behaviour can also be seen. 

It has no functional impact that I could tell. 

![image](https://cloud.githubusercontent.com/assets/168870/19242453/1cf10a2e-8f0b-11e6-8aaa-ccbf3ea9aac3.png)

timReynolds:
This issue is caused by UTC offset, after some investigation isSameDay is returning true as day has an offset of one. 

Output from logic outlined below;

day **2017-04-01T00:00:00+01:00**
startDate 2017-03-15T00:00:00Z
hoverDate 2017-03-31T00:00:00Z
!!startDate && !endDate && (day.isBetween(startDate, hoverDate) false
**isSameDay(hoverDate, day)) true**
**isForwardRange true**

timReynolds:
@ljharb these aren't the only issue I've seen because of UTC offsets which I assume, but am not convinced by, are actually caused by the addition of DST. Calculating isOutsideRange has similar issues across DST from memory. 

I started to look into this in more depth but didn't have enough knowledge about the functionality that would be affected. If this is DST related would setting dates to have time stamps of noon not be a simple solution? Most of the issues I've seen is causes by having a date with DST calculated against one without.

majapw:
What is confusing to me... is that all of these days (`hoverDate`, `startDate`, `day`) should be calculated with the same utc offset, since they're all ultimately coming from the `getCalendarMonthWeeks` method. The March 31st/April 1st crossover is not a DST date in any year (in any country as far as I can tell) so why is the UTC offset changing across this border? 

We can change the time stamp to noon, but I am still confused.

alexanbj:
@majapw See my comment in https://github.com/airbnb/react-dates/issues/111#issuecomment-252827168. I think that is what causes this problem.

timReynolds:
@majapw I agree this isn't a cross over for DST but I couldn't work out what else would cause the issue. 

I've had the same thoughts as @alexanbj about not setting the UTC offset, which feels like it would solve many of these issues. But I believe adding UTC offsets fixes at least one other issue.

majapw:
[...]

PR #114, opened 2016-10-11
Modifies all dates in the calendar to have their time set to noon

majapw:
This is a fix for https://github.com/airbnb/react-dates/issues/110

I also double-checked this change in the Brazilian time zone and noted that there is no reoccurrence of the DST bug that caused there to be an extra day in the month of October. It might be good to double-check Australia as well.

I'm not like incredibly happy about this, but it seems to work. :|

to: @timReynolds @ljharb 
FYI: @airbnb/webinfra @wmartins

### DIFF
--- a/test/utils/getCalendarMonthWeeks_spec.js
+++ b/test/utils/getCalendarMonthWeeks_spec.js
@@ -1,6 +1,7 @@
 import moment from 'moment';
 import { expect } from 'chai';
 
+import isSameDay from '../../src/utils/isSameDay';
 import getCalendarMonthWeeks from '../../src/utils/getCalendarMonthWeeks';
 
 const today = moment();
@@ -25,6 +26,40 @@ describe('getCalendarMonthWeeks', () => {
     });
 
     expect(isIncluded).to.equal(true);
+  });
+
+  it('all days have a time of 12PM', () => {
+    weeks.forEach((week) => {
+      week.forEach((day) => {
+        if (day) {
+          expect(day.hours()).to.equal(12);
+        }
+      });
+    });
+  });
+
+  describe('Daylight Savings Time issues', () => {
+    it('last of February does not equal first of March', () => {
+      const february = getCalendarMonthWeeks(today.clone().month(1));
+      const lastWeekOfFebruary = february[february.length - 1].filter(Boolean);
+      const lastOfFebruary = lastWeekOfFebruary[lastWeekOfFebruary.length - 1];
+
+      const march = getCalendarMonthWeeks(today.clone().month(2));
+      const firstOfMarch = march[0].filter(Boolean)[0];
+
+      expect(isSameDay(lastOfFebruary, firstOfMarch)).to.equal(false);
+    });
+
+    it('last of March does not equal first of April', () => {
+      const march = getCalendarMonthWeeks(today.clone().month(2));
+      const lastWeekOfMarch = march[march.length - 1].filter(Boolean);
+      const lastOfMarch = lastWeekOfMarch[lastWeekOfMarch.length - 1];
+
+      const april = getCalendarMonthWeeks(today.clone().month(3));
+      const firstOfApril = april[0].filter(Boolean)[0];
+
+      expect(isSameDay(lastOfMarch, firstOfApril)).to.equal(false);
+    });
   });
 
   describe('enableOutsideDays arg is false', () => {
--- a/test/utils/toMomentObject_spec.js
+++ b/test/utils/toMomentObject_spec.js
@@ -1,6 +1,7 @@
 import moment from 'moment';
 import { expect } from 'chai';
 
+import isSameDay from '../../src/utils/isSameDay';
 import toMomentObject from '../../src/utils/toMomentObject';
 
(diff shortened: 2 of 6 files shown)


majapw:
WAIT I GET IT.

majapw:
just kidding, thanks for making me write tests @ljharb. The original code didn't actually do what I expected. :) This should fix things. PTAL!

test/utils/toMomentObject_spec.js
```diff
@@ -20,6 +20,10 @@ describe('toMomentObject', () => {
     expect(toMomentObject()).to.equal(null);
   });
 
+  it('output has time of 12PM', () => {
```
ljharb: it might be nice to test the actual conditions - like, that the thing resulting in the extra day being highlighted returns the correct result

majapw: This becomes a bit harder to test I think because it... depends on a lot of different factors. I mean, I guess we can check that the first of april/last of march and first of march/last of april do not make isSameDay true but that seems like a bad test?

wmartins:
Hey @majapw, I was just fiddling with the code here and I think I've found a simple (and reliable) solution for this problem.

First of all, the `utcOffset` fix was done to fix the strange behavior of [`moment#add`](https://github.com/airbnb/react-dates/blob/73b3ff689eb627197bb67ddb0ec2853b64b735c0/src/utils/getCalendarMonthWeeks.js#L19), that was adding an extra day when there is a timezone change (i.e.: Brazil entering in daylight saving time). The problem is that this `utc` date is being propagated to more parts of the library, and this is causing the problem, because every other method is now using this "mutated" object (causing the problem with `isSameDay` as reported by @timReynolds).

So, I think that if you do the reverse process when using the day, you'll be "safe".

In this case, there are just three simple changes.

The first is in [`prevDay`](https://github.com/airbnb/react-dates/blob/73b3ff689eb627197bb67ddb0ec2853b64b735c0/src/utils/getCalendarMonthWeeks.js#L13):

```javascript
// old
const prevDay = enableOutsideDays && currentDay.clone().subtract(i + 1, 'day');
// new
const prevDay = enableOutsideDays && currentDay.clone().subtract(i + 1, 'day').local();
```

The second is in [`currentWeek.push`](https://github.com/airbnb/react-dates/blob/73b3ff689eb627197bb67ddb0ec2853b64b735c0/src/utils/getCalendarMonthWeeks.js#L18)

```javascript
// old
currentWeek.push(currentDay.clone());
// new
currentWeek.push(currentDay.clone().local());
```

And the third in [`nextDay`](https://github.com/airbnb/react-dates/blob/73b3ff689eb627197bb67ddb0ec2853b64b735c0/src/utils/getCalendarMonthWeeks.js#L32):

```javascript
// old
const nextDay = enableOutsideDays && currentDay.clone().add(count, 'day');
// new
const nextDay = enableOutsideDays && currentDay.clone().add(count, 'day').local();
```

What do you think about that @majapw?

---

More about [`moment#local`](http://momentjs.com/docs/#/manipulating/local/).

majapw:
Does that actually still solve the DST issue @wmartins? I kind of feel like the setting the time to the middle of the day is still a bit safer and a bit simpler.

wmartins:
I would need to test a little bit more to confirm more cases, but if you
think that it is more simple to set the hours, I think it is not a problem.

Just investigated the issue a little bit and brought this alternative
solution. But I think your solution is fine!

Em qua, 12 de out de 2016 12:28, Maja Wichrowska <notifications@github.com>
escreveu:

majapw:
@ljharb I added some tests that I confirmed used to fail (with the utcOffset) and now pass. Can you PTAL and then we can get this bad boy merged in. :)

test/utils/getCalendarMonthWeeks_spec.js
```diff
@@ -27,6 +28,40 @@ describe('getCalendarMonthWeeks', () => {
...
+          expect(day.hours()).to.equal(12);
+        }
+      });
+    });
+  });
+
+  describe('Daylight Savings Time issues', () => {
+    it('last of February does not equal first of March', () => {
+      const february = getCalendarMonthWeeks(today.clone().month(1));
+      const lastWeekOfFebruary = february[february.length - 1].filter(day => !!day);
```
ljharb: fwiw `.filter(Boolean)` works here too :-)

test/utils/getCalendarMonthWeeks_spec.js
```diff
@@ -27,6 +28,40 @@ describe('getCalendarMonthWeeks', () => {
...
+      const lastOfFebruary = lastWeekOfFebruary[lastWeekOfFebruary.length - 1];
+
+      const march = getCalendarMonthWeeks(today.clone().month(2));
+      const firstOfMarch = march[0].filter(day => !!day)[0];
```
ljharb: This should probably use `.find(Boolean)` instead

majapw:
Awesome! I fixed the tests that were breaking as a result of changing the date's time. :) So we should be set now. Will merge in after tests pass.

PR #116, opened 2016-10-12
refactor: remove array-includes dependency

mrtnbroder:
I saw no need for the array-includes dependency if you can fallback to a simple `indexOf`

Notice: The builds fails currently because of something else

### DIFF
--- a/package.json
+++ b/package.json
@@ -94,7 +94,6 @@
     "webpack": "^1.13.2"
   },
   "dependencies": {
-    "array-includes": "^3.0.2",
     "classnames": "^2.2.5",
     "react-moment-proptypes": "^1.2.0",
     "react-portal": "^2.2.1"
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -3,7 +3,6 @@ import ReactDOM from 'react-dom';
 import moment from 'moment';
 import cx from 'classnames';
 import Portal from 'react-portal';
-import includes from 'array-includes';
 
 import isTouchDevice from '../utils/isTouchDevice';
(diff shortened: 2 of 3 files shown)


ljharb:
No thanks - dependencies are a good thing, ~ is far too clever to ever be a good idea to use in code, and includes handles NaN and -0 while indexOf does not.

mrtnbroder:
Do you mind explaining why dependencies are a good thing, @ljharb? Cause imo they are not, since you're adding 1. weight to your app 2. depend on third party stuff that you can't directly control

And I don't really get your argument for the NaN and -0 since we're just checking for a simple string within an array here? I get your point about the `~` being too magical here though.

ljharb:
Sure - one idea is that devs shouldn't have to think about quirks. This is a date library, and NaNs are a thing - even if in this case, we're just comparing to a string literal (where I agree, of course, indexOf works fine). Having this dependency in the package and having an example of its use encourages proper use of it in the future.

Second, if you're using this package (the official polyfill for `Array#includes`) in your top-level app (as I hope you are) then it'd be deduped anyways, so the net weight added to this package would be zero. Regardless, https://wzrd.in/standalone/array-includes@latest (minus the browserify boilerplate that you'd need in your app anyways), it's a super tiny package that will gzip down to virtually nothing. [Bloat](https://twitter.com/trek/status/227082154109186049) just [isn't a huge concern](https://twitter.com/paul_irish/status/227203681462927361) in a practical sense.

Third, I happen to maintain that module, so I do directly control it - but either way, it's important to trust the ecosystem. If a dependency changes in a way that violates semver, we'll just stop using it - and our tests cover the things that use it, so we'll find out immediately - but most dependencies follow semver strictly.

PR #118, opened 2016-10-12
Attaches DayPicker caret to inputs instead of the picker component

majapw:
So I've been thinking about this problem specifically for https://github.com/airbnb/react-dates/pull/83, and after @travisbloom introduced `anchorDirection` in (https://github.com/airbnb/react-dates/pull/72, it became pretty clear that we were doing a lot of grossness to make carets consistently align. Combine this with the fact that we have to override caret styles at Airbnb for medium breakpoints, and the point is that I was getting annoying.

While it makes some sense to attach the caret to the `DateRangePicker__picker`, I think it actually makes more sense for placement to attach them to the inputs. 

to: @airbnb/webinfra

### DIFF
--- a/css/DateInput.scss
+++ b/css/DateInput.scss
@@ -14,6 +14,28 @@
   vertical-align: middle;
 }
 
+.DateInput--focused::before,
+.DateInput--focused::after {
+  content: "";
+  display: inline-block;
+  position: absolute;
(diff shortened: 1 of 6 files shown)


wyattdanger:
LGTM

travisbloom:
so much cleaner, nice @majapw

PR #119, opened 2016-10-13
Allow DateRangePicker to have `minimumNights={0}`

### LINKED ISSUE #24: DateRangePicker can't select one day
In DateRangePicker is no ability to select one day?

LukeAskew:
Fixes #24

### DIFF
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -94,6 +94,7 @@ export default class DateRangePicker extends React.Component {
   }
 
   onDayClick(day, modifiers, e) {
+    const { minimumNights } = this.props;
     if (e) e.preventDefault();
     if (includes(modifiers, 'blocked')) return;
 
@@ -104,20 +105,22 @@ export default class DateRangePicker extends React.Component {
       this.props.onFocusChange(END_DATE);
 
       startDate = day;
+
       if (isInclusivelyAfterDay(day, endDate)) {
         endDate = null;
       }
     } else if (focusedInput === END_DATE) {
-      if (isInclusivelyBeforeDay(day, startDate)) {
+      const firstAllowedEndDate = startDate && startDate.clone().add(minimumNights, 'days');
+
+      if (!startDate) {
+        endDate = day;
+        this.props.onFocusChange(START_DATE);
+      } else if (isInclusivelyAfterDay(day, firstAllowedEndDate)) {
+        endDate = day;
+        this.props.onFocusChange(null);
+      } else {
         startDate = day;
         endDate = null;
-      } else {
-        endDate = day;
-        if (!startDate) {
-          this.props.onFocusChange(START_DATE);
-        } else {
-          this.props.onFocusChange(null);
-        }
       }
     }
 
@@ -268,9 +271,9 @@ export default class DateRangePicker extends React.Component {
   }
 
   isDayAfterHoveredStartDate(day) {
-    const { startDate, endDate } = this.props;
+    const { startDate, endDate, minimumNights } = this.props;
     const { hoverDate } = this.state;
-    return !!startDate && !endDate && isNextDay(hoverDate, day) &&
+    return !!startDate && !endDate && isNextDay(hoverDate, day) && minimumNights > 0 &&
       isSameDay(hoverDate, startDate);
   }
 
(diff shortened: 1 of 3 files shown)


LukeAskew:
Tests added 👍

src/components/DateRangePicker.jsx
```diff
@@ -108,7 +109,7 @@ export default class DateRangePicker extends React.Component {
         endDate = null;
       }
     } else if (focusedInput === END_DATE) {
-      if (isInclusivelyBeforeDay(day, startDate)) {
+      if (minimumNights > 0 && isInclusivelyBeforeDay(day, startDate)) {
```
majapw: Won't this allow you to select end dates before the start date, if minimumNights is set to zero?

ljharb: Please add a test that ensures that you can't :-)

LukeAskew: Yeah just verified. Let me apply a fix.

LukeAskew: Amended.

src/components/DateRangePicker.jsx
```diff
@@ -270,10 +276,10 @@ export default class DateRangePicker extends React.Component {
   }
 
   isDayAfterHoveredStartDate(day) {
-    const { startDate, endDate } = this.props;
+    const { startDate, endDate, minimumNights } = this.props;
     const { hoverDate } = this.state;
     return !!startDate && !endDate && isNextDay(hoverDate, day) &&
-      isSameDay(hoverDate, startDate);
+      isSameDay(hoverDate, startDate) && minimumNights > 0;
```
LukeAskew: This is problematic. Patching now.

majapw:
I wanna check out this code and play around with it a bit before merging it in. :)

src/components/DateRangePicker.jsx
```diff
@@ -110,7 +111,12 @@ export default class DateRangePicker extends React.Component {
     } else if (focusedInput === END_DATE) {
       if (isInclusivelyBeforeDay(day, startDate)) {
```
majapw: So this is not quite the logic we want in this if statement. I think the following is better:
```
const firstAllowedEndDate = startDate && startDate.clone().add(minimumNights, 'days');
if (!isInclusivelyAfterDay(day, firstAllowedEndDate)) {
  startDate = day;
  endDate = null;
} else {
  endDate = day;
  if (!startDate) {
    this.props.onFocusChange(START_DATE);
  } else {
    this.props.onFocusChange(null);
  }
}
```

src/components/DateRangePicker.jsx
flyingalex: Did this work now?

LukeAskew: Yes. The next day should only appear hovered is the minimum nights is at least 1.

src/components/DateRangePicker.jsx
```diff
@@ -104,20 +105,26 @@ export default class DateRangePicker extends React.Component {
...
-      } else {
+      const firstAllowedEndDate = startDate && startDate.clone().add(minimumNights, 'days');
+
+      if (!startDate) {
+        endDate = day;
+        this.props.onFocusChange(START_DATE);
+      } else if (isInclusivelyAfterDay(day, firstAllowedEndDate)) {
         endDate = day;
         if (!startDate) {
           this.props.onFocusChange(START_DATE);
```
majapw: This case will never happen because it'll be cause by line 115. We should delete this.

src/components/DateRangePicker.jsx
```diff
@@ -268,9 +275,9 @@ export default class DateRangePicker extends React.Component {
   }
 
   isDayAfterHoveredStartDate(day) {
-    const { startDate, endDate } = this.props;
+    const { startDate, endDate, minimumNights } = this.props;
     const { hoverDate } = this.state;
-    return !!startDate && !endDate && isNextDay(hoverDate, day) &&
+    return !!startDate && !endDate && (isNextDay(hoverDate, day) && minimumNights > 0) &&
```
majapw: I don't think we need that extraneous parentheses around both of these

LukeAskew: You do because I am amending the `isNextDay` condition to be dependent on `minimumNights`. If it's not wrapped and `minimumNights={0}`, then the adjacent dates don't get "hovered".

LukeAskew:
Alrighty folks. I think all the feedback is addressed and the commits have been squashed. 😎  🎸

majapw:
Thanks @LukeAskew!

PR #121, opened 2016-10-13
Allows selecting the end date first for unattached vertically oriented datepickers

majapw:
This was behavior from back when we assume the vertical orientation would only ever be used on mobile devices and with a portal. This fixes the assumption.

to: @airbnb/webinfra

### DIFF
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -159,10 +159,10 @@ export default class DateRangePicker extends React.Component {
   }
 
   onEndDateFocus() {
-    const { startDate, onFocusChange, orientation, disabled } = this.props;
+    const { startDate, onFocusChange, withFullScreenPortal, disabled } = this.props;
 
-    if (!startDate && orientation === VERTICAL_ORIENTATION && !disabled) {
-      // Since the vertical datepicker is full screen, we never want to focus the end date first
+    if (!startDate && withFullScreenPortal && !disabled) {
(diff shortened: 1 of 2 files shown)


src/components/DateRangePicker.jsx
```diff
@@ -159,10 +159,10 @@ export default class DateRangePicker extends React.Component {
   }
 
   onEndDateFocus() {
-    const { startDate, onFocusChange, orientation, disabled } = this.props;
+    const { startDate, onFocusChange, withFullScreenPortal, disabled } = this.props;
 
-    if (!startDate && orientation === VERTICAL_ORIENTATION && !disabled) {
-      // Since the vertical datepicker is full screen, we never want to focus the end date first
+    if (!startDate && withFullScreenPortal && !disabled) {
```
ljharb: do we want to keep this as semver-minor by leaving the vertical orientation check in place?

majapw: well but the old behavior was like, broken. So I think this needs to stay.

test/components/DateRangePicker_spec.jsx
```diff
@@ -894,12 +894,12 @@ describe('DateRangePicker', () => {
       });
     });
 
-    describe('props.orientation = VERTICAL_ORIENTATION', () => {
+    describe('props.withFullScreenPortal is truthy', () => {
```
majapw: This is the test @ljharb

PR #137, opened 2016-10-20
WIP: Implement with styles

majapw:
This is a huge change! Will probably require a major release.

All the conversions are not done yet.

Also we're going to need to create some sort of generator for a css stylesheet as well as an interface for `react-with-styles` that applies the classes... so as to support the regular css stylesheets for those who want them.

idk! This is what a 10 hour flight has netted me thus far basically (I had to stop at some point because my eyes felt like they were bleeding X_x).


package.json
```diff
@@ -94,11 +94,14 @@
     "webpack": "^1.13.2"
   },
   "dependencies": {
+    "aphrodite": "^0.5.0",
```
ljharb: Doesn't this mean that everyone has to install aphrodite even when not using it?

lencioni: Yes, you need to do this in a way that doesn't couple this project to a particular interface. @adamrneary is doing something similar with Rheostat right now, so it might be a good idea to sync up about how to best accomplish this. Relevant PR: https://github.com/airbnb/rheostat/pull/38

PR #142, opened 2016-10-22
Added support for `required` property

### LINKED ISSUE #136: Support `required` attribute on inputs
Seems the MVP to have at least the ability to set `required` to the `input` elements.

FezVrasta:
Fixes #136.


FezVrasta:
I think it's the exact opposite, I use it exactly because it's a standard across the whole browser, so the people don't have to learn a new pattern on each website.

PR #145, opened 2016-10-24
[WIP] Extract DayPicker modifiers as DayPickerInstrumented

moonboots:
Hey @majapw, I'm working on extracting the `DayPicker` modifier logic into a separate component so it can be used independent of a triggering `DateRangePicker`. This logic might even make sense baked completely in `DayPicker`.

### DIFF
--- a/src/components/DayPickerInstrumented.jsx
+++ b/src/components/DayPickerInstrumented.jsx
@@ -4,7 +4,6 @@ import moment from 'moment';
 import cx from 'classnames';
 import Portal from 'react-portal';
 import includes from 'array-includes';
-import TetherComponent from 'react-tether';
 
 import isTouchDevice from '../utils/isTouchDevice';
 import toMomentObject from '../utils/toMomentObject';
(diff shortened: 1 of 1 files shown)


majapw:
@moonboots gonna take a look at this today! sorry for the delaaaaays

majapw:
@moonboots we can close this PR now, right?

src/components/DayPickerInstrumented.jsx
```diff
@@ -0,0 +1,365 @@
...
+
+  onDayMouseLeave() {
+    if (this.isTouchDevice) return;
+
+    this.setState({
+      hoverDate: null,
+    });
+  }
+
+  onEndDateChange(endDateString) {
```
majapw: I think that `onEndDateChange` and `onStartDateChange` belong in the `DateRangePicker` still. As do the `onFocusChange` handlers. Basically, anything that deals with the inputs should be in `DateRangePicker`, anything that deals with the modifiers/`DayPicker` only should be here.

PR #146, opened 2016-10-25
:house_with_garden: Build every time package is installed

FenrisWulf:


ljharb:
Just like #147, this would be a very bad idea. Building needs to happen prepublish; consumers should never need to do it.

PR #147, opened 2016-10-25
Build on install

timglaser:


package.json
```diff
@@ -4,6 +4,7 @@
   "description": "A responsive and accessible date range picker component built with React",
   "main": "index.js",
   "scripts": {
+    "postinstall": "npm run build",
```
ljharb: most definitely we wouldn't want to build on install.

package.json
```diff
@@ -32,7 +33,7 @@
     "version:major": "npm --no-git-tag-version version major",
     "preversion": "npm run test && npm run check-changelog && npm run check-only-changelog-changed",
     "postversion": "git commit package.json CHANGELOG.md -m \"Version $npm_package_version\" && npm run tag && git push && git push --tags && npm publish --registry=https://registry.npmjs.org/",
-    "prepublish": "in-publish && safe-publish-latest && npm run build || not-in-publish",
+    "prepublish": "npm run build",
```
ljharb: the `in-publish` and `safe-publish-latest` checks are required to gate publishing - we don't actually want to run the build upon installing.

PR #154, opened 2016-10-29
Makes Nav Arrows Configurable

### LINKED ISSUE #112: Make navigation arrows configurable
It'd be nice if we could pass in an element in place of the current svg arrows.

SecurityInsanity:
fixes #112 

This PR adds in the ability to specify `navPrev`, and `navNext` to DayPicker incase they want to override the default SVGs that come bundled with react-dates (e.g. `ChevronUp`, `ChevronDown`, `LeftArrow`, and `RightArrow`).

If `navPrev`, or `navNext` aren't specified then DayPicker will fall back to the current behavior of:

``` js
isVertical ? <ChevronUp /> : <LeftArrow />;
```

for the previous button, and:

``` js
isVertical ? <ChevronDown /> : <RightArrow />;
```

for the next button.

I also tried to make this work as much with existing code as possible. As such even with custom icons the class names: `DayPicker__nav--prev`, and `DayPicker__nav--next` are still rendered even with custom icons. If you really want to target something inside of your custom icon, you can always tag class names/ids onto the props you pass in. E.g.:

```
<DayPicker
  navPrev={
    <span className={'MySuperAwesomeIcon'}>
       'Prev'
    </span>
  }
/>
```


src/components/DayPicker.jsx
```diff
@@ -288,22 +290,36 @@ export default class DayPicker extends React.Component {
...
+      navPrev,
+      navNext,
+    } = this.props;
+
     const isVertical = this.isVertical();
 
+    let navPrevIcon = navPrev;
+    let navNextIcon = navNext;
+    if (!navPrevIcon) {
+      navPrevIcon = isVertical ? <ChevronUp /> : <LeftArrow />;
```
ljharb: it seems like we'd want callers to be able to perform the same ternary, so they don't have to know about the "isVertical" implementation.

What would you think about making `navPrev` and `navNext` be functions, that get called with `isVertical`, so that `navPrev` would default to `vertical => (vertical ? <ChevronUp /> : <LeftArrow />)`?

SecurityInsanity: I thought about that initially, but my main argument in my head against it was. "isVertical is a prop being passed in like navPrev/navNext". E.g. they already know which is being rendered because the caller themselves is passing in the property.

That being said if there's a situation I'm not thinking of where they are unsure of it being vertical/not vertical, or we just determine it's "better" :tm: I'd be happy to make that change.

ljharb: You make a fair counterpoint. I'm not 100% certain. Maybe @majapw can weigh in

src/components/DayPicker.jsx
```diff
@@ -43,6 +43,8 @@ const propTypes = {
   onPrevMonthClick: PropTypes.func,
   onNextMonthClick: PropTypes.func,
   onOutsideClick: PropTypes.func,
+  navPrev: PropTypes.node,
+  navNext: PropTypes.node,
```
SecurityInsanity: Sure I can add `null` here.

src/components/DayPicker.jsx
SecurityInsanity: Glad to know my thinking wasn't crazy :smile:

SecurityInsanity:
I thought about that quite a bit, but the main thing I drew the line at is where do we draw the 'supported' line.

I'd be happy to test this on resolutions, and find some nice defaults. However, is there some precedent at AirBnB, e.g. what does AirBnB normally test on as the lowest resolution/on the highest resolution? Do we want to support text (like I have currently in the storybook), or do we only want to support icons?

Basically I'm just looking for some basic rules that I can then build on to say "this is supported", and "this isn't" so that way we don't just open up a catch-all "anything you stick in this field is supported".

ljharb:
@SecurityInsanity anything that propTypes allow to be put into the prop is supported, whether we officially declare it or not - if we want to limit what we have to support, a custom propType validator (along with documentation, of course) is the best way to do that.

SecurityInsanity:
I agreed with you @majapw about allowing more "custom" buttons if you will, and I've gone ahead and split the icon into two classes as you requested.

I actually split up in the exact way you suggested because I thought it fit well. As such by not using default styling all your getting is your button position in the correct place. However, the default styles will go ahead  with adding the coloring/border/etc.

Lastly I took a look at splitting the Navigation into it's own component, and wasn't really happy with any of the solutions. Of course if we still decide this is wanted I can make the change. Some of the things I found where:
- Making the navigation it's own component means it has to take in the props `navPrev`, or `navNext`. This means the navigation component has to do one of the following:
  - Make the icons required to pass in (not a huge fan).
  - Get passed in a prop that determines if we should render the default "vertical"/"horizontal" icon.
  - Gets passed in orientation, and has it's own implementation of `isVertical`.

All in all the navigation is a fairly small function. So I don't like any of the proposed solutions in my head.

src/components/DayPicker.jsx
```diff
@@ -288,22 +294,44 @@ export default class DayPicker extends React.Component {
...
+    let navPrevIcon = navPrev;
+    let navNextIcon = navNext;
+    if (!navPrevIcon) {
+      navPrevIcon = isVertical ? <ChevronUp /> : <LeftArrow />;
+    }
+    if (!navNextIcon) {
+      navNextIcon = isVertical ? <ChevronDown /> : <RightArrow />;
+    }
+
+    const prevClassNames = cx('DayPicker__nav--prev', {
```
majapw: I think having them separate is fine

src/components/DayPicker.jsx
```diff
@@ -55,6 +58,9 @@ const defaultProps = {
   orientation: HORIZONTAL_ORIENTATION,
   withPortal: false,
   hidden: false,
+  navPrev: null,
+  navNext: null,
+  useNavDefaultStyles: true,
```
majapw: Should the `DateRangePicker` or `SingleDatePicker` ever have the ability to change this or should this remain kind of like a hidden prop?

majapw:
I still think we will eventually pull out the navigation (it makes a lot of sense once we look at implementing `react-with-styles`) but I do not think that is necessary to do in this PR. 

Um, this is looking good! My one open Q is about whether or not the `DateRangePicker` or `SingleDatePicker` should control the prop `useDefaultNavStyles`. I feel like this is not something we should have as a prop on the DRP, but rather should be set to false immediately when the user passes in a custom nav element. What do you think?

src/components/DayPicker.jsx
SecurityInsanity: Hmm. I totally forgot about those to while implementing this to be honest. I think both of them should have the ability to change this prop.

SecurityInsanity:
Ah I didn't realize you were looking at implementing `react-with-styles`. I think that might be a good solution to go down the road. I'll take a look at splitting it up, but if I don't find a solution I'm happy with I'll just leave it if it's not necessary.

I like the idea of having this prop be set to false immediately after a prop custom prop has been passed in. If that's the case then it probably shouldn't be a prop, and rather a local variable to the function `renderNavigation`. So I'll go ahead, and do that. (I'll also clean up the storybook implementation, and leave a note in the docs).

majapw:
SGTM, @SecurityInsanity! Let me know when you make that change so I can do a final pass and we'll get this bad boy merged in. :)

SecurityInsanity:
@majapw I've gone ahead and split the `DayPicker`'s navigation into it's own component `DayPickerNavigation` (I decided afterall passing in isVertical isn't as ugly as I imagined).

I've also implemented the "default" classes being applied only if a custom prop hasn't been passed in. (Moving it to the local function).

Last but not least the documentation has been updated to include the details about `navPrev`, and `navNext`.

P.S. I hear your "literally dying" so get well soon (I've found some Jasmine tea generally helps!)

css/DayPickerNavigation.scss
```diff
@@ -0,0 +1,96 @@
...
+    font-size: 2.5em;
+    padding: 5px;
+    display: inline-block;
+    position: relative;
+    height: 100%;
+    width: 50%;
+  }
+
+  .DayPicker__nav--next {
+    border-left: 0;
```
majapw: default probably?

css/DayPickerNavigation.scss
```diff
@@ -0,0 +1,96 @@
...
+    height: 52px;
+    width: 100%;
+    z-index: 2;
+  }
+
+  .DayPicker__nav--prev,
+  .DayPicker__nav--next {
+    text-align: center;
+    font-size: 2.5em;
+    padding: 5px;
```
majapw: make padding and maybe font/text-align be default as well?

src/components/DayPicker.jsx
```diff
@@ -6,11 +6,7 @@ import cx from 'classnames';
 
 import OutsideClickHandler from './OutsideClickHandler';
 import CalendarMonthGrid from './CalendarMonthGrid';
-
-import LeftArrow from '../svg/arrow-left.svg';
-import RightArrow from '../svg/arrow-right.svg';
-import ChevronUp from '../svg/chevron-up.svg';
-import ChevronDown from '../svg/chevron-down.svg';
+import DayPickerNavigation from './DayPickerNavigation';
```
majapw: Awesome.

src/components/DayPicker.jsx
```diff
@@ -288,24 +288,19 @@ export default class DayPicker extends React.Component {
...
-          onClick={this.handleNextMonthClick}
-        >
-          {isVertical ? <ChevronDown /> : <RightArrow />}
-        </span>
-      </div>
+      <DayPickerNavigation
+        navPrev={navPrev}
+        navNext={navNext}
+        isVertical={this.isVertical()}
+        handlePrevMonthClick={this.handlePrevMonthClick}
```
majapw: I've been in the slow business of renaming `handle...` event methods to `on...`. Let's do that here and call the props `onPrevMonthClick` and `onNextMonthClick`

src/components/DayPickerNavigation.jsx
```diff
@@ -0,0 +1,75 @@
...
+    navPrev,
+    navNext,
+    isVertical,
+    handlePrevMonthClick,
+    handleNextMonthClick,
+  } = props;
+
+  let navPrevIcon = navPrev;
+  let navNextIcon = navNext;
+  let isDefaualtNavPrev = false;
```
majapw: typo! Should be `isDefaultNavPrev`.

src/components/DayPickerNavigation.jsx
```diff
@@ -0,0 +1,75 @@
...
+  if (!navPrevIcon) {
+    isDefaualtNavPrev = true;
+    navPrevIcon = isVertical ? <ChevronUp /> : <LeftArrow />;
+  }
+  if (!navNextIcon) {
+    isDefaultNavNext = true;
+    navNextIcon = isVertical ? <ChevronDown /> : <RightArrow />;
+  }
+
+  const prevClassNames = cx('DayPicker__nav--prev', {
```
majapw: Since we've got the nice new structure here, let's rename the classes to match. I also think the prev/nav make more sense as elements rather than modifiers so maybe we'd have:
```
DayPickerNavigation
DayPickerNavigation--horizontal
DayPickerNavigation--vertical
-- DayPickerNavigation__prev
-- DayPickerNavigation__prev--default
-- DayPickerNavigation__next
-- DayPickerNavigation__next--default
```

css/DayPickerNavigation.scss
```diff
@@ -0,0 +1,96 @@
...
+  .DayPicker__nav-prev-default {
+    svg {
+      height: 19px;
+      width: 19px;
+      fill: $react-dates-color-gray-light;
+    }
+  }
+}
+
+.DayPicker--vertical {
```
majapw: We don't want the `DayPicker` classes bleeding into this file so let's add a `DayPickerNavigation--vertical` and `DayPickerNavigation--horizontal` class since we've got access to the prop anyways.

README.md
```diff
@@ -186,11 +194,12 @@ The `monthFormat` prop abides by [moment's date formatting rules](http://momentj
   monthFormat: PropTypes.string,
 ```
 
-The `phrases` prop is an object that contains all the English language phrases currently part of the class. As of now, we only have two such phrases and neither are visible but they are used for screen-reader navigation of the datepicker.
+The `phrases` prop is an object that contains all the English language phrases currently part of the class. As of now, we only have three such phrases and none are visible but they are used for screen-reader navigation of the datepicker.
```
SecurityInsanity: No problem! Realized I hadn't added it in my previous PR.

css/DayPickerNavigation.scss
```diff
@@ -0,0 +1,96 @@
...
+
+.DayPicker--horizontal {
+  .DayPicker__nav {
+    position: relative;
+  }
+
+  .DayPicker__nav--prev,
+  .DayPicker__nav--next {
+    border-radius: 3px;
+    padding: 6px 9px;
```
SecurityInsanity: I think the `border-radius` should be moved. Not sure about the padding though as I think the position might look weird without this. I'll look into it though.

css/DayPickerNavigation.scss
SecurityInsanity: :+1:

css/DayPickerNavigation.scss
SecurityInsanity: Yep those totally belong there.

css/DayPickerNavigation.scss
SecurityInsanity: I'm not sure how this affect positioning, but I'll def see how it looks without.

src/components/DayPicker.jsx
SecurityInsanity: Sure thing!

src/components/DayPickerNavigation.jsx
SecurityInsanity: :astonished: I just mash buttons on a keyboard. It's a miracle this is my first typo.

src/components/DayPickerNavigation.jsx
SecurityInsanity: I'll go ahead, and rename those.

SecurityInsanity:
@majapw I've gone ahead, and updated the things as you requested. Hopefully I didn't typo something else this time :smile:

css/DayPickerNavigation.scss
```diff
@@ -0,0 +1,90 @@
...
+  .DayPickerNavigation__next {
+    display: inline-block;
+    position: relative;
+    height: 100%;
+    width: 50%;
+  }
+
+  .DayPickerNavigation__prev--default,
+  .DayPickerNavigation__next--default {
+    border-left: 0;
```
majapw: I think the `border-left` needs to be only on `.DayPickerNavigation__next--default`

(since the navigation buttons here are side by side, it basically allows for `| prev | next |` instead of `| prev || next |`)

src/components/DayPickerNavigation.jsx
```diff
@@ -0,0 +1,79 @@
...
+    navPrev,
+    navNext,
+    isVertical,
+    onPrevMonthClick,
+    onNextMonthClick,
+  } = props;
+
+  let navPrevIcon = navPrev;
+  let navNextIcon = navNext;
+  let isDefualtNavPrev = false;
```
majapw: still here. 😭

src/components/DayPickerNavigation.jsx
```diff
@@ -0,0 +1,79 @@
...
+
+  onPrevMonthClick: PropTypes.func,
+  onNextMonthClick: PropTypes.func,
+};
+const defaultProps = {
+  navPrev: null,
+  navNext: null,
+  isVertical: false,
+
+  handleNextMonthClick() {},
```
majapw: change these to be onNextMonthClick and onPrevMonthClick as well

css/DayPickerNavigation.scss
SecurityInsanity: I'll take a look into adding it there.

src/components/DayPickerNavigation.jsx
SecurityInsanity: hur durr

src/components/DayPickerNavigation.jsx
SecurityInsanity: :Seppuku:

SecurityInsanity:
Fixed those last few things, and double checked my spelling. Let's hope that doesn't backfire.

src/components/SingleDatePicker.jsx
```diff
@@ -198,6 +198,8 @@ export default class SingleDatePicker extends React.Component {
       numberOfMonths,
       orientation,
       monthFormat,
+      navPrev,
+      navNext,
```
majapw: Probably we should do the same thing with default props on the SDP and DRP

src/components/DateRangePicker.jsx
```diff
@@ -335,6 +335,8 @@ export default class DateRangePicker extends React.Component {
       numberOfMonths,
       orientation,
       monthFormat,
+      navPrev,
+      navNext,
```
majapw: probably let's add default props here as well

majapw:
If you can rebase and squash down the commits into one that would be great!

SecurityInsanity:
I made a huge mess squashing these locally :sob: I'm going to close this, and open a new PR referencing this one.

majapw:
Lol okay! whoops. D:

(I was looking for this PR and got distressed when it disappeared, lol)

Issue #62, opened 2016-09-04
Auto Hide when selecting end Date is not good at all.


nsisodiya:
When We select end date, Calendar autohide. this is not a Good user experience. 
when User click, on start and end Date, he expect too duration basically visually. Seeing shaded portion after clicking on end date is a mental `reconfirmation` by user. When He click first date, popup stay as it is. So he expect same behaviour when he click end date too. Once he mentally confirm that he selected correct duration, he will click outside and then you can hide the popup.

majapw:
This is something we are working on with our designers. We may be moving towards a model where users have to close the picker and focus on the next element themselves. However, this may only be targeted at keyboard users. I will update as this progress continues.

majapw:
We've actually added a prop to prevent this behavior here - https://github.com/airbnb/react-dates/pull/157 so this should soon be possible to avoid if necessary!

majapw:
This is fixed now. You can use the `keepOpenOnDateSelect` prop to prevent this behavior.

PR #157, opened 2016-11-02
Adds a prop that keeps the DayPicker open when you select a date

majapw:
to: @moonboots @airbnb/webinfra 

The prop needs a better name. Also this needs tests. But I'm literally dying.

However, this behavior was pretty easy to implement so that's cool.

### DIFF
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -47,6 +47,7 @@ const defaultProps = {
   disabled: false,
   required: false,
   reopenPickerOnClearDates: false,
+  keepOpenOnDateSelect: false,
   initialVisibleMonth: () => moment(),
   navPrev: null,
   navNext: null,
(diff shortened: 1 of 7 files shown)


moonboots:
lgtm thanks!

src/components/DateRangePicker.jsx
```diff
@@ -47,6 +47,7 @@ const defaultProps = {
   disabled: false,
   required: false,
   reopenPickerOnClearDates: false,
+  keepOpen: false,
```
ljharb: maybe `keepOpenOnSelect`?

stories/SingleDatePicker.js
```diff
@@ -70,12 +70,17 @@ storiesOf('SingleDatePicker', module)
       showClearDate
     />
   ))
-  .add('with clear dates button (Picker Displayed)', () => (
+  .add('eopens DayPicker on clear dates', () => (
```
ljharb: `s/eopens/reopens`

majapw:
Addressed comments! PTAL @ljharb

majapw:
This also fixes https://github.com/airbnb/react-dates/issues/62

PR #158, opened 2016-11-02
force correct text-align on calendar

calcsam:
If DatePicker is placed in div with a different text-align, for example text-align:center, css class inheritance from parent will break the calendar month picker & make dates invisible. This fixes the problem.

### DIFF
--- a/css/CalendarMonthGrid.scss
+++ b/css/CalendarMonthGrid.scss
@@ -3,6 +3,7 @@ $react-dates-width-day-picker: 300px;
 .CalendarMonthGrid {
   background: #fff;
   z-index: 0;
+  text-align: initial;
 }
 
 .CalendarMonthGrid--animating {


css/CalendarMonthGrid.scss
```diff
@@ -3,6 +3,7 @@ $react-dates-width-day-picker: 300px;
 .CalendarMonthGrid {
   background: #fff;
   z-index: 0;
+  text-align: initial;
```
ljharb: Since initial doesn't work in all browsers we support ( http://caniuse.com/#search=initial ) could we also add, above this, `text-align: left;`, so that this bug can be fixed in those browsers too?

hassankhan:
Can this be merged soon, please?

majapw:
I believe this was fixed by https://github.com/airbnb/react-dates/pull/258

PR #159, opened 2016-11-02
Fixes DayPicker height issue on open

majapw:
When the initialVisibleMonth has a different height than the current month, the DayPicker does not adjust immediately. I noticed this in prod on airbnb.com. Whoops!

to: @airbnb/webinfra


jpollard-cs:
thanks for fixing this - thought one of my external styles was causing this to happen!

PR #162, opened 2016-11-03
Add bottomPanel and portalMarginTop props

moonboots:
to: @majapw 

Adds `bottomPanel` property for adding elements below `DayPicker` and `portalMarginTop` for tweaking space between input and portal/daypicker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,16 @@ The `anchorDirection` prop indicates whether the calendar is anchored to the rig
   withFullScreenPortal: PropTypes.bool
 ```
 
+The `bottomPanel` prop accepts a React element that will be displayed below `DayPicker`.
+```
+  bottomPanel: PropTypes.element
+```
+
+The `portalMarginTop` prop sets the space between `DateRangePickerInput` and the `DayPicker` popout.
+```
+  portalMarginTop: PropTypes.number
+```
+
 **Input presentation:**
 
 The `startDateId` and `endDateId` props are assigned to the actual `<input>` DOM elements for accessibility reasons. They default to the values of the `START_DATE` and `END_DATE` constants.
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -135,6 +135,16 @@ The `anchorDirection` prop indicates whether the calendar is anchored to the rig
   withFullScreenPortal: PropTypes.bool
 ```
 
+The `bottomPanel` prop accepts a React element that will be displayed below `DayPicker`.
+```
+  bottomPanel: PropTypes.element
+```
+
+The `portalMarginTop` prop sets the space between `DateRangePickerInput` and the `DayPicker` popout.
```
majapw: this should not be named `portalMarginTop` because the `DayPicker` is not necessarily in a portal (and in fact this margin does not affect the case where it is in a portal)

src/components/DateRangePicker.jsx
```diff
@@ -441,7 +447,7 @@ export default class DateRangePicker extends React.Component {
           })}
           attachment={`top ${anchorDirection}`}
           targetAttachment={`bottom ${anchorDirection}`}
-          offset="-23px 0"
+          offset={`-${portalMarginTop}px 0`}
```
majapw: We're going to want some logic somewhere to hide the caret when the container gets too close to the input I think.

src/components/DateRangePicker.jsx
```diff
@@ -405,6 +408,8 @@ export default class DateRangePicker extends React.Component {
             <CloseButton />
           </button>
         }
+
+        {bottomPanel}
```
majapw: Maybe put this inside a namespaced `<div className="DateRangePicker__bottom-panel">...</div>` or something?

moonboots:
Closing with a new PR that just adds bottom panel

Issue #139, opened 2016-10-20
Support a way to use your components without "Tether"


yaycmyk:
Sometimes you just want to render a date picker in-place...

majapw:
Can you describe what problems the tethering is causing for you exactly? The idea was to replicate the previous behavior but with the enhanced responsiveness that comes with tethering. If it broke behavior for you, it'd be good to know how.

yaycmyk:
If you are rendering a picker component inside another that is portalled in a way similar to the Yether component, it's difficult to do event detection via bubbling, etc.

ljharb:
React components are supposed to be self-contained - I wouldn't think you'd want to rely on events bubbling out of a React component anyways.

Your outside click detection would be most robust if it used the [`contains`](https://developer.mozilla.org/en-US/docs/Web/API/Node/contains) API, on the ref for the datepicker, to see if it's an inside click - that way it wouldn't matter how the HTML was structured.

yaycmyk:
@ljharb I do in fact use that method, however if one of the inner children does not actually render inside the tree like `Tether` does, then `contains` is useless.

yaycmyk:
Trust me, I've done the research on this and developed my own way of dealing with it for the time being. However, it would be better to simply offer a method of using the component without the tether. For example, sometimes form designs do not need the input field to actually be visible and the date picker lives by itself, adjusting a hidden input.

ljharb:
I'm in complete agreement that there should be a way to use it without tethering :-) i just don't think the use case should be "event bubbling".

yaycmyk:
> For example, sometimes form designs do not need the input field to actually be visible and the date picker lives by itself, adjusting a hidden input.

There you go. In that situation, you'd want to render in-place since it technically doesn't have anything specific to "tether" to.

thomasthesecond:
Another case that I can describe is one that I'm currently experiencing, and that is using the DateRangePicker component inside of another Tether component. In the mobile view of our app, our filter controls (which includes the date picker) render in a pop-over/modal style panel, which is tethered. In this case, I think it would be nice to have the option.

majapw:
Oh yeah, I am in fact running into this now when rendering the datepicker in a modal and it is gross. I was wondering if maybe being able to indicate the DOM element the tether attaches itself to might be sufficient? I really don't want to lose the responsiveness associated with the tether, but I *think* that maybe allowing the tether parent to be something other than the body might help (also with z-index issues).

[...]

PR #163, opened 2016-11-04
Removes the short-lived tether dependency in react-dates

majapw:
Instead, we're just going to do the thing internally.

Fix for https://github.com/airbnb/react-dates/issues/139 as well.

to: @moonboots @airbnb/webinfra 

Ugh

### DIFF
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -4,9 +4,9 @@ import moment from 'moment';
 import cx from 'classnames';
 import Portal from 'react-portal';
 import includes from 'array-includes';
-import TetherComponent from 'react-tether';
 
 import isTouchDevice from '../utils/isTouchDevice';
+import getResponsiveContainerStyles from '../utils/getResponsiveContainerStyles';
 import toMomentObject from '../utils/toMomentObject';
 import toLocalizedDateString from '../utils/toLocalizedDateString';
 
@@ -54,6 +54,7 @@ const defaultProps = {
 
   orientation: HORIZONTAL_ORIENTATION,
   anchorDirection: ANCHOR_LEFT,
+  horizontalMargin: 0,
   withPortal: false,
   withFullScreenPortal: false,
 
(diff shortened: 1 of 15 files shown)


src/components/DateRangePicker.jsx
```diff
@@ -378,7 +424,11 @@ export default class DateRangePicker extends React.Component {
     const onOutsideClick = !withFullScreenPortal ? this.onOutsideClick : undefined;
 
     return (
-      <div className={this.getDayPickerContainerClasses()}>
+      <div
+        ref={ref => { this.dayPickerContainer = ref; }}
```
ljharb: the linter should be requiring arg parens here

src/components/SingleDatePicker.jsx
```diff
@@ -224,7 +268,11 @@ export default class SingleDatePicker extends React.Component {
     const onOutsideClick = !withFullScreenPortal ? this.onClearFocus : undefined;
 
     return (
-      <div className={this.getDayPickerContainerClasses()}>
+      <div
+        ref={ref => { this.dayPickerContainer = ref; }}
```

majapw:
I think this is actually going to be a major bump because of the changes in DOM structure and classnames.

src/components/DateRangePicker.jsx
```diff
@@ -265,6 +282,34 @@ export default class DateRangePicker extends React.Component {
     }
   }
 
+  /* istanbul ignore next */
+  responsivizePickerPosition() {
+    const { anchorDirection } = this.props;
+    const { dayPickerContainerStyles } = this.state;
+
+    const containerClientRect = this.dayPickerContainer.getBoundingClientRect();
+    const documentBodyStyles = window.getComputedStyle(document.body);
```
mereskin: This call is fairly expensive because it traverses the entire cascade and builds CSSDeclaration. Maybe we can call `responsivizePickerPosition` inside `requestAnimationFrame`, what do you think?

majapw:
Most of these additions cannot really be well tested without DOM manipulation so I'm going to merge this in as is. :/

Issue #126, opened 2016-10-16
Doesn't open on click


FezVrasta:
So... I'm using it in this way:

```
import '!style!css!sass!react-dates/css/variables.scss';
import '!style!css!sass!react-dates/css/styles.scss';

function Test() {
  return (
   <DateRangePicker />
  );
}
```

It get properly rendered but when I click on the fields, the calendars are not shown...why?

majapw:
Hi @FezVrasta! The `DateRangePicker` is a fully controlled component which means that you need to create a wrapper component that maintains the focused element and selected date in its state and updates those values using the provided `onFocusChange` and `onDatesChange` callbacks. You can see an example of the code you'd need for this kind of wrapper component here: https://github.com/airbnb/react-dates/blob/master/examples/DateRangePickerWrapper.jsx

Basically, without this setup, no interactions will work.

FezVrasta:
Thanks!

mattgabor:
I'm running into the same issue, properly rendered but no response on click, but here's how my component setup looks:

`            <DateRangePicker
              startDate={null}
              endDate={null}
              focusedInput={focusedInput}
              onDatesChange={this.onDatesChange}
              onFocusChange={this.onFocusChange}
            />
`

The onDatesChange and onFocusChange functions are identical to the wrapper, they simply call setState, and focusedInput is set to null in the constructor

realisation:
I have a similar problem where I have provided the required props, but on click nothing renders. Do I still need a wrapper?

erin-doyle:
@realisation can yo provide some code to show?  Otherwise it's hard to say what the problem may be with your approach.

realisation:
I was simply using it without any wrapper component in my jsx. Inside an
enclosing parent div, I had two single date pickers with a span between
them. But then I made a wrapper component doing a singledatepicker alone,
by itself, in a parent div and it works just fine.

PR #164, opened 2016-11-04
Use stricter prop types to decrease confusion for new users

coreylight:
Related to https://github.com/airbnb/react-dates/issues/126

### DIFF
--- a/src/shapes/FocusedInputShape.js
+++ b/src/shapes/FocusedInputShape.js
@@ -5,4 +5,10 @@ import {
   END_DATE,
 } from '../../constants';
 
-export default PropTypes.oneOf([START_DATE, END_DATE]);
+export default function(props, propName, componentName){
+  const item = props[propName];
+  if (item === START_DATE || item === END_DATE || item === ''){
+    return null;
+  }
+  return new Error(`${propName} must be an empty string, "${START_DATE}", or "${END_DATE}"`);
+}


src/shapes/FocusedInputShape.js
```diff
@@ -5,4 +5,10 @@ import {
...
 } from '../../constants';
 
-export default PropTypes.oneOf([START_DATE, END_DATE]);
+export default function(props, propName, componentName){
+  const item = props[propName];
+  if (item === START_DATE || item === END_DATE || item === ''){
+    return null;
+  }
+  return new Error(`${propName} must be an empty string, "${START_DATE}", or "${END_DATE}"`);
+}
```
ljharb: this file needs a trailing newline

src/shapes/FocusedInputShape.js
```diff
@@ -5,4 +5,10 @@ import {
   END_DATE,
 } from '../../constants';
 
-export default PropTypes.oneOf([START_DATE, END_DATE]);
+export default function(props, propName, componentName){
+  const item = props[propName];
+  if (item === START_DATE || item === END_DATE || item === ''){
+    return null;
+  }
+  return new Error(`${propName} must be an empty string, "${START_DATE}", or "${END_DATE}"`);
```
ljharb: `PropTypes.oneOf` handles this already. if you want to allow an empty string, do `PropTypes.oneOf([START_DATE, END_DATE, ''])` - but I'm not sure why we'd want to allow an empty string.

src/shapes/FocusedInputShape.js
```diff
@@ -5,4 +5,10 @@ import {
   END_DATE,
 } from '../../constants';
 
-export default PropTypes.oneOf([START_DATE, END_DATE]);
+export default function(props, propName, componentName){
```
ljharb: also this now lacks an `.isRequired`, like most proptypes.

src/shapes/FocusedInputShape.js
coreylight: I couldn't get that particular piece of code to work:
`PropTypes.oneOf([START_DATE, END_DATE, ''])`
but if you can, then yes that seems better!

majapw: Wait, but currently we accept `START_DATE`, `END_DATE`, and `null`... do we need something else? Why do you want something else?

coreylight: `null` seems to be the case when a prop is either omitted or explicitly passed null. In my case, I omitted it and was not warned, so I had some confusion when first using this lib. This PR was my attempt to rectify that situation

ljharb: that's because it's not required - if you want it to be required, `PropTypes.oneOf([START_DATE, END_DATE]).isRequired` does that

majapw:
Hmm, maybe the problem you're trying to solve (that it's unclear that you need to set up the `onFocusChange`, `focusedInput`, `onDatesChange`, etc. props to get a functional datepicker) would be better addressed by an addition to the README?

PR #107, opened 2016-10-06
Added Getting Started instructions

arthurvi:
Instructions I was missing to get started using this date picker
See https://github.com/airbnb/react-dates/issues/45
closes #45

PR #168, opened 2016-11-08
Clarified "How to Install" in readme

savovs:
Hi guys, I'm super unfamiliar with shell scripts and it took me a while to decipher the readme. I thought it might be useful to clarify it for people who don't understand like me ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,13 +26,17 @@ To run that demo on your own computer:
 
 ## How to Use
 
-Ensure packages are installed with correct version numbers by running:
+Ensure packages are installed with correct version numbers by running the following shell script: 
   ```sh
   (
     export PKG=react-dates;
     npm info "$PKG" peerDependencies --json | command sed 's/[\{\},]//g ; s/: /@/g; s/ *//g' | xargs npm install --save "$PKG"
   )
   ```
+If you're new to shell scripts, create an install.sh file, copy the command above to it and run by typing this in your terminal: 
+```sh
+bash install.sh
+```
 
   Which produces and runs a command like:
 


README.md
```diff
@@ -26,13 +26,17 @@ To run that demo on your own computer:
...
 
-Ensure packages are installed with correct version numbers by running:
+Ensure packages are installed with correct version numbers by running the following shell script: 
   ```sh
   (
     export PKG=react-dates;
     npm info "$PKG" peerDependencies --json | command sed 's/[\{\},]//g ; s/: /@/g; s/ *//g' | xargs npm install --save "$PKG"
   )
   ```
+If you're new to shell scripts, create an install.sh file, copy the command above to it and run by typing this in your terminal:
```
ljharb: I feel like more useful info might be, "copy and paste this into your terminal"?

savovs: That didn't work for me for some reason (http://i.imgur.com/sxu7UKL.png)

ljharb: Are you on Windows, and is that git bash? This is a shell script, so the only thing on Windows it would work on is BashOnWindows.

jpollard-cs: @ljharb not sure if you saw it, but I put together a powershell script that should get the trick done - it's in the comments here: https://github.com/airbnb/react-dates/pull/107

jpollard-cs: @ljharb sorry just saw that you had replied on that PR a few days ago

jpollard-cs: @savovs it looks like you're using cmder - this won't work with cmder (I tried that too - it also won't work in Cygwin)
I would suggest adding a note that you should use Git Bash for this if you're on Windows 
I'm not sure how close that is to BashOnWindows, but I have only verified with Git Bash

majapw:
I'm going to be updating the docs completely at the end of next week so I'm going to close this in the meantime and try to incorporate these changes down the line since they appear to have been abandoned.

PR #170, opened 2016-11-08
Add bottom panel prop

moonboots:
to: @majapw 

Adds `bottomPanel` prop addressing [comments from previous PR](https://github.com/airbnb/react-dates/pull/162#discussion_r86472391). With the new tether changes, I no longer need the margin top property.

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,6 +135,11 @@ The `anchorDirection` prop indicates whether the calendar is anchored to the rig
   withFullScreenPortal: PropTypes.bool
 ```
 
+The `bottomPanel` prop accepts a React element that will be displayed below `DayPicker`.
+```
+  bottomPanel: PropTypes.element
+```
+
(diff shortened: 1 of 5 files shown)


youngzhaosignifyd:
Does the bottomPanel prop still exist? I can't find it in the [README](https://github.com/airbnb/react-dates)

majapw:
Hi @youngzhaosignifyd, you can add content to the bottomPanel using `renderCalendarInfo`

Basically, you want to pass in something like `renderCalendarInfo={() => <div>Hello World!</div>!. Check out https://github.com/airbnb/react-dates/blob/master/stories/DateRangePicker_calendar.js#L125-L132 for an example config.

youngzhaosignifyd:
@majapw Ah ok. Thanks! 

Should I create a new PR to add this info to the README?

Issue #156, opened 2016-11-01
Minimum and Max Dates on Single Date Picker


andrefox333:
Question,

Is there a way to pass a min/max date value to the single date picker? I'd like the user to only select dates that are obviously not in the past or way in the future...

Nice plugin by the way :)

Thanks!

Andre

ljharb:
@andrefox333 you can pass an `isOutsideRange` predicate function as a prop that can return `true` or `false` if the date is valid, which would allow you to restrict which dates are selectable.

andrefox333:
@ljharb, thanks for your reply :), I'm having a couple issues with the predicate function. If I have two input fields (startDate, endDate), I'd like my endDate input to not be able to select anything past the startDate. Very similar to how the date picker range component works. 

How does the isOutsideRange function know that the days from the startDate and before are not selectable?

I figured passing in the startDate into the function and do some checking with it would be enough but it's not working as intended. What should I be passing into the isOutsideRange function?

ljharb:
If you want a start and end date linked like that, why not use the DateRangePicker?

ljharb:
Either way, you can do it with your own wrapper component around two single date pickers, and that wrapper component would know onSelect/onChange what dates had been selected.

majapw:
@ljharb is right. The SingleDatePicker you're using for your end date needs to use your current start date in the logic for `isOutsideRange` somehow, which means you need a wrapper for both instances that would know about both the selected startDate and endDate

simonemonetti:
@andrefox333 just in case you still need it, I've submitted a PR that could solve your problem. Check it out -> #172

andrefox333:
@simonemonetti Thank you :)

PR #172, opened 2016-11-08
Add minimum date option to SingleDatePicker 

simonemonetti:
Hi guys,
I couldn't set a specific start date to `<SingleDatePicker />` and then I realized [I was not the only one](https://github.com/airbnb/react-dates/issues/156), so I basically added a `minDate` prop that allows you to set a range of available dates starting from the minimum date provided.

It could be pretty useful in case you have to handle departure / return fields using SingleDatePicker instead of DateRangePicker.

### DIFF
--- a/src/components/SingleDatePicker.jsx
+++ b/src/components/SingleDatePicker.jsx
@@ -44,8 +44,8 @@ const defaultProps = {
 
   isDayBlocked: () => false,
   disabledDays: [],
-  minDate: moment(),
-  isOutsideRange: (day, minDate) => !isInclusivelyAfterDay(day, minDate),
+  minimumDate: moment(),
+  isOutsideRange: (day, minimumDate) => !isInclusivelyAfterDay(day, minimumDate),
   enableOutsideDays: false,
   numberOfMonths: 2,
   orientation: HORIZONTAL_ORIENTATION,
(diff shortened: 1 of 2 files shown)


src/components/SingleDatePicker.jsx
```diff
@@ -44,7 +44,8 @@ const defaultProps = {
 
   isDayBlocked: () => false,
   disabledDays: [],
-  isOutsideRange: day => !isInclusivelyAfterDay(day, moment()),
+  minDate: moment(),
```
ljharb: letters are free, let's name this `minimumDate` :-)

majapw:
Wait, what exactly is this solving? Why couldn't you just do this by passing in a different `isOutsideRange` function?

Like if you want a minimumDate, couldn't you just do:
```
render() {
  const minimumDate = moment().add(1, 'week');
  return (
    <SingleDatePicker 
      ...
      isOutsideRange={(day) => day.isBefore(minimumDate)}
    />
  );
}
```

simonemonetti:
@majapw hi Maja and thanks! You're right, that should work too. 
However I still prefer to use a different (optional) prop rather than override `isOutsideRange` (which wasn't explained very clearly on the documentation).

majapw:
Hi @simonemonetti, I'd prefer not to add duplicate props to the components that accomplish the same thing. This seems like it is more of a documentation issue than an actual problem with the property. Let me think on how to rewrite the description of `isOutsideRange` (I've been meaning to revamp the README anyways).

DeanPoulin:
I was just implementing this functionality myself and tried a few things and got lost. Thanks @majapw for pointing me in the right direction on how to use isOutsideRange properly. Works perfectly, and the calendar doesn't allow the user to select dates earlier than the date I've defined. 👌

PR #174, opened 2016-11-09
removes check for touch devices when detecting disabled state

ViktoriyaSavkina:


majapw:
So the reason this check is here is not to pop up the keyboard on mobile devices. I was thinking about changing this to `readOnly={isTouchDevice}` to try and solve this issue. Can you check if the component works if you change it that way?

ViktoriyaSavkina:
@majapw you're right, adding readonly attribute prevents the keyboard from popping up. Updated PR to    set readonly for touch enabled devices

src/components/DateInput.jsx
```diff
@@ -121,7 +121,8 @@ export default class DateInput extends React.Component {
           placeholder={placeholder}
           autoComplete="off"
           maxLength={10}
-          disabled={disabled || this.isTouchDevice}
+          disabled={disabled}
+          readonly={this.isTouchDevice}
```
majapw: I think this needs to be camel case - https://facebook.github.io/react/docs/dom-elements.html#all-supported-html-attributes

ViktoriyaSavkina:
👍

Issue #175, opened 2016-11-10
currentMonth state is only reset once and never updated after that


ViktoriyaSavkina:
`currentMonth` state field along with `hasSetInitialVisibleMonth` flag are initialized in the constructor of the DatePicker component.
After `componentWillReceiveProps` is ran first, `hasSetInitialVisibleMonth` is set to true which causes the `currentMonth` field to never update, making it impossible to open datepicker with dates on different months.
https://github.com/airbnb/react-dates/blob/master/src/components/DayPicker.jsx#L100

majapw:
This is intentional behavior. Can you describe the usecase you need to change the month more than once for?

timhwang21:
Hi,

Based on this discussion, it sounds like it is intended for the calendar to always open on the same month, regardless of the currently selected values. Is that correct?

I noticed on Storybook, changing the month, selecting dates, closing the calendar, and reopening the calendar opens the picker based on the `startDate` selected.

On the other hand, in use the calendar seems to always open based on today's date, regardless of what value is selected, even after reducing the component down to the most minimal set of props possible.

I was wondering if I was missing something? If this is a bug, do you know where I could start looking to determine why my calendar is always opening based on today's date?

Thank you!

novartema:
Looks like it was fixed in new versions. I got same behavior in 7+ version. I installed 8.2.0 version and it works good.

PR #176, opened 2016-11-10
removes check if initial month was set before

ViktoriyaSavkina:
Fixes the issue on not being able to update current month more then once. https://github.com/airbnb/react-dates/issues/175

### DIFF
--- a/src/components/DayPicker.jsx
+++ b/src/components/DayPicker.jsx
@@ -76,7 +76,6 @@ export default class DayPicker extends React.Component {
   constructor(props) {
     super(props);
 
-    this.hasSetInitialVisibleMonth = !props.hidden;
     this.state = {
       currentMonth: props.hidden ? moment() : props.initialVisibleMonth(),
       monthTransition: null,
@@ -96,8 +95,7 @@ export default class DayPicker extends React.Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    if (!this.hasSetInitialVisibleMonth && !nextProps.hidden) {
-      this.hasSetInitialVisibleMonth = true;
+    if (!nextProps.hidden) {
       this.setState({
         currentMonth: nextProps.initialVisibleMonth(),
       });


ViktoriyaSavkina:
@travisbloom can you foresee any problems with this change?

majapw:
The original design of this property was that it could only be set once. I think we'd have to rename it from `initialVisibleMonth` to something else if we make this change.

I guess since this is only done when the `DayPicker` is hidden, this alleviates my worry of the weird transition issues of arbitrarily changing the month.

Can you describe the use case that you need this for?

ViktoriyaSavkina:
The use case is having multiple entries on a page with different dates associating with them. For example, in the screenshot below first record has a date in December and the second one in November. When clicking on either of them, date picker date changes to the appropriate date, but the initialMonth stays whatever it was set to first.
screen shot 2016-11-10 at 11 19 15 am
My code looks like this
```
<SingleDatePicker
            id="version-data-picker"
            anchorDirection={ANCHOR_LEFT}
            displayFormat="MM/DD/YY"
            numberOfMonths={1}
            focused={isDatePickerFocused}
            date={versionDate}
            initialVisibleMonth={() => versionDate}
            onDateChange={onDatePickerDateChange(date)}
            onFocusChange={onDatePickerFocusChange}
          />
```

ViktoriyaSavkina:
@majapw only worry with changing the name would be the major version bump needed since it's a breaking change. BTW how often do you guys version code changes in this repo?

ljharb:
Wouldn't each row have their own distinct date picker? It seems like you're reusing it across rows.

ViktoriyaSavkina:
@ljharb I am using a single datepicker component wrapped in a pure function component that gets date passed into it. I don't think I can control when react creates a new instance of datepicker vs re-rendering existing one. 

It seems like setting an arbitrary flag in constructor like this should be an anti-pattern. It doesn't  really do anything but ensures that initial month can be set only once. But even in this case it could be stored on state instead?

For airbnb use case, imagine an user opens datepicker, navigates 4 months ahead, and selects the date. Clicking on the datepicker input once will cause it to open on the selected month (4 months from the current day) for the first time. Then you navigate 4 months further from there and select another date. The condition in `componentWillRecieveProps` won't run anymore. So clicking on datepicker again will open it on the wrong month unless React will figure out to remount the component. And I don't think devs should rely on react internals.

@majapw yes the datepicker is hidden when date/month changes, so there is no weird transitions.

ViktoriyaSavkina:
@majapw would you like me to proceed with updating prop name to `currentMonth` or `visibleMonth`?

ViktoriyaSavkina:
Thinking about the naming some more, I think `initialVisibleMonth` prop name is still relevant, correct me if I'm wrong. @majapw @ljharb any update on status of this PR please? Is there anything else I need to do to get it merged?

ingro:
I think that this cause some problem.
For example here: http://airbnb.io/react-dates/?selectedKind=SingleDatePicker&selectedStory=with%20month%20specified%20on%20open&full=0&down=1&left=1&panelRight=0&downPanel=kadirahq%2Fstorybook-addon-actions%2Factions-panel

It's impossible to choose a date outside of the initial month.

ViktoriyaSavkina:
@ingro thanks for reporting this, i'll take a look shortly

ljharb:
(also reported as #186)

Issue #181, opened 2016-11-15
Cannot read property 'getBoundingClientRect' of null at DateRangePicker.responsivizePickerPosition


trongthanh:
Version: react-dates@4.0.0

How to reproduce:

- Render SingleDatePicker with portal: `<SingleDatePicker withPortal={true} />`
- Make sure the modal date picker overlay is closed.
- Resize the browser
- This error is thrown:
```
Uncaught TypeError: Cannot read property 'getBoundingClientRect' of null
    at DateRangePicker.responsivizePickerPosition ((index):54459)
```

AFAIK, The `dayPickerContainer` is not in the DOM when overlay is closed, so always get reference to it in the resize handler is naive. Besides, `componentWillUnmount`(where the resize hook is removed) doesn't work with the Portal transport when the picker is closed.

Ikana:
I get the same error, but on the storybook, under the section SingleDatePicker specifically under 
- horizontal with portal
- horizontal with fullscreen portal
- vertical with full screen portal

majapw:
Okay, I understand this problem. Responsivizing the daypicker doesn't make sense when we are using a portal (since the daypicker is not rendered when it is not showing) also because the daypicker is positioned in the center of the screen. We should conditionally call the function only when `withPortal` and `withFullScreenPortal` are both false.

PR #183, opened 2016-11-16
Remove invokation of responsivizePickerPosition from componentDidMount

Ikana:
This addressed the issue #181 on the storyboard, but should the method responsivizePickerPosition be invoked from an other method?

### DIFF
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -290,8 +290,8 @@ export default class DateRangePicker extends React.Component {
     if (!withPortal && !withFullScreenPortal) {
       const containerRect = this.dayPickerContainer.getBoundingClientRect();
       const currentOffset = dayPickerContainerStyles[anchorDirection] || 0;
-      const containerEdge = isAnchoredLeft ? containerRect[ANCHOR_RIGHT] :
-        containerRect[ANCHOR_LEFT];
+      const containerEdge =
+        isAnchoredLeft ? containerRect[ANCHOR_RIGHT] : containerRect[ANCHOR_LEFT];
 
       this.setState({
         dayPickerContainerStyles: getResponsiveContainerStyles(
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -283,23 +283,25 @@ export default class DateRangePicker extends React.Component {
   }
 
   responsivizePickerPosition() {
-    const { anchorDirection, horizontalMargin } = this.props;
+    const { anchorDirection, horizontalMargin, withPortal, withFullScreenPortal } = this.props;
     const { dayPickerContainerStyles } = this.state;
 
     const isAnchoredLeft = anchorDirection === ANCHOR_LEFT;
+    if (!withPortal && !withFullScreenPortal) {
+      const containerRect = this.dayPickerContainer.getBoundingClientRect();
+      const currentOffset = dayPickerContainerStyles[anchorDirection] || 0;
+      const containerEdge = isAnchoredLeft ? containerRect[ANCHOR_RIGHT] :
+        containerRect[ANCHOR_LEFT];
 
-    const containerRect = this.dayPickerContainer.getBoundingClientRect();
-    const currentOffset = dayPickerContainerStyles[anchorDirection] || 0;
-    const containerEdge = isAnchoredLeft ? containerRect[ANCHOR_RIGHT] : containerRect[ANCHOR_LEFT];
-
-    this.setState({
-      dayPickerContainerStyles: getResponsiveContainerStyles(
-        anchorDirection,
-        currentOffset,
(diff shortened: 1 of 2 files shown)


majapw:
This is not the correct solution. Based on my comment https://github.com/airbnb/react-dates/issues/181#issuecomment-261098119, let's change the `responsivizePickerPosition` in both files to be:

ljharb:
@Ikana please rebase on the command line so any merge commits are excised. Thanks!

Ikana:
I think I broke something the tests pass on my local repo.

Ikana:
I think it has to do with the recent update of react, the error is:

`Error: Cannot find module 'react/lib/CSSPropertyOperations'`

And on the blog post for `15.4.0` mentioned that the private apis did change

ljharb:
React 15.4 did break things; #184 will fix it.

Ikana:
so it wasn't me :sweat_smile:

ljharb:
Nope! rebase and you should be set.

src/components/SingleDatePicker.jsx
```diff
@@ -89,7 +89,6 @@ export default class SingleDatePicker extends React.Component {
   /* istanbul ignore next */
   componentDidMount() {
     window.addEventListener('resize', this.responsivizePickerPosition);
-    this.responsivizePickerPosition();
```
majapw: Can you make the same change as you did in `DateRangePicker` here?

majapw:
Once you make the last change, can you squash all your commit into one and rebase on the latest master?

Ikana:
Sure thing

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Change Log
 
+## v4.0.2
```
ljharb: @Ikana please do a *rebase* on the latest master rather than clicking "update branch", which creates a merge commit.

majapw: Hi @Ikana! Basically in your terminal you can run the following steps:
```
git checkout master
git pull airbnb master
git checkout fix-get-client-bounding-rect
git rebase master
git push -f origin fix-get-client-bound-rect
```

If you don't have airbnb configured as a remote branch, you'll have to do:
```
git remote add airbnb git@github.com:airbnb/react-dates.git
```

src/components/DateRangePicker.jsx
```diff
@@ -98,8 +98,11 @@ export default class DateRangePicker extends React.Component {
   }
 
   componentDidMount() {
+    const { withPortal, withFullScreenPortal } = this.props;
     window.addEventListener('resize', this.responsivizePickerPosition);
-    this.responsivizePickerPosition();
+    if (!withPortal && !withFullScreenPortal) {
```
majapw: We don't need this check here because we have it on line 293.

majapw:
@ljharb this should be good to go, can you do a last pass?

Ikana:
Hey @majapw thanks a lot for the patience and the help with git

majapw:
No problem @Ikana! :D Thank you so much for the contribution. :)

src/components/DateRangePicker.jsx
```diff
@@ -283,23 +283,25 @@ export default class DateRangePicker extends React.Component {
...
-        currentOffset,
-        containerEdge,
-        horizontalMargin
-      ),
-    });
+    if (!withPortal && !withFullScreenPortal) {
+      const containerRect = this.dayPickerContainer.getBoundingClientRect();
+      const currentOffset = dayPickerContainerStyles[anchorDirection] || 0;
+      const containerEdge =
+        isAnchoredLeft ? containerRect[ANCHOR_RIGHT] : containerRect[ANCHOR_LEFT];
```
ljharb: can this line be moved to the previous one, so there's no linebreak after an assignment operator?

src/components/SingleDatePicker.jsx
```diff
@@ -185,23 +185,26 @@ export default class SingleDatePicker extends React.Component {
...
-        currentOffset,
-        containerEdge,
-        horizontalMargin
-      ),
-    });
+    if (!withPortal && !withFullScreenPortal) {
+      const containerRect = this.dayPickerContainer.getBoundingClientRect();
+      const currentOffset = dayPickerContainerStyles[anchorDirection] || 0;
+      const containerEdge =
+        isAnchoredLeft ? containerRect[ANCHOR_RIGHT] : containerRect[ANCHOR_LEFT];
```

src/components/DateRangePicker.jsx
majapw: it's too long, and I preferred having the ternary on one line rather than splitting after the `?`. Which do you think is better?

ljharb: i always prefer a long line to arbitrary line breaks - ling length restrictions are always silly, and a poor proxy for readability.

PR #184, opened 2016-11-16
Update `react-portal` version

daltones:
`react-portal` v2 was relying on private React APIs and was broken with React v15.4.0.

It just simplified a few things in v3 and there's no breaking change for `react-dates`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,7 +97,7 @@
     "array-includes": "^3.0.2",
     "classnames": "^2.2.5",
     "react-moment-proptypes": "^1.2.0",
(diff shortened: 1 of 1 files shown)


ljharb:
What were the breaking changes in `react-portal`? It's unfortunate they couldn't fix that in v2.

daltones:
The breaking change is that you can't set `className` and `style` (the only props supported) to the portal container anymore. Instead you can simply wrap your container in another `<div>` with every prop you want.

See https://github.com/tajo/react-portal/pull/105

Issue #177, opened 2016-11-10
Width calculate bug in modal view


subicura:
If react-dates rendered in hidden element(like modal), initializeDayPickerWidth method set dayPickerWidth zero. So, next and prev method is not working normally.

This issue related with #163

With tether, date picker popup will render in body element. But now, react-dates render in hidden(modal) element.

Calculate element size method should be fixed.

I tested it with bootstrap-modal.

subicura:
$('.CalendarMonth')[0].offsetWidth => 0

majapw:
How are you showing/hiding your modal? Are you using `display: none` to do so? 

I am thinking that we could use `props.hidden` to help alleviate this problem. Maybe if instead of calling `initializeDayPickerWidth` on `componentDidMount`, we initalized it in `componentWillReceiveProps` and only when the DayPicker is about to be opened that would help to address this issue.

subicura:
Popup element uses `display: none` and `display: block`.

I'm thinking `props.hidden` is not good solution. Because I use same form in popup and plain page. Change prop solution is little complicated.

jQuery's `width()` method can calculate hidden(display:none) element's size correctly. I'm thinking change calculate method. It doesn't need any more code. (like props..)

majapw:
I don't really see us introducing jquery back into this library. `props.hidden` is not something that you as a consumer have to set, but rather is an internal prop to the `DayPicker` from the `DateRangePicker` based on whether or not the `DayPicker` is showing yet. If we call `initializeDayPickerWidth` when `hidden` is about to change from false to true instead of on `componentDidMount`, then I believe this would solve the problem. This could even be a fix for https://github.com/airbnb/react-dates/issues/46.

subicura:
OK. I understand. It's good. Thanks! :)

subicura:
any news??

majapw:
Hi @subicura, can you check if #196 helps?

subicura:
@majapw It works! Now, next and prev method is OK. Thanks! 👍

majapw:
Awesome. 👍 I will merge that in ASAP and do a release.

Issue #105, opened 2016-10-04
SingleDatePicker doesn't work within a form with other fields


dkwin:
Hey guys

```
I'm using SingleDatePicker along side with some bootstrap forms and it seems that the SingleDatePicker doesn't render fully.
```

I'm using React .15 with Moment 2.14 and React Dates 3.1.1

OnFocus

![react-datebug2](https://cloud.githubusercontent.com/assets/3934306/19060040/b34748c0-89b2-11e6-80d9-fbe6c85de79e.gif)

After Clicking Calendar

![react-datebug](https://cloud.githubusercontent.com/assets/3934306/19060014/79e279e2-89b2-11e6-8359-2bf355f1d0cc.gif)

majapw:
Hmm, I notice that you are using tabs on this interface. Are you controlling your tabs by rendering each time or are you showing/hiding them with `display: none` when they are hidden?

If you are hiding them with CSS, I wonder if this is related to https://github.com/airbnb/react-dates/issues/46.

bsr203:
has the same issue using with bootstrap 4. Once clicked the Calendar everything seems ok.
I render without a tab.

tvanro:
We also encountered the same issue and upgrading react-dates from 3.3.1 to 3.3.2 resolved it.
I'm pretty sure it's because of the tether library that is added in #83

savovs:
Having a similar issue when using DayPicker in a form with tabs that are hidden with `display: none`.
[http://i.imgur.com/qM9uxAv.gifv](url)

majapw:
I believe this should be fixed in v4.1.1. 

@dkwin @bsr203 @tvanro @savovs Please check it out and let me know if this is not the case.

Issue #46, opened 2016-08-30
DayPicker height incorrectly set if element is hidden when rendered


donaldharvey:
I'm using `DateRangePicker`, and the container I rendered it in had `display:none` initially. This meant that when `adjustDayPickerHeight()` was called, the calculated height was very small. When the container was then shown, the day picker retained this initial height and didn't function correctly. 

For now I've just made the component render when the container is shown, but it'd be good if this workaround wasn't necessary :)

ljharb:
The picker has no way of knowing when the container is suddenly made visible - the proper fix is indeed to only render the picker when it's going to be visible.

donaldharvey:
Right - but it'd surely be better, if possible, to adjust the height based on the component state and not the DOM, or to find a way to avoid explicitly setting the height altogether?

ljharb:
I don't know of any way to avoid that browser limitation here.

jonathanong:
i'm seeing the same issue, but i don't have `display: none` anywhere. i do use webpack and this is only the dev environment, so i'm thinking my styles aren't being applied until after the component mounts.

however, i noticed that the height is correctly adjusted when clicking prev/next months. could the daterangepicker adjust height when start/end date is selected, i.e. when `focusedInput` goes from `null` to non-null?

at render:

![screen shot 2016-08-31 at 7 50 48 pm](https://cloud.githubusercontent.com/assets/643505/18153774/7a7b56de-6fb4-11e6-999f-a50d01f03591.png)

after clicking next:

![screen shot 2016-08-31 at 7 50 53 pm](https://cloud.githubusercontent.com/assets/643505/18153775/7a7bbde0-6fb4-11e6-86e0-b851c88fa7d6.png)

jonathanong:
let me know if that makes sense, and i'll make a PR!

majapw:
Hi @jonathanong! That's surprising that you are seeing height issues without `display: none` anywhere. Can I see your set-up so we can see what the core issue is? I've never seen that without OP's original set-up.

As for the original problem, @donaldharvey I think that maybe once we have some sort of inline-style implementation where we can set the calendar day height and other details as props, we would be able to do the height calculation without reaching into the DOM. Unfortunately, we are not quite there yet.

jonathanong:
[...]

PR #196, opened 2016-11-22
Adjusts DayPicker height and initializes width when DayPicker is about to be shown

majapw:
This should hopefully fix https://github.com/airbnb/react-dates/issues/177 and also https://github.com/airbnb/react-dates/issues/105 and https://github.com/airbnb/react-dates/issues/46. Lol. Lots of dupes.

The idea is that while we use `visibility: hidden` in place of `display: none` when hiding the `DayPicker` so we can measure it properly, we can't actually guarantee that some parent isn't hiding the whole thing using `display: none` (in some modal implementations, for instance). This alleviates the issue somewhat by doing these DOM calculations right before we show the `DayPicker` (which can only happen when you click on or focus on a visible `DateRangePicker` component).

@subicura does this fix your issue?

Please take a look @airbnb/webinfra!

### DIFF
--- a/src/components/DayPicker.jsx
+++ b/src/components/DayPicker.jsx
@@ -96,16 +96,22 @@ export default class DayPicker extends React.Component {
   componentDidMount() {
     if (this.isHorizontal()) {
       this.adjustDayPickerHeight();
-      this.initializeDayPickerWidth();
     }
   }
 
   componentWillReceiveProps(nextProps) {
-    if (!this.hasSetInitialVisibleMonth && !nextProps.hidden) {
-      this.hasSetInitialVisibleMonth = true;
-      this.setState({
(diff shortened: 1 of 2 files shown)


src/components/DayPicker.jsx
```diff
@@ -96,16 +96,22 @@ export default class DayPicker extends React.Component {
...
     }
   }
 
   componentWillReceiveProps(nextProps) {
-    if (!this.hasSetInitialVisibleMonth && !nextProps.hidden) {
-      this.hasSetInitialVisibleMonth = true;
-      this.setState({
-        currentMonth: nextProps.initialVisibleMonth(),
-      });
+    if (!nextProps.hidden) {
```
ljharb: i wonder if perhaps `componentDidUpdate` would be a better place to do this computation, since that should fire every time the component rerenders?

PR #199, opened 2016-11-24
fix width when more than two months

### LINKED ISSUE #191: Weekday headers render issue
For some reason Saturday in the weekday headers renders below Sunday. See screenshots. I'm on version `3.6.0`
screen shot 2016-11-18 at 2 21 19 pm

SecurityInsanity:
fixes #191 (or I should say attempts to).

So this commit fixes weekday headers wrapping when have more than two months, with a couple caveats.

- This leaves a weird `${widthPercentage + 1}` inside the `width` part of a style block. I'm not a huge fan of this, because our width just needs to be a tiny bit larger, but adding that what anywhere such as:

  `const widthPercentage = (100 / numberOfMonths) + 1;` or `const widthPercentage = 101 / numberOfMonths;` all screw very much so with the calculation of `left`, and ends up becoming an uglier solution (IMO).

- The calculation of `left` seems to be off just a tiny bit for any odd month past 3. This is currently in master, as well I attempted to make it a little better, but that "little" better really is only a "little". However rendering more than 4 months seems very rare, and only would really work on screen resolutions greater than 1920.

Admittedly I'm not fully happy with this solution, but banging my head against a wall with math doesn't seem to be leading to any _better_ solutions. So I'm hoping someone will be smarter than I, and help point something out or build off the work I've done.

### DIFF
--- a/css/DayPicker.scss
+++ b/css/DayPicker.scss
@@ -41,16 +41,20 @@
 
   li {
     display: inline-block;
-    width: 39px;
+    width: 40px;
     text-align: center;
   }
 }
 
 .DayPicker--horizontal .DayPicker__week-header {
-  padding: 0 22px 0 13px;
+  padding: 0 22px 0 9px;
 
   &:first-of-type {
-    padding: 0 13px 0 22px;
+    padding: 0 13px 0 16px;
+  }
+
+  &:last-of-type {
+    padding: 0 13px 0 9px;
   }
 }
 
--- a/src/components/DayPicker.jsx
+++ b/src/components/DayPicker.jsx
@@ -314,7 +314,7 @@ export default class DayPicker extends React.Component {
 
     const widthPercentage = 100 / numberOfMonths;
     const horizontalStyle = {
-      width: `${widthPercentage}%`,
+      width: `${widthPercentage + 1}%`,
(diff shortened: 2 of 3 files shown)


majapw:
Sorry, I've been AWOL from this repo for a while. Do you think that this might be a border issue @SecurityInsanity? My particular concern about this issue is that it seems a bit hacky. :/ I still want to understand why this is happening before we move forward. I will devote some time to this while I'm punting performance reviews. :P

SecurityInsanity:
Hey @majapw , I've tried messing around with borders but admittedly I didn't want to make too many huge changes to the borders as sizings started changing in many places when I did that to get it to fit.

I 100% agree that my change is 100% hacky, I was merely uploading this to get a bit of foresight into why this happened. I can retake a look at the borders, and see if I can make something work regardless of all the things being changed, but if someone else figures it out before me that'd be awesome. This is seriously mind boggling.

majapw:
Hey @SecurityInsanity, so I poked around and kind of came to the conclusion that maybe we should just abandon the percentage thing. I mean we don't... technically need it? I put together a non-variable solution at https://github.com/airbnb/react-dates/pull/221

SecurityInsanity:
Closing out in favor of aforementioned #221

PR #205, opened 2016-12-02
v4 merge

SungwonKim:


package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "react-dates",
-  "version": "3.1.1",
+  "name": "trucking-react-dates",
```
ljharb: if you'd like to publish a fork, i'd recommend using scoped modules instead.

Issue #120, opened 2016-10-13
SingleDatePicker gets refocused if window put out and back into focus


jkudish:
If you open the SingleDatePicker and select a date, then go to a different app, and come back to the browser, and click anywhere on the page, the date picker re-opens even thought it wasn't set into focus itself.

This happens in the demo you have linked in the readme, so it's not specific to my implementation.

I'll look into fixing it myself next week if you don't get to it before me.

majapw:
Huh, that's weird. Happens with the DateRangePicker as well.

jkudish:
Any updates on this @majapw ?

majapw:
(We're always able to accept PRs though :) )

jkudish:
Thanks. I'll try and see if I can figure out the fix.

majapw:
Sorry, I have been out of the country so I haven't been able to really dig
into this. I might have some time this week, but likely it will have to
wait until the following week.

Valentin1918:
Does this bug was already fixed?

jkudish:
I can reproduce this bug on airbnb.com as well. I'm surprised it's not higher priority to fix this for you.

I agree @Valentin1918's fix is not ideal...

ljharb:
@jkudish it's being worked on as we speak.

It's not been a high priority because it doesn't actually block any functionality nor break anything, it's just an irritating extra click/tab to close the datepicker in the rare case when the last field focused was the date input, and you switch to another window, and switch back.

jkudish:
Understood. Glad you're working on it. I'm just circling back on it myself to see if there's a workaround.

I guess in my application it's a bit higher priority because of how we use it. We have it setup as a full screen date picker + a lot of tabbing in/out of that window because it's an internal project management app and users are often multi-tasking.

ljharb:
From @majapw and I's spelunking it seems like this is actual normal, expected browser behavior, so we're looking into implementing a workaround for this specific datepicker case - in other words, it's not a "bug" so much as an artifact of datepickers not being a well-integrated native UI, thus necessitating libraries like this :-)

jkudish:
Can you clarify what you mean by it being expected browser behavior? Now you've peaked my curiosity... 😉 

BTW, I appreciate the work you've done with this and that it's open source, definitely saves us a lot of time, thank you! 😄

majapw:
@jkudish PR incoming! :)

PR #210, opened 2016-12-12
Fixed auto focusing (bug #120)

Valentin1918:


ljharb:
How does this fox a bug? Can you add a test that covers the bug? Also, please don't just comment out code - if it's commented out it should be deleted.

Valentin1918:
If src/component/DateInput.jsx inside an input (line 120) was passed `onFocus={onFocus}` , so logically it was made knowingly, based on some reasons. That's why I decided not to delete it but comment. My particular case was exactly the same as discussed in bug #120. And this little change, fixed issue for me. It seems should be some additional parameter to pass in component, which will manage passing or not onFocus method in Input tag.

.npmignore
```diff
@@ -38,3 +38,4 @@ public
 react-dates-demo.gif
 stories
 test
+src
```
ljharb: Please revert this, the original source must be published to npm.

package.json
```diff
@@ -1,41 +1,13 @@
 {
-  "name": "react-dates",
-  "version": "4.1.0",
+  "name": "react-dates-custom",
```
ljharb: If you're going to publish a fork, please use scoped modules instead of polluting npm's global namespace. You can also npm install from git.

src/components/DateInput.jsx
```diff
@@ -117,7 +117,7 @@ export default class DateInput extends React.Component {
           value={value}
           onChange={this.onChange}
           onKeyDown={this.onKeyDown}
-          onFocus={onFocus}
+          //onFocus={onFocus}
```
majapw: So this is unlikely to be the solution given that this likely breaks keyboard navigation.

majapw:
This should be fixed by https://github.com/airbnb/react-dates/pull/212 in v4.1.1

src/components/DateInput.jsx
Valentin1918: Hmm, so it was a keyboard navigation (probably right-left arrows?), if yes, the problem occurred during swiping in Mac (using Ctrl+leftArrow/rightArrow). As an advice, you can use a controller, which will remove listener on yours keyUp event and add a default keyUp (if arrows are user with Ctrl), and reverse in case if arrows was used separately... something like that.

PR #213, opened 2016-12-14
Adding a today modifier to DayPickerRangeController

### LINKED ISSUE #211: Maybe bring back modifiers on DateRangePicker
I've seen that modifiers prop has been removed from DateRangePicker and now only exists on DayPicker. I was wondering if the prop could be brought back so it's easier to extend with custom classes and so on.

I wanted to use modifiers to add a "today" class but maybe that's a generic enough feature that it could be added by default instead? I dunno and leave that decision up to you guys.

ChristerGustavsson:
fixes #211

### DIFF
--- a/src/components/DayPickerRangeController.jsx
+++ b/src/components/DayPickerRangeController.jsx
@@ -107,10 +107,15 @@ export default class DayPickerRangeController extends React.Component {
     };
 
     this.isTouchDevice = isTouchDevice();
+    this.today = moment();
 
     this.onDayClick = this.onDayClick.bind(this);
     this.onDayMouseEnter = this.onDayMouseEnter.bind(this);
     this.onDayMouseLeave = this.onDayMouseLeave.bind(this);
+  }
+
+  componentWillUpdate() {
+    this.today = moment();
   }
 
   onDayClick(day, modifiers, e) {
@@ -223,6 +228,10 @@ export default class DayPickerRangeController extends React.Component {
     return isDayBlocked(day) || isOutsideRange(day) || this.doesNotMeetMinimumNights(day);
   }
 
+  isToday(day) {
+    return isSameDay(day, this.today);
+  }
+
   render() {
     const {
       isDayBlocked,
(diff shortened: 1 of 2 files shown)


src/components/DayPickerRangeController.jsx
```diff
@@ -223,6 +224,10 @@ export default class DayPickerRangeController extends React.Component {
     return isDayBlocked(day) || isOutsideRange(day) || this.doesNotMeetMinimumNights(day);
   }
 
+  isToday(day) {
+    return day.isSame(this.today, 'day');
```
majapw: Can you change this to use the `isSameDay` helper method? It should already be included in this file I think and it protects slightly better against falsey values

src/components/DayPickerRangeController.jsx
```diff
@@ -107,6 +107,7 @@ export default class DayPickerRangeController extends React.Component {
     };
 
     this.isTouchDevice = isTouchDevice();
+    this.today = moment();
```
ljharb: Whether something is a touch device won't change, but "today" will - imagine a component created at 11:59PM. Let's not cache this value. Maybe it should be a prop that's a function that defaults to returning `moment` - that way, it can be invoked every render, but can be overridden for tests.

majapw: I have a concern about the datepicker being open for a while, or someone using it across day boundaries. Let's leave it in the constructor but also maybe update it in `componentWillUpdate` (reset it as `this.today = moment()` again there.

majapw:
So this is perfect and will add a `CalendarMonth__day--today` class to the current day that you can style on. I think we can leave it with no styling in the default version.

src/components/DayPickerRangeController.jsx
majapw: I think just updating this value in `componentWillUpdate` will accomplish the same thing and create slightly fewer new moment objects... isToday is going to be called potentially hundreds of times per render.

majapw: But `componentWillUpdate` does get called on every render, right?

ljharb: Either that, or when different props and state will be rendered - I'm not 100% sure.

majapw:
Sweet! Can you rebase and the squash the commits down into one?

majapw:
Once @ljharb gives his thumbs up, we'll get this in! :)

PR #215, opened 2016-12-16
Adds back DayPickerWidth initialization on componentDidMount

### LINKED ISSUE #214: PrevMonth & NextMonth don't work correctly after upgrade to 4.1.1
After recently the updated to `4.1.1` from `4.1.0`, the `PrevMonth` & `NextMonth` arrows in `DayPicker` don't work as expected. Tested in Chrome (54.0.2840.98 64-bit) & Safari.

The `NextMonth` doesn't work at all (month won't change), whereas `PrevMonth` will cause 2 months overlay at first click and then stop working as well, like below.
screen shot 2016-12-15 at 4 51 34 pm

- However, in Chrome, if `click` (i guess because of the `focus` effect) within the days area first, they can work correctly.
- In Safari, even after `click` a day, the `Prev` & `Next` can only work once and initial month will never appear again (the element is still in DOM, but `data-visible` won't be set to true), for example, start with `Dec`, u can at most go next to `Jan` and then `Prev` will lead directly to `Nov`, vice versa.

Any known changes in 4.1.1 will cause above situation?

majapw:
I realize while that only doing this in `componentWillReceiveProps` while it works for a component that shows/hides the `DayPicker`, it does not get triggered before the `DayPicker` is visible if you are using that component directly. D: So uh, let's fix that. 

Fixes #214 

to: @airbnb/webinfra @yangaobo

### DIFF
--- a/src/components/DayPicker.jsx
+++ b/src/components/DayPicker.jsx
@@ -96,6 +96,7 @@ export default class DayPicker extends React.Component {
   componentDidMount() {
     if (this.isHorizontal()) {
       this.adjustDayPickerHeight();
+      this.initializeDayPickerWidth();
     }
   }
 


majapw:
We don't have DOM/mount tests yet on this repo. :/ Maybe we can set that up?

PR #216, opened 2016-12-19
Fix DateRangePicker not passing onDay handlers

omrimor:
Originally the DateRangerPicker did not map the onDay handlers correctly.
This passes the handlers properly to the next component.

### DIFF
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -181,6 +181,10 @@ export default class DateRangePicker extends React.Component {
       endDate,
       minimumNights,
       keepOpenOnDateSelect,
+      onDayMouseEnter,
+      onDayMouseLeave,
+      onDayMouseDown,
+      onDayTouchTap,
     } = this.props;
     const { dayPickerContainerStyles } = this.state;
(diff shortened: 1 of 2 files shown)


majapw:
Hi @omrimor. This was actually an error on my part from a past refactor. The `DateRangePicker` does not, by design, take these props (and in fact, even if we were to pass these on as you suggested, the `DayPickerRangeController` does nothing with them, so the point is moot). The correct solution is actually deleting lines 200 through 204 in the original code. I have that code ready in another commit.

If you would like to have your own logic for the `CalendarDay` user interactions, I would recommend using the `DayPicker` as is.

omrimor:
Hi @majapw. Thanks for the replay. I want to use the DateRangePicker in our project but extend the functionality based on our UX design requirement.

The logic there is exactly what we're looking for but we want to be able to use hooks for the mouse enter so we can perform added logic on hover.
I was hoping to avoid writing all the logic that is needed when using the DayPicker component.

also, when using the onPrevMonthClick/onNextMonthClick functions, is there a way to get current month viewed to preform added logic?

Issue #178, opened 2016-11-10
Need help/hints to debug this behaviour


ciaoben:
I don't know if it is a bug or I am missing something, but I have a pretty big react/redux application, and simply using the single date picker wrapper component from your examples.

The picker is nested pretty deeply in the Application tree (9/10 level of nesting) and I when I open it and hover with the mouse on the days my entire app became unusable. Here a screenshot from my timeline:

![schermata 2016-11-10 alle 17 07 49](https://cloud.githubusercontent.com/assets/4526061/20184266/6003b40e-a768-11e6-9692-acd4a4c59895.png)

As you can see every mouseout event it triggers a lot of work. Initially I though I was mistakenly attaching a state update to some event, so I stripped out everything and left only your original single date picker wrapper component. To be sure that no state was changing I monitored the root component of my application and the other components that listen the "redux" store and I am sure that none of them is rendering or receiving new props. They do no update. 

At this point I am clueless, any suggestion or advice? Can be a bug?

**UPDATE**: I used the react dev-tools to highlight the components that are updating, and it is clear that only the component is updating.... so how can be possible that exactly the same component of your example lags so much in the context of my application?

majapw:
Huh, this is really weird. There are def some performance improvements that I have been working on, but it's strange to me that nesting the component would somehow dramatically exacerbate the issue. I wonder if @lencioni has some thoughts on this. He knows a lot of things on web performance and has been helping me iterate on some improvements.

acp31:
Hey @ciaoben I had this same problem with my application. My problem was related to the isDayBlocked function. To set the hover state the application checks to see if the day is blocked, in order to do so the application needs to loop through all the blocked days on each hover to check if the current day is supposed to be blocked. This extra looping caused my application scripting time to skyrocket especially when I added in more blocked days.

My advice would be to take advantage of the constant time lookup of an object. Inside the componentDidMount function place an operation that takes all of your blocked days and places them in an object. Then in the isDayBlocked all you need to do is put an if/else statement that returns a boolean if the current hovered date is in the blocked dates object(you may need to format the dates so they match up)

I was actually going to suggest this as a feature, it seems that the extra looping can give the isDayBlocked function at worst a time complexity of O(n^2) which can be really slow. But with this suggestion the isDayBlocked function at absolute best can have a time complexity O(1) but most likely to be O(n), which is still faster than in the example that was used. 

[...]

Issue #140, opened 2016-10-21
Lots of issues


catalin-luntraru:
Not trying to be negative here, but the documentation is lacking the standard installation procedure.
I had to presume it's

`> npm install --save react-date react-addons-shallow-compare moment`

Any why would I even need `react-addons-shallow-compare`, it's not like I need to run the unit tests on this library myself.

Second, there's no documentation to set the code in your app. I had to figure it out myself that you have to write

```
import { SingleDatePicker } from 'react-dates';
import 'react-dates/css/variables.scss';
import 'react-dates/css/styles.scss';

<SingleDatePicker {...this.props} id="date_input" date={null} focused={false} onDateChange={this.onDateChange} onFocusChange={this.onFocusChange} />
```

And third, the div layer with the calendar is hidden so I had to disable `visibility: hidden` with chrome dev tools to at least see what this calendar is all about. Every time I would hover over a calendar date it would take 40-50% of my quad core CPU and I'm running Lubuntu, Probably on windows, the CPU would run out the window.

I'm amazed why this library has 2.5k stars, unless it's just that people know about airbnb.

TLDR: This library needs basic documentation for install and usage, not an api dump. And performance needs a serious check.

ljharb:
Regarding the first item, the documentation tells you exactly how to install - run the command [here](https://github.com/airbnb/react-dates#how-to-use).

catalin-luntraru:
I said it lacks a standard installation procedure.
I don't think this `npm install --save react-dates moment@>=#.## react@>=#.## react-dom@>=#.## react-addons-shallow-compare@>=#.##` is standard.
Or this:

```
 (
    export PKG=react-dates;
    npm info "$PKG" peerDependencies --json | command sed 's/[\{\},]//g ; s/: /@/g; s/ *//g' | xargs npm install --save "$PKG"
  )
```

ljharb:
@catalin-luntraru that latter script IS the standard installation procedure, because there does not exist any other way of automatically installing the proper peer dependencies into your package.json. The former, as it says, is an example of the command that the latter generates.

A naive `npm install --save` command for a module that has peerDependencies is highly likely to install the wrong versions, which is why that shell script exists.

ljharb:
As for `react-addons-shallow-compare`, that is required, at runtime, by React itself, so yes, you do need it installed.

majapw:
[...]

PR #217, opened 2016-12-19
Performance Improvements

majapw:
I'm marking this as breaking due to the removal of the modifiers prop from `CalendarDay`. It's a fairly minor change all in all, because it's not really being used at all throughout the library, but nevertheless breaking. 

This does a number of small performance improvements:
1. Opening the `DRP`/`SDP` no longer causes 4 renders (I had restructured the `OutsideClickHandler` logic when the tether was introduced and this basically reverts to the pre-tether world). 
2. The `DateInput` only has an `onFocus` action instead of an `onFocus` action on the input and an `onClick` on the input's wrapper. Previously, when we were disabling the input on touch screens, this was a necessary action. However, `readonly` inputs do still receive focus events on touch devices afaict so this is no longer necessary.
3. The `CalendarMonthGrid` component now keeps track of the month moment objects used for its children in its state and only updates them as necessary (instead of creating new moment objects on every render).
4. The `CalendarMonth` component also now keeps track of the day moment objects used for its children in its state and only updates them as necessary.
5. The `CalendarDay` no longer takes a modifiers prop. It was only using these props in its user interaction callbacks and it was hard not to rerender each time as the modifiers array was being reinstantiated each time.

I used `react-addons-perf` to track the improvements. The following wasted time printouts are from 5 seconds of basic interaction with the component.

**Before:**
screen shot 2016-12-19 at 8 28 04 am

**After:**
screen shot 2016-12-19 at 8 26 58 am

There's likely still some work that can be done, but this seems like a good start.

Should help with https://github.com/airbnb/react-dates/issues/178 and https://github.com/airbnb/react-dates/issues/140

to: @airbnb/webinfra @ljharb @lencioni

### DIFF
--- a/src/components/CalendarDay.jsx
+++ b/src/components/CalendarDay.jsx
@@ -7,7 +7,6 @@ export const TOUCHSTART_TIMEOUT = 200;
 
 const propTypes = {
   day: momentPropTypes.momentObj,
-  modifiers: PropTypes.arrayOf(PropTypes.string),
   onDayClick: PropTypes.func,
   onDayMouseDown: PropTypes.func,
   onDayMouseUp: PropTypes.func,
@@ -20,7 +19,6 @@ const propTypes = {
 
 const defaultProps = {
   day: moment(),
-  modifiers: [],
   onDayClick() {},
   onDayMouseDown() {},
   onDayMouseUp() {},
@@ -41,61 +39,61 @@ export default class CalendarDay extends React.Component {
     return shallowCompare(this, nextProps, nextState);
   }
 
-  handleDayClick(day, modifiers, e) {
-    this.props.onDayClick(day, modifiers, e);
+  onDayClick(day, e) {
+    this.props.onDayClick(day, e);
   }
 
-  handleDayMouseDown(day, modifiers, e) {
-    this.props.onDayMouseDown(day, modifiers, e);
+  onDayMouseDown(day, e) {
+    this.props.onDayMouseDown(day, e);
   }
 
-  handleDayMouseUp(day, modifiers, e) {
-    this.props.onDayMouseUp(day, modifiers, e);
+  onDayMouseUp(day, e) {
+    this.props.onDayMouseUp(day, e);
   }
 
-  handleDayMouseEnter(day, modifiers, e) {
-    this.props.onDayMouseEnter(day, modifiers, e);
+  onDayMouseEnter(day, e) {
+    this.props.onDayMouseEnter(day, e);
   }
 
-  handleDayMouseLeave(day, modifiers, e) {
-    this.props.onDayMouseLeave(day, modifiers, e);
+  onDayMouseLeave(day, e) {
+    this.props.onDayMouseLeave(day, e);
   }
 
-  handleDayTouchStart(day, modifiers, e) {
+  onDayTouchStart(day, e) {
     this.hasActiveTouchStart = true;
     setTimeout(() => {
       this.hasActiveTouchStart = false;
     }, TOUCHSTART_TIMEOUT);
 
-    this.props.onDayTouchStart(day, modifiers, e);
+    this.props.onDayTouchStart(day, e);
   }
 
-  handleDayTouchEnd(day, modifiers, e) {
+  onDayTouchEnd(day, e) {
     if (this.hasActiveTouchStart) {
       this.hasActiveTouchStart = false;
-      this.handleDayTouchTap(day, modifiers, e);
+      this.onDayTouchTap(day, e);
     }
 
-    this.props.onDayTouchEnd(day, modifiers, e);
+    this.props.onDayTouchEnd(day, e);
   }
 
-  handleDayTouchTap(day, modifiers, e) {
-    this.props.onDayTouchTap(day, modifiers, e);
+  onDayTouchTap(day, e) {
+    this.props.onDayTouchTap(day, e);
   }
 
   render() {
-    const { day, modifiers } = this.props;
+    const { day } = this.props;
 
     return (
       <div
         className="CalendarDay"
-        onMouseEnter={(e) => { this.handleDayMouseEnter(day, modifiers, e); }}
-        onMouseLeave={(e) => { this.handleDayMouseLeave(day, modifiers, e); }}
-        onMouseDown={(e) => { this.handleDayMouseDown(day, modifiers, e); }}
-        onMouseUp={(e) => { this.handleDayMouseUp(day, modifiers, e); }}
-        onClick={(e) => { this.handleDayClick(day, modifiers, e); }}
-        onTouchStart={(e) => { this.handleDayTouchStart(day, modifiers, e); }}
-        onTouchEnd={(e) => { this.handleDayTouchEnd(day, modifiers, e); }}
+        onMouseEnter={e => this.onDayMouseEnter(day, e)}
+        onMouseLeave={e => this.onDayMouseLeave(day, e)}
+        onMouseDown={e => this.onDayMouseDown(day, e)}
+        onMouseUp={e => this.onDayMouseUp(day, e)}
+        onClick={e => this.onDayClick(day, e)}
+        onTouchStart={e => this.onDayTouchStart(day, e)}
+        onTouchEnd={e => this.onDayTouchEnd(day, e)}
       >
         <span className="CalendarDay__day">{day.format('D')}</span>
       </div>
--- a/src/components/CalendarDay.jsx
+++ b/src/components/CalendarDay.jsx
@@ -40,45 +40,53 @@ export default class CalendarDay extends React.Component {
   }
 
   onDayClick(day, e) {
-    this.props.onDayClick(day, e);
+    const { onDayClick } = this.props;
+    onDayClick(day, e);
   }
 
   onDayMouseDown(day, e) {
-    this.props.onDayMouseDown(day, e);
+    const { onDayMouseDown } = this.props;
+    onDayMouseDown(day, e);
   }
 
   onDayMouseUp(day, e) {
-    this.props.onDayMouseUp(day, e);
+    const { onDayMouseUp } = this.props;
+    onDayMouseUp(day, e);
   }
 
   onDayMouseEnter(day, e) {
-    this.props.onDayMouseEnter(day, e);
+    const { onDayMouseEnter } = this.props;
+    onDayMouseEnter(day, e);
   }
 
   onDayMouseLeave(day, e) {
-    this.props.onDayMouseLeave(day, e);
+    const { onDayMouseLeave } = this.props;
+    onDayMouseLeave(day, e);
   }
 
   onDayTouchStart(day, e) {
+    const { onDayTouchStart } = this.props;
     this.hasActiveTouchStart = true;
     setTimeout(() => {
       this.hasActiveTouchStart = false;
     }, TOUCHSTART_TIMEOUT);
 
-    this.props.onDayTouchStart(day, e);
+    onDayTouchStart(day, e);
   }
 
   onDayTouchEnd(day, e) {
+    const { onDayTouchEnd } = this.props;
     if (this.hasActiveTouchStart) {
       this.hasActiveTouchStart = false;
       this.onDayTouchTap(day, e);
     }
 
-    this.props.onDayTouchEnd(day, e);
+    onDayTouchEnd(day, e);
   }
 
   onDayTouchTap(day, e) {
-    this.props.onDayTouchTap(day, e);
+    const { onDayTouchTap } = this.props;
+    onDayTouchTap(day, e);
   }
 
   render() {
--- a/src/components/CalendarMonth.jsx
+++ b/src/components/CalendarMonth.jsx
@@ -112,14 +112,14 @@ export default class CalendarMonth extends React.Component {
           <tbody className="js-CalendarMonth__grid">
             {weeks.map((week, i) => (
               <tr key={i}>
-                {week.map((day, j) => {
+                {week.map((day, dayOfWeek) => {
                   const modifiersForDay = getModifiersForDay(modifiers, day);
                   const className = cx('CalendarMonth__day', {
                     'CalendarMonth__day--outside': !day || day.month() !== month.month(),
                   }, modifiersForDay.map(mod => `CalendarMonth__day--${mod}`));
 
                   return (
-                    <td className={className} key={j}>
+                    <td className={className} key={dayOfWeek}>
                       {day &&
                         <CalendarDay
                           day={day}
--- a/src/components/CalendarMonth.jsx
+++ b/src/components/CalendarMonth.jsx
@@ -1,6 +1,7 @@
 /* eslint react/no-array-index-key: 0 */
 
 import React, { PropTypes } from 'react';
+import shallowCompare from 'react-addons-shallow-compare';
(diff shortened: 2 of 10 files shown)


src/components/CalendarDay.jsx
```diff
@@ -41,61 +39,61 @@ export default class CalendarDay extends React.Component {
     return shallowCompare(this, nextProps, nextState);
   }
 
-  handleDayClick(day, modifiers, e) {
-    this.props.onDayClick(day, modifiers, e);
+  onDayClick(day, e) {
+    this.props.onDayClick(day, e);
```
ljharb: Not a big deal, but it'd be good to destructure out the prop and then call it, so that its `this` value won't be `this.props` (same throughout)

src/components/CalendarMonthGrid.jsx
```diff
@@ -73,6 +76,23 @@ export default class CalendarMonthGrid extends React.Component {
...
 
+  componentWillReceiveProps(nextProps) {
+    const { initialMonth, numberOfMonths } = nextProps;
+    const { months } = this.state;
+
+    const hasMonthChanged = !this.props.initialMonth.isSame(initialMonth, 'month');
+    const hasNumberOfMonthsChanged = this.props.numberOfMonths !== numberOfMonths;
+    if (hasMonthChanged || hasNumberOfMonthsChanged) {
+      if (initialMonth.isAfter(this.props.initialMonth)) {
+        months.shift();
```
ljharb: This is mutating the state array directly, rather than using setState - is this intentional? Can we avoid the mutation?

src/components/CalendarMonthGrid.jsx
```diff
@@ -159,7 +163,29 @@ export default class CalendarMonthGrid extends React.Component {
         style={getTransformStyles(transformValue)}
         onTransitionEnd={onMonthTransitionEnd}
       >
-        {months}
+        {months.map((month, i) => {
+          const isVisible =
+            (i >= firstVisibleMonthIndex) && (i < firstVisibleMonthIndex + numberOfMonths);
+          return (
+            <CalendarMonth
+              key={`CalendarMonth-${i}`}
```
ljharb: This key just needs to be `i` - ie, only unique within the array.

src/components/DateRangePicker.jsx
```diff
@@ -263,32 +265,36 @@ export default class DateRangePicker extends React.Component {
       onFocusChange,
     } = this.props;
 
+    const onOutsideClick = !withPortal && !withFullScreenPortal ? this.onOutsideClick : () => {};
```
ljharb: It would be more performant to cache this noop function outside the render path, so react can know when it's the same object. Alternatively, set it to null/undefined, and it won't be passed at all?

src/components/SingleDatePicker.jsx
```diff
@@ -325,27 +325,31 @@ export default class SingleDatePicker extends React.Component {
 
     const dateString = this.getDateString(date);
 
+    const onOutsideClick = !withPortal && !withFullScreenPortal ? this.onOutsideClick : () => {};
```
ljharb: Same comment here re noop vs undefined

src/components/CalendarDay.jsx
```diff
@@ -41,61 +39,61 @@ export default class CalendarDay extends React.Component {
...
-        onClick={(e) => this.handleDayClick(day, modifiers, e)}
-        onTouchStart={(e) => this.handleDayTouchStart(day, modifiers, e)}
-        onTouchEnd={(e) => this.handleDayTouchEnd(day, modifiers, e)}
+        onMouseEnter={(e) => this.onDayMouseEnter(day, e)}
+        onMouseLeave={(e) => this.onDayMouseLeave(day, e)}
+        onMouseDown={(e) => this.onDayMouseDown(day, e)}
+        onMouseUp={(e) => this.onDayMouseUp(day, e)}
+        onClick={(e) => this.onDayClick(day, e)}
+        onTouchStart={(e) => this.onDayTouchStart(day, e)}
+        onTouchEnd={(e) => this.onDayTouchEnd(day, e)}
```
lencioni: If you are going for performance, you might also want to consider binding these event handlers in the constructor, and then re-binding only when `this.props.day` changes.

src/components/CalendarMonth.jsx
```diff
@@ -53,71 +54,93 @@ export function getModifiersForDay(modifiers, day) {
...
+    const { weeks } = this.state;
+    const monthTitle = month.format(monthFormat);
+
+    const calendarMonthClasses = cx('CalendarMonth', {
+      'CalendarMonth--horizontal': orientation === HORIZONTAL_ORIENTATION,
+      'CalendarMonth--vertical': orientation === VERTICAL_ORIENTATION,
+    });
+
+    return (
+      <div className={calendarMonthClasses} data-visible={isVisible}>
```
lencioni: unrelated: in addition to `data-visible` (or maybe instead of it?) we might want to use `aria-hidden`.

src/components/CalendarMonth.jsx
```diff
@@ -53,71 +54,93 @@ export function getModifiersForDay(modifiers, day) {
...
+        <table>
+          <caption className="CalendarMonth__caption js-CalendarMonth__caption">
+            <strong>{monthTitle}</strong>
+          </caption>
+
+          <tbody className="js-CalendarMonth__grid">
+            {weeks.map((week, i) =>
+              <tr key={i}>
```
lencioni: Array index keys should generally be avoided. I think it might be okay in this case, but it would still be better to use something that identifies the `week` object instead if that is possible.

src/components/CalendarMonth.jsx
```diff
@@ -53,71 +54,93 @@ export function getModifiersForDay(modifiers, day) {
...
+                {week.map((day, j) => {
+                  const modifiersForDay = getModifiersForDay(modifiers, day);
+                  const className = cx('CalendarMonth__day', {
+                    'CalendarMonth__day--outside': !day || day.month() !== month.month(),
+                  }, modifiersForDay.map(mod => `CalendarMonth__day--${mod}`));
+
+                  return (
+                    <td className={className} key={j}>
```
lencioni: Same comment here regarding array index and `j`.

src/components/CalendarMonth.jsx
```diff
@@ -53,71 +54,93 @@ export function getModifiersForDay(modifiers, day) {
...
+    const { month, enableOutsideDays } = nextProps;
+    if (!month.isSame(this.props.month)) {
+      this.setState({
+        weeks: getCalendarMonthWeeks(month, enableOutsideDays),
+      });
+    }
+  }
+
+  shouldComponentUpdate(nextProps, nextState) {
+    return shallowCompare(this, nextProps, nextState);
```
lencioni: How much benefit does this give you now that you've moved the moment objects to state?

src/components/CalendarMonthGrid.jsx
lencioni: This definitely needs to be changed. You should use the callback version of setState:

```js
this.setState((previousState) => {
  ...
  return {...};
});
```

src/components/CalendarMonthGrid.jsx
```diff
@@ -95,11 +115,22 @@ export default class CalendarMonthGrid extends React.Component {
     this.props.onMonthTransitionEnd();
   }
 
+  getMonths(initialMonth, numberOfMonths) {
+    let month = initialMonth.clone().subtract(1, 'month');
```
lencioni: As this function is written, it is a little difficult to grok. To avoid doing mental acrobatics, what do you think about moving this into your loop? It shouldn't be too hard if methods like `add` or `subtract` can take negative numbers.

src/components/CalendarDay.jsx
ljharb: oh actually, that's a good point - they can be bound in the constructor and simply look up `this.props.day` every time, so there's no need for any rebinding.

lencioni: Ah, yes of course!

ljharb: (they actually could have done this in the first place, too)

sontek:
@majapw Are you still working on this?  We just finished our implementation of `react-dates` but are holding off on shipping it live because of the bad performance even on iphone6.  I think this PR might solve that for us

sontek:
@majapw Also, I noticed airbnb just uses another picker on mobile.  Is that because of performance or were there other criteria they used when doing that?

Wondering if I should be following the same workflow?

ljharb:
@sontek any place we're not using react-dates is because we haven't migrated it over yet (i.e., legacy)

majapw:
Hi @sontek! Sorry for the delay! I've been on vacation. I'll be returning to this this week and hopefully shipping it. :)

sontek:
@majapw No worries on the delay, everyone deserves a vacation! :)  I'm just excited to see work on the performance since its the only thing holding us back on using this!

sontek:
@majapw Would it make sense to break up some of these changes into smaller pieces so it'll be easier to get some of the changes into a release?

src/components/CalendarMonthGrid.jsx
lencioni: setState is async, so if you need to modify state based on the previous state, you need to use the callback version of setState otherwise it may end up doing an incorrect thing, which could be a subtle cause of bugs.

lencioni: Of course, if you don't need anything from the previous state, then you don't need to use the callback version.

src/components/CalendarMonthGrid.jsx
lencioni: Ideally this would not include the index at all and would be based off of unique datae. But what you have here is functionally equivalent to a plain index with unnecessarily added complexity of the string.

src/components/CalendarMonthGrid.jsx
majapw: This is in `componentWillReceiveProps` and is directly mutating the current state which might be a problem, but is doing so based on the change in props not on a change in state.

I don't think we need the callback, but I can figure out a way to do this with `setState` instead of mutating the state under the hood.

src/components/CalendarMonth.jsx
lencioni: `day` is some sort of moment object, right? I'd go with something like this `key={day.dayOfYear()}`. I think it might be more important for the `tr` to have a better key than it is for this `td` to have a better key, but you might as well make them all better at the same time.

majapw: It is not always a moment object. Sometimes it is `null` and also @ljharb mentioned that these only have to be unique within the array, no?

ljharb: yes, `key` only needs to be unique within the array. That said, when it *is* a moment object, making it something actually unique and tied to the row might make things more performant, even if some of the items use an index.

src/components/CalendarMonth.jsx
majapw: It still diminishes the render count significantly so I think it is meaningful to leave this.

src/components/CalendarMonth.jsx
majapw: Since some of the day variables may equal null, how do you propose using something different for the moment objects and not having it collide with the keys for the null values? I think having it tied to day of week makes the most sense.

ljharb: Day of the week, and index, seem like they wouldn't conflict - alternatively, `'day_' + dayNumber` wouldn't collide either.

src/components/CalendarMonth.jsx
```diff
@@ -53,71 +54,93 @@ export function getModifiersForDay(modifiers, day) {
...
+        <table>
+          <caption className="CalendarMonth__caption js-CalendarMonth__caption">
+            <strong>{monthTitle}</strong>
+          </caption>
+
+          <tbody className="js-CalendarMonth__grid">
+            {weeks.map((week, i) => (
+              <tr key={i}>
+                {week.map((day, j) => {
+                  const modifiersForDay = getModifiersForDay(modifiers, day);
```
timhwang21: I'm not sure how much you guys like comments from randoms on your PRs, but I'm just wondering: as I mentioned in #245, is it possible to run the calculations for days conditionally? As I understand it, even months not currently displayed are having their modifiers calculated. Is there any reason this isn't being done just for visible months? Thanks!

majapw: If you don't run the calculations for non-visible months, you get weird behavior for transitions. For example, if part of the selected range is in the month before or the month after, it will look unselected during the transition and then flash to the selected teal color.

src/components/CalendarMonth.jsx
majapw: `j` is already day of the week, like it goes from 0 - 6. I guess i don't understand what the problem is with this key...

ljharb: in that case maybe the only problem is the variable name :-)

src/components/CalendarMonth.jsx
timhwang21: Would visible months ±1 on either side work then? I don't imagine many use cases would want to display more than 3 calendars at once, and running the calculations for 5 calendars sounds more appealing than all 12.

majapw: That is the current behavior @timhwang21

majapw:
@ljharb @lencioni I have addressed all comments. Can one of you give me a stamp? I'd like to get this merged in today.

majapw:
Almost all of the additions to the code are in lifecycles methods. I think I'm going to invest some time into getting the full mount tests working, but I'll probably merge through this coveralls check.

src/components/DateRangePicker.jsx
```diff
@@ -186,7 +192,7 @@ export default class DateRangePicker extends React.Component {
     } = this.props;
     const { dayPickerContainerStyles } = this.state;
 
-    const onOutsideClick = !withFullScreenPortal ? this.onOutsideClick : undefined;
+    const onOutsideClick = !withFullScreenPortal && withPortal ? this.onOutsideClick : undefined;
```
ljharb: This is fine but not wrapping the condition in parens makes me nervous.

src/components/DateRangePicker.jsx
```diff
@@ -263,32 +265,36 @@ export default class DateRangePicker extends React.Component {
       onFocusChange,
     } = this.props;
 
+    const onOutsideClick = !withPortal && !withFullScreenPortal ? this.onOutsideClick : undefined;
```
ljharb: also here

src/components/SingleDatePicker.jsx
```diff
@@ -276,7 +276,7 @@ export default class SingleDatePicker extends React.Component {
       selected: day => this.isSelected(day),
     };
 
-    const onOutsideClick = !withFullScreenPortal ? this.onClearFocus : undefined;
+    const onOutsideClick = !withFullScreenPortal && withPortal ? this.onClearFocus : undefined;
```
ljharb: and here

src/components/SingleDatePicker.jsx
```diff
@@ -336,27 +336,31 @@ export default class SingleDatePicker extends React.Component {
 
     const dateString = this.getDateString(date);
 
+    const onOutsideClick = !withPortal && !withFullScreenPortal ? this.onOutsideClick : undefined;
```
ljharb: and here

PR #220, opened 2016-12-19
Fixes issue where a system without autoprefixing will not allow for month transitions in safari

### LINKED ISSUE #195: DateRangePicker not working correctly on iOS8 iPad iPhone
In the midst of device testing, we're experiencing a strange issue with the DateRangePicker in mobile Safari on iOS8 on iPad/iPhone only.

When tapping the nex/prev month buttons the calendar slides only one month back or forward and no further.  Is anyone else seeing this or have any suggestions to debug?

Thanks

Al

majapw:
Fixes #195 

So I believe that adding this prefixes on the transition css prop should fix the issue @awhenderson was having where their implementation would not transition properly in Safari. I think that we have an autoprefixer at airbnb that automagically does this for us, but given that not everyone has this set-up, this should patch the problem for now. 

@awhenderson let me know if this fixes your issue! 

to: @airbnb/webinfra

### DIFF
--- a/css/CalendarMonthGrid.scss
+++ b/css/CalendarMonthGrid.scss
@@ -6,6 +6,8 @@ $react-dates-width-day-picker: 300px;
 }
 
 .CalendarMonthGrid--animating {
+  -webkit-transition: transform 0.2s ease-in-out;
+  -moz-transition: transform 0.2s ease-in-out;
   transition: transform 0.2s ease-in-out;
   z-index: 1;
 }


majapw:
Yeah. I figure we can push this out as a quick fix and then I can put
together another PR to add autoprefixing directly in the repo. :)

lpk3nny:
-webkit-transition: **-webkit-**transform 0.2s ease-in-out;

lpk3nny:
i mean, it will be fixed by autoprefixer of course, but in this quick fix u need to use prefixed transform

awhenderson:
Thanks @majapw, @ljharb, @lpk3nny tried prefixed transform and no change in problem for me, is it worth trying an autoprefixer tool?  Any node package recommendations?

awhenderson:
Disregard my last comment - now working for me.  

Thanks @majapw, @ljharb, @lpk3nny

majapw:
Jokes @lpk3nny is right. The transition is not the problem so much as the transform. I will update that. Will make that change.

majapw:
Did it in https://github.com/airbnb/react-dates/commit/92c2cfdb5fe4221defc6e74d5e2d01d0e6c6f461. Going to do a new release now.

PR #221, opened 2016-12-19
Fix week header width issue over 2 months

majapw:
So I started thinking about https://github.com/airbnb/react-dates/issues/191 as well as @SecurityInsanity's solution https://github.com/airbnb/react-dates/pull/199

I think the end that I came to was that... the percentage code was trying too hard. The best bet for the week headers would be to get the style to match the actual `CalendarMonth`s directly, right?

Well, turns out we have access to the width we're using right now both in the css and in the JS (incidentally, it's not very moddable at presence), so let's just go with that. This change does that and thus allows for like a real, > 2 months implementation without looking gross.

to: @SecurityInsanity @airbnb/webinfra

### DIFF
--- a/css/DayPicker.scss
+++ b/css/DayPicker.scss
@@ -26,11 +26,17 @@
   position: relative;
 }
 
+.DayPicker--horizontal .DayPicker__week-headers {
+  margin-left: 9px;
+}
+
 .DayPicker__week-header {
   color: $react-dates-color-placeholder-text;
   position: absolute;
+  width: $react-dates-width-day-picker;
   top: 62px;
   z-index: 2;
+  padding: 0 13px;
   text-align: left;
 
   ul {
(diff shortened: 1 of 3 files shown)


SecurityInsanity:
I actually like this idea a lot better. Rather than trying to calculate this weird percentage (where I was still caught up). Seems a lot better to do it this way.

I noticed you added a story for 3 months, I assume (based on looking at this PR), that this should scale to >4 months too, which is always cool sauce.

I think I'll go ahead, and close #199 out in favor of this one.

majapw:
Awesome. Thanks for the vote of confidence (and doing the hard work of determining the percentages are a bad idea X_x)!

And yeah, I checked it at 4 months and it works as well, but it didn't seem necessary to add both stories. 

🎉 incidentally, i'm on a PR/issue rooollll today. #vacation

SecurityInsanity:
Yea I think adding both stories is a bit unnecessary as long as it shows >2 months.

As for the vacation you aren't alone! I think I made all of my PRs except #199 to this repo while on vacation as well. :tada: #vacation (clearly the PR didn't end up working out because I wasn't on vacation.)

PR #222, opened 2016-12-19
Adding an input icon to the begining of the DateRangePickerInput

AntiFish03:
I needed a simple icon added to the beginning of the Date Range Picker.  And I thought that others might also like a simple decoration / interaction point.

showCalendarIcon defaults to false, added using
```
<DatePicker
    showCalendarIcon
/>
```
Interaction uses `onStartDateFocus`

Calendar Icon was extracted from font-awesome 4.7.0 licensing is fully open source, http://fontawesome.io/license/

Added storybook entry, Readme entry and tests.


lencioni:
I imagine that people will want to be able to customize the icon to match their sites. Instead of baking the icon in, it seems better to allow a component to be passed in that would enable people to use whatever decoration they wanted. What do you think @majapw?

AntiFish03:
I can change that if needed but I just followed the choices made with the arrow and the close button of baking them in.

majapw:
Hi @AntiFish03! I think it makes sense to have it be configurable a la the month navigation buttons but maybe default to the SVG you've provided. Can you also post a screenshot of how this looks?

Thanks!

AntiFish03:
Here is the screenshot.
image

AntiFish03:
Also, I am more than happy to make the requested change give me a few minutes to adjust it.

README.md
```diff
@@ -177,9 +177,11 @@ If the `showClearDates` prop is set to true, an `x` shows up in the input box th
   showClearDates: PropTypes.bool,
 ```
 
-If the `showCalendarIcon` prop is set to true, a calendar icon is displayed at the start of the input box. Interaction is the same as focusing the start date.
+If the `showInputIcon` prop is set to true, a calendar icon is displayed at the start of the input box. Interaction is the same as focusing the start date.
+Optionally, it will display a React node set to `props.customInputIcon`
 ```
-  showCalendarIcon: PropTypes.bool
+  showInputIcon: PropTypes.bool,
+  customInputIcon: PropTypes.node
```
ljharb: could we just have a `inputIcon` prop that's either `null` or a `node`? I'm not sure I see the point in having both.

AntiFish03: > but maybe default to the SVG you've provided.

If it is both optional and customizable it needs to know when to display, when to display the default, and when to display the custom icon.

AntiFish03:
I am happy to adjust this however it needs to be to allow the input icon.

Whether there is a baked in default or customizable, or both. Just need to know which way to take the code.

README.md
ljharb: ah, i didn't realize there'd be an icon included by default. Might it be simpler to go with a single `inputIcon` prop, and then export the default icon so that users can pass that in if they want it?

AntiFish03: Possibly... Its your call. The icon is an export out of the Font Awesome SVG, under its open license. So its definitely not a custom calendar icon.

However, this is more simple if you are just going to use the basic input icon because trying to use the traditional method of adding to the HTML below and adjusting its positioning could create user interaction difficulties (i.e the icon steals the click).

And this version has a customizable capability that someone with a truly custom calendar icon could use without difficulty.

ljharb: true - i'm torn. We could do `PropTypes.oneOfType(PropTypes.oneOf([true]), PropTypes.node)` but that seems icky.

AntiFish03: Wouldn't `PropTypes.OneOfType([PropTypes.bool, PropTypes.node])` work.  It would be less exact but its also less `icky`

AntiFish03: And thinking about it a little further that would actually be more exact not less, Because false, true, node should all be valid values. We are just choosing to think of it as null, true, node.

majapw: Hmm, whatever we choose to use here will likely also end up influencing how we implement customization on the clear dates SVG.

AntiFish03: I am happy to make changes in what ever direction... I have a need for this and would rather assist the community using this than maintain a fork...

majapw:
@AntiFish03 
I think the best step would be to leave it as two props but have them be very slightly different. Instead of `showInputIcon` and `customInputIcon`, I would have `showDefaultInputIcon` and `customInputIcon`. If you have either of the two props, the icon will show. The only difference will be whether it is the default option or the custom option. 

For defensive purposes, if both props are there, the custom icon should take precedence.

Does that seem good?

AntiFish03:
@majapw, I made the changes you suggested.

ljharb:
@AntiFish03 would you mind rebasing on top of latest master on the command line, so that there's no merge commits? If you have trouble, please let us know and we can take care of it for you (but please don't close this PR, delete the branch, or open a new one)

AntiFish03:
@ljharb, I would but I am away from a consistent internet connection at the moment.

ljharb:
@AntiFish03 np, i've gone ahead and rebased it for you.

AntiFish03:
@ljharb thanks

src/components/DateRangePickerInput.jsx
```diff
@@ -118,15 +123,27 @@ export default class DateRangePickerInput extends React.Component {
...
 
+    const inputIcon = customInputIcon || (<CalendarIcon />);
+
     return (
       <div
         className={cx('DateRangePickerInput', {
           'DateRangePickerInput--disabled': disabled,
         })}
       >
+        {(showDefaultInputIcon || customInputIcon !== null) &&
```
majapw: nit this could just be `(showDefaultInputIcon || customInputIcon)`

Issue #224, opened 2016-12-19
Pointer events get passed to previous (invisible) month 


Gnito:
Pointer events should be disabled when using enableOutsideDays (tested with single month DayPicker).
This is a problem if previous month has more week rows than current month. 

This could be fixed with simply adding ```pointer-events: none;```

CalendarMonth.scss:
```css
.CalendarMonth {
  text-align: center;
  padding: 0 13px;
  vertical-align: top;

  &:first-of-type {
    position: absolute;
    z-index: -1;
    opacity: 0;
    pointer-events: none;
  }
```

majapw:
This seems reasonable to me. Also I just reproed and agree, it's weird and we should probably fix that. 

One thing is that maybe we should set `pointer-events` back to `auto` for `.CalendarMonth--animating` so that one can still interact with the calendar during the transition. Other than that I support this change.

majapw:
Should be fixed in v4.3.0! Thanks!

PR #227, opened 2016-12-20
Remove pointer events from previous (invisible) month

Gnito:
This is a problem if previous month has more week rows than the current month.


Gnito:
https://github.com/airbnb/react-dates/issues/224

majapw:
LGTM! Thanks!

PR #229, opened 2016-12-21
Set input value to use ISO format by default

majapw:
If we don't add a layer of abstraction to format the date client side, the input value will be submitted directly to the server. As a result, I think we should not be using the display format but rather a standardized ISO-8601 format. 

to: @jeffsee55 @airbnb/webinfra

### DIFF
--- a/src/components/DateInput.jsx
+++ b/src/components/DateInput.jsx
@@ -6,7 +6,8 @@ import isTouchDevice from '../utils/isTouchDevice';
 const propTypes = {
   id: PropTypes.string.isRequired,
   placeholder: PropTypes.string, // also used as label
-  dateValue: PropTypes.string,
+  displayValue: PropTypes.string,
+  inputValue: PropTypes.string,
   focused: PropTypes.bool,
   disabled: PropTypes.bool,
   required: PropTypes.bool,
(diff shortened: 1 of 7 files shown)


majapw:
I just realized this won't actually work for custom display formats.

majapw:
@ljharb let's try again. Lol. This is messier, but also less broken. PTAL?

src/components/DateInput.jsx
```diff
@@ -88,15 +90,17 @@ export default class DateInput extends React.Component {
...
+      inputValue,
       focused,
       showCaret,
       onFocus,
       disabled,
       required,
     } = this.props;
 
-    const value = dateValue || dateString;
+    const displayText = displayValue || inputValue || dateString || placeholder;
```
ljharb: does this also need an `|| ''` at the end?

yo dawg, i heard you like defaults

src/components/DateRangePickerInput.jsx
```diff
@@ -23,7 +23,9 @@ const propTypes = {
   onClearDates: PropTypes.func,
 
   startDate: PropTypes.string,
+  startDateValue: PropTypes.string,
```
ljharb: these need defaultProps (the linter checks this now)

Issue #228, opened 2016-12-21
When giving a date from store to the component, the calander shows latest month


amitripshtos:
Hey,
There is an annoying bug, where I have SingleDatePicker which gets the date from my redux store.
Everything works well, until I leave the page (but the date is still in the store)

when I come back to the page the SingleDatePicker exists in, the date is saved well , but when I open the component I see December as month instead the date I chose.

Picture for example:
I choose a date from september
![image](https://cloud.githubusercontent.com/assets/10770124/21378759/479a35c4-c751-11e6-9772-ff710aa1b513.png)

I leave the page to another page, and then come back:
![image](https://cloud.githubusercontent.com/assets/10770124/21378787/7cced8f8-c751-11e6-8c03-1d667fbb008b.png)

It should show month september instead december (which is the latest date)
Thanks for your help ! :D

majapw:
Are you using `initialVisibleMonth` to set the month or is it like, you open the calendar, navigate to a month, select a date and then go to another window and come back? Can you share your set-up of the `SingleDatePicker`?

amitripshtos:
I'm doing the second option by setting the value only.

I think the component should know which month to show by the value, nop? It makes sense

majapw:
Well, mostly I am concerned about the fact that when you leave the page and return, it doesn't seem to save the visible month... which is not something that should happen unless the `DayPicker` is unmounted. Again, can you share your set-up?

While I think that it maybe makes sense to open it to where the specified date is, I feel like this is not the case in all situations. I think for instance, saving where the user navigated to makes sense in some scenarios. Right now, we have the `initialVisibleMonth` prop which controls what month is visible but only the first time the datepicker is open. Thereafter, it should rely on the saved state of where was navigated to. Basically, unless you've refreshed the page, it should stay open to the same month you were at.

amitripshtos:
I do unmount the component, but the value is saved in my redux store. So when I remount the component, I want the month to be set. I will use the initial visible month to fix that.

Thanks!

majapw:
Sweet! Definitely, I think it is worth considering setting the default value of `initialVisibleMonth` to the date if one is passed so maybe that's a change worth making. :)

amitripshtos:
That would be awesome.. I will try to create a pr

slecorvaisier:
@majapw is there a way to change the displayed month while keeping the datepicker open ?
[...]

PR #230, opened 2016-12-22
Now, SingleDatePicker will use 'date' value as initialVisibleMonth if…

amitripshtos:
Fixing #228

### DIFF
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -48,7 +48,7 @@ const defaultProps = {
   required: false,
   reopenPickerOnClearDates: false,
   keepOpenOnDateSelect: false,
-  initialVisibleMonth: () => moment(),
+  initialVisibleMonth: null,
   navPrev: null,
   navNext: null,
 
@@ -207,19 +207,21 @@ export default class DateRangePicker extends React.Component {
       withPortal,
       withFullScreenPortal,
       enableOutsideDays,
-      initialVisibleMonth,
       focusedInput,
       startDate,
       endDate,
       minimumNights,
       keepOpenOnDateSelect,
       renderDay,
+      initialVisibleMonth,
     } = this.props;
     const { dayPickerContainerStyles } = this.state;
 
     const onOutsideClick = (!withFullScreenPortal && withPortal)
       ? this.onOutsideClick
       : undefined;
+    const initialVisibleMonthThunk =
+      initialVisibleMonth || (() => (startDate || endDate || moment()));
 
     return (
       <div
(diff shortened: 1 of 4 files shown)


src/components/SingleDatePicker.jsx
```diff
@@ -277,6 +277,7 @@ export default class SingleDatePicker extends React.Component {
     };
 
     const onOutsideClick = !withFullScreenPortal ? this.onClearFocus : undefined;
+    const onInitialVisibleMonth = initialVisibleMonth ? initialVisibleMonth : () => (date || moment())
```
majapw: I think it would be better to remove `initialVisibleMonth` from the `this.props` deconstruction and then change this line to:
```
const initialVisibleMonth = this.props.initialVisibleMonth || () => (date || moment());
```
`onInitialVisibleMonth` doesn't make very much sense to me as a variable name.

You're also missing a semicolon which is likely causing a lint error.

majapw:
@ljharb could you take a look?

@milesj this should resolve the issue you were encountering!

src/components/DateRangePicker.jsx
```diff
@@ -214,6 +213,8 @@ export default class DateRangePicker extends React.Component {
     const onOutsideClick = (!withFullScreenPortal && withPortal)
       ? this.onOutsideClick
       : undefined;
+    const initialVisibleMonth =
```
ljharb: i'd still prefer to destructure this out of props, and then use a different variable name here.

src/components/SingleDatePicker.jsx
```diff
@@ -299,9 +299,8 @@ export default class SingleDatePicker extends React.Component {
       selected: day => this.isSelected(day),
     };
 
-    const onOutsideClick = (!withFullScreenPortal && withPortal)
-      ? this.onClearFocus
-      : undefined;
+    const onOutsideClick = (!withFullScreenPortal && withPortal) ? this.onClearFocus : undefined;
+    const initialVisibleMonth = this.props.initialVisibleMonth || (() => (date || moment()));
```

src/components/DateRangePicker.jsx
majapw: ideas for a good name?

ljharb: `initialVisibleMonthThunk`?

amitripshtos:
Hey guys , sorry of being unactive (a lot of work these days)
I saw that you guys fixed the PR issues, is everything is over? Do you want me to pull from master before merging?

ljharb:
@amitripshtos if you'd mind fixing the remaining comments, and *rebasing* (not merging) master, that'd be appreciated

majapw:
@ljharb addressed your comments and rebased. Should be good to go now!

PR #244, opened 2017-01-06
Set vertical day picker height using number of months

moonboots:
to: @majapw @ljharb 

For vertical DayPickers, calculate the height with `numberOfMonths` instead of using a fixed value, which currently truncate halfway into the second month. This allows vertically scrollable DayPickers on mobile by increasing the month count.

### DIFF
--- a/css/DayPicker.scss
+++ b/css/DayPicker.scss
@@ -57,6 +57,56 @@
   left: 50%;
 }
 
+.DayPicker.DayPicker--scrollable {
+  height: 100%;
+
+  .DayPicker__week-header {
+    top: 0;
+    display: table-row;
+    border-bottom: 1px solid $react-dates-color-border;
+    background: white;
+  }
+
+  .CalendarMonth__caption {
+    padding-bottom: 5px;
+    padding-top: 5px;
+  }
+
+  .transition-container--vertical {
+    padding-top: 20px;
+    height: 100%;
+    position: absolute;
+    top: 0;
+    bottom: 0;
+    right: 0;
+    left: 0;
+    overflow-y: scroll;
+  }
+
+  .DayPickerNavigation__prev {
+    display: none;
+  }
+
+  .DayPickerNavigation__next {
+    width: 100%;
+  }
+
+  .DayPickerNavigation--vertical {
+    position: relative;
+  }
+
+  .CalendarMonthGrid--vertical {
+    overflow-y: scroll;
+  }
+
+  &.DayPicker--vertical .DayPicker__week-header {
+    margin-left: 0;
+    left: 0;
+    width: 100%;
+    text-align: center;
+  }
+}
+
 .transition-container {
   position: relative;
   overflow: hidden;
--- a/src/components/DayPicker.jsx
+++ b/src/components/DayPicker.jsx
@@ -28,6 +28,7 @@ const propTypes = {
   withPortal: PropTypes.bool,
   hidden: PropTypes.bool,
   initialVisibleMonth: PropTypes.func,
+  scrollable: PropTypes.bool,
 
   navPrev: PropTypes.node,
   navNext: PropTypes.node,
@@ -55,6 +56,7 @@ const defaultProps = {
   orientation: HORIZONTAL_ORIENTATION,
   withPortal: false,
   hidden: false,
+  scrollable: false,
 
   initialVisibleMonth: () => moment(),
 
@@ -142,10 +144,12 @@ export default class DayPicker extends React.Component {
       currentMonth: props.hidden ? moment() : props.initialVisibleMonth(),
       monthTransition: null,
       translationValue: 0,
+      scrollableMonthMultiple: 1,
     };
 
     this.onPrevMonthClick = this.onPrevMonthClick.bind(this);
     this.onNextMonthClick = this.onNextMonthClick.bind(this);
+    this.addScrollableMonths = this.addScrollableMonths.bind(this);
     this.updateStateAfterMonthTransition = this.updateStateAfterMonthTransition.bind(this);
   }
 
@@ -220,6 +224,17 @@ export default class DayPicker extends React.Component {
     this.setState({
       monthTransition: NEXT_TRANSITION,
       translationValue,
+    });
+  }
+
+  addScrollableMonths(e) {
+    if (e) e.preventDefault();
+    if (this.props.onNextMonthClick) {
+      this.props.onNextMonthClick(e);
+    }
+
+    this.setState({
+      scrollableMonthMultiple: this.state.scrollableMonthMultiple + 1,
     });
   }
 
@@ -314,6 +329,20 @@ export default class DayPicker extends React.Component {
     );
   }
 
+  renderAddScrollableMonths() {
+    const {
+      navNext,
+    } = this.props;
+
+    return (
+      <DayPickerNavigation
+        onNextMonthClick={this.addScrollableMonths}
+        navNext={navNext}
+        isVertical={this.isVertical()}
+      />
+    );
+  }
+
   renderWeekHeader(index) {
     const horizontalStyle = {
       left: index * CALENDAR_MONTH_WIDTH,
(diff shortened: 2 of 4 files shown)


src/components/DayPicker.jsx
```diff
@@ -279,10 +280,11 @@ export default class DayPicker extends React.Component {
       }
     });
 
+    const monthHeight = Math.max(...heights);
+    this.setState({ monthHeight });
     const newMonthHeight = Math.max(...heights) + MONTH_PADDING;
```
ljharb: let's not compute `Math.max` twice - please reuse `monthHeight` here.

majapw:
@moonboots Have you looked at what this does for the attached vertical `DayPicker`?

moonboots:
@majapw For the attached vertical daypicker, the months expands in height but still stick in the same position. I'm working on a new `scrollable` prop to gate the vertical height changes. The same prop could also enable related changes like a shared, sticky week header and a button to add more months (instead of vertically paging).

@ljharb Fixed the double `Math.max` call, added a story, and working on a spec to check the computed height

moonboots:
@majapw Here's a gif of the scrollable vertical datepicker I'm picturing with a shared week header, 12 months of scroll, and a single next button that will append months instead of paging:

scrollable-datepicker

I'm cleaning up the code and adding specs, but please let me know if you see any interaction issues.

majapw:
Is that behavior totally replacing the vertical attached DayPicker or is the `scrollable` prop you mentioned?

How's the perf looking?

moonboots:
These behaviors would all be enabled with the `scrollable` prop and wouldn't be used when attached. The scrolling performance is smooth, and I'll test the day selection on some phones to make sure there are no issues.

moonboots:
@majapw I've pushed the `<DayPicker scrollable />` changes, PTAL. Here's a [link to a storybook demo](https://moonboots.github.io/react-dates/?selectedKind=DayPicker&selectedStory=vertically%20scrollable%20with%2012%20months&full=0&down=1&left=1&panelRight=0&downPanel=kadirahq%2Fstorybook-addon-actions%2Factions-panel). My goal is to preserve the existing vertical pagination behavior unless the `scrollable` prop is set. Here's a summary of the changes; please let me know if anything doesn't make sense or could be cleaned up:

* The vertical months that would currently be hidden by the `1.75 * CALENDAR_MONTH_WIDTH` height are revealed with a [css `overflow-y: scroll;`](https://github.com/airbnb/react-dates/pull/244/files#diff-8ff9be8e09eae49f1c2327060c1acab1R83).
* I repurposed the [navigation next button](https://github.com/airbnb/react-dates/pull/244/files#diff-883056b4b11263241c339f927e93411bR338) to [increase the number of visible months](https://github.com/airbnb/react-dates/pull/244/files#diff-883056b4b11263241c339f927e93411bR237) instead of paginating.

moonboots:
Eventually I want to wrap this `<DateRangePickerInputController />` and `<DayPickerRangeController scrollable />` combination into a reusable component:

screen shot 2017-01-10 at 7 07 54 pm

css/DayPicker.scss
```diff
@@ -57,6 +57,56 @@
   left: 50%;
 }
 
+.DayPicker.DayPicker--scrollable {
```
majapw: why both selectors? I think we can just use `.DayPicker--scrollable`

css/DayPicker.scss
```diff
@@ -57,6 +57,56 @@
...
+  height: 100%;
+
+  .DayPicker__week-header {
+    top: 0;
+    display: table-row;
+    border-bottom: 1px solid $react-dates-color-border;
+    background: white;
+  }
+
+  .CalendarMonth__caption {
```
majapw: So ideally, we don't want any `CalendarMonth` stylings to be in the `DayPicker.scss` file. Having that confusion is going to make converting to `react-with-styles` more of a pain. The solution is to pass down the scrollable prop as far as it needs to go down the tree and add a `CalendarMonth--scrollable` class and a `DayPickerNavigation--scrollable` class and so forth.

css/DayPicker.scss
```diff
@@ -57,6 +57,56 @@
...
+    height: 100%;
+    position: absolute;
+    top: 0;
+    bottom: 0;
+    right: 0;
+    left: 0;
+    overflow-y: scroll;
+  }
+
+  .DayPickerNavigation__prev {
```
majapw: Same comment about the `DayPickerNavigation` namespace here

css/DayPicker.scss
```diff
@@ -57,6 +57,56 @@
...
+
+  .DayPickerNavigation__next {
+    width: 100%;
+  }
+
+  .DayPickerNavigation--vertical {
+    position: relative;
+  }
+
+  .CalendarMonthGrid--vertical {
```
majapw: And the `CalendarMonthGrid` namespace here

src/components/DayPicker.jsx
```diff
@@ -142,10 +144,12 @@ export default class DayPicker extends React.Component {
       currentMonth: props.hidden ? moment() : props.initialVisibleMonth(),
       monthTransition: null,
       translationValue: 0,
+      scrollableMonthMultiple: 1,
```
majapw: What does this variable mean?

src/components/DayPicker.jsx
```diff
@@ -223,6 +227,17 @@ export default class DayPicker extends React.Component {
     });
   }
 
+  addScrollableMonths(e) {
```
majapw: nit: should be `addScrollableMonth` since it only adds one at a time

src/components/DayPicker.jsx
```diff
@@ -28,6 +28,7 @@ const propTypes = {
   withPortal: PropTypes.bool,
   hidden: PropTypes.bool,
   initialVisibleMonth: PropTypes.func,
+  scrollable: PropTypes.bool,
```
majapw: So this prop is just going to be a clusterfuck if we happen to have the orientation set to horizontal, right? `airbnb-prop-types` has some shit about exclusive boolean props, but that's not quite what's happening here. Could we create a rule for this?

src/components/DayPickerRangeController.jsx
```diff
@@ -41,6 +41,7 @@ const propTypes = {
   withPortal: PropTypes.bool,
   hidden: PropTypes.bool,
   initialVisibleMonth: PropTypes.func,
+  scrollable: PropTypes.bool,
```
majapw: We don't want to be able to do this from the `DateRangePicker` level?

majapw:
Also, we're still missing tests.

moonboots:
Thanks @majapw, I'll address comments in this PR and open a new PR because the scope has changed to include the vertical scroll.

src/components/DayPicker.jsx
moonboots: To enforce the vertical requirement, I can add a new orientation instead of using this additional prop.

src/components/DayPicker.jsx
moonboots: I'm adding new months in batches instead of one by one, and this variable tracks how many batches have been added. For example, if there are initially 12 scrollable months and the user clicks the next button, I'm adding another 12 months by incrementing this variable.

src/components/DayPicker.jsx
majapw: Not a bad idea. 👍

src/components/DayPickerRangeController.jsx
moonboots: The current `DateRangePicker`'s popup calendar doesn't seamlessly fit the new scrolling calendars, and my goal is to create a new component like `ScrollableDateRangePicker` that combines a fixed `DateInput` header with the scrolling region below.

majapw: This might make the orientation idea a bit sketchy since the DRP and SDP take an orientation prop

Issue #203, opened 2016-12-01
Q/A: Adjusting the size of date range picker ?


cemremengu:
Is it possible to make date range picker smaller than the default size or make it fluid without fiddling with the scss ?

majapw:
Hi @cemremengu, what specifically are you trying to adjust? Like the individual `CalendarDay` squares?

majapw:
This exists now!

cemremengu:
@majapw I was watching from shadows due to my work load. :eyes:  Thank you for your hard work!

PR #249, opened 2017-01-12
Adds daySize prop to the DRP/SDP/DP that scales everything accordingly

majapw:
Basically, this conflates all of the code associated with the width of the `DayPicker` into the JS side of things (which again, will be nice for using `react-with-styles`). The DRP/SDP now take a `daySize` prop (specifically a numerical pixel value) and then does all of its calculations based on that. 

This should be a fix for https://github.com/airbnb/react-dates/issues/203

~~I'm going to add one more commit that modifies the height calculations to use this number as well instead of reaching into the DOM. ~~
Just kidding! The fact that the month title size is still not available to me is a problem. We'll just have to wait til everything is in one place with `react-with-styles`

to: @airbnb/webinfra @moonboots @ljharb 
fyi: @cemremengu

### DIFF
--- a/src/components/CalendarMonth.jsx
+++ b/src/components/CalendarMonth.jsx
@@ -24,6 +24,7 @@ const propTypes = {
   enableOutsideDays: PropTypes.bool,
   modifiers: PropTypes.object,
   orientation: ScrollableOrientationShape,
+  daySize: PropTypes.number,
   onDayClick: PropTypes.func,
   onDayMouseEnter: PropTypes.func,
   onDayMouseLeave: PropTypes.func,
@@ -39,6 +40,7 @@ const defaultProps = {
   enableOutsideDays: false,
   modifiers: {},
   orientation: HORIZONTAL_ORIENTATION,
+  daySize: 39,
   onDayClick() {},
   onDayMouseEnter() {},
   onDayMouseLeave() {},
@@ -80,6 +82,7 @@ export default class CalendarMonth extends React.Component {
       onDayMouseEnter,
       onDayMouseLeave,
       renderDay,
+      daySize,
     } = this.props;
 
     const { weeks } = this.state;
@@ -104,6 +107,7 @@ export default class CalendarMonth extends React.Component {
                 {week.map((day, dayOfWeek) => (
                   <CalendarDay
                     day={day}
+                    daySize={daySize}
                     isOutsideDay={!day || day.month() !== month.month()}
                     modifiers={modifiers}
                     key={dayOfWeek}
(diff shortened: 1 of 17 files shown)


src/components/DateRangePicker.jsx
```diff
@@ -199,10 +202,6 @@ export default class DateRangePicker extends React.Component {
           orientation={orientation}
           enableOutsideDays={enableOutsideDays}
           numberOfMonths={numberOfMonths}
-          onDayMouseEnter={this.onDayMouseEnter}
-          onDayMouseLeave={this.onDayMouseLeave}
-          onDayMouseDown={this.onDayClick}
-          onDayTouchTap={this.onDayClick}
```
moonboots: Are these props unused?

src/utils/getCalendarMonthWidth.js
```diff
@@ -0,0 +1,5 @@
+const CALENDAR_MONTH_PADDING = 9;
```
moonboots: Can we make this a property as well?

src/components/DateRangePicker.jsx
majapw: yeah. this was a mistake.

src/utils/getCalendarMonthWidth.js
majapw: probs. the `DAY_PICKER_PADDING is also separate. This is the padding on each month (and between months as well).

src/components/CalendarMonth.jsx
```diff
@@ -36,6 +37,7 @@ const defaultProps = {
   enableOutsideDays: false,
   modifiers: {},
   orientation: HORIZONTAL_ORIENTATION,
+  daySize: 39,
```
ljharb: why 39?

src/components/CalendarMonthGrid.jsx
```diff
@@ -55,6 +57,7 @@ const defaultProps = {
   onDayTouchTap() {},
   onMonthTransitionEnd() {},
   transform: 'none',
+  daySize: 39,
```
ljharb: if `39` is important, could it be a reused constant, instead of being hardcoded in multiple places?

src/components/CalendarMonthGrid.jsx
```diff
@@ -32,6 +33,7 @@ const propTypes = {
   onDayTouchTap: PropTypes.func,
   onMonthTransitionEnd: PropTypes.func,
   transformValue: PropTypes.string,
+  daySize: PropTypes.number,
```
ljharb: should this perhaps be a finite, positive integer? If so, `airbnb-prop-types` has `nonNegativeInteger` for just this use case :-D

zachfeldman:
Heeeeey @majapw how's this going? I was considering trying to break it off myself in the next few days as it would be very helpful for my inline picker being mobile friendly!

majapw:
Closing in favor of #406

PR #258, opened 2017-01-20
Fix issue when parent element of component has 'text-align:center',…

pebf:
The calendar layout breaks when the DatePicker's parent element is 'text-align:center'. I fixed the problem.


ljharb:
Duplicate of #257.

ljharb:
Actually my mistake, this is related but affects a different file.

PR #266, opened 2017-01-26
Add aria-describedby attribute to date input

### LINKED ISSUE #261: Accessibility improvement
Hi,

I'm using the date range picker and I would like to include an additional message for screen reader users to inform them of the constraints around what they can type in the inputs.

Would you be open to me submitting a pull request that adds an `aria-describedby` to the DateInput ?

This way you could write an additional message to users, to let them know about minimum nights, the date format they should type in and blocked out dates etc.

allanpope:
When using a screen reader you are able to type in a date but you don't know about any restrictions the date component might have.

This pull request adds the ability to inform screen reader users about the constraints of the component they are using via the `aria-describedby` attribute.

Any text passed into the new `screenReaderInputMessage` prop will be read out when a screen reader user interacts with the input.

Closes #261

### DIFF
--- a/README.md
+++ b/README.md
@@ -199,6 +199,11 @@ If the `required` prop is set to true, the input will have to be filled before t
   required: PropTypes.bool,
 ```
 
+The `screenReaderInputMessage` prop accepts a contextual message for screen readers. When an input is focused, the `screenReaderInputMessage` prop value is read. This can inform users about constraints, such as the date format, minimum nights, blocked out dates, etc.
+```
+  screenReaderInputMessage: PropTypes.string,
+```
+
 **Custom Navigation Icons:**
 
 The `navPrev` and `navNext` props are used to assign custom icons to the "Next", and "Previous" arrows for the top navigation. If you do specify a custom icon you'll have to do the styling of the button yourself (e.g. the color, border, and things of that nature). All that comes "out of the box" is your button being positioned correctly.
--- a/src/components/DateInput.jsx
+++ b/src/components/DateInput.jsx
@@ -8,6 +8,7 @@ const propTypes = {
   placeholder: PropTypes.string, // also used as label
   displayValue: PropTypes.string,
   inputValue: PropTypes.string,
+  screenReaderMessage: PropTypes.string,
   focused: PropTypes.bool,
   disabled: PropTypes.bool,
   required: PropTypes.bool,
@@ -23,6 +24,7 @@ const defaultProps = {
   placeholder: 'Select Date',
   displayValue: '',
   inputValue: '',
+  screenReaderMessage: '',
   focused: false,
   disabled: false,
   required: false,
@@ -92,6 +94,7 @@ export default class DateInput extends React.Component {
       placeholder,
       displayValue,
       inputValue,
+      screenReaderMessage,
       focused,
       showCaret,
       onFocus,
@@ -101,6 +104,7 @@ export default class DateInput extends React.Component {
 
     const displayText = displayValue || inputValue || dateString || placeholder || '';
     const value = inputValue || displayValue || dateString || '';
+    const screenReaderMessageId = `DateInput__screen-reader-message-${id}`;
 
     return (
       <div
@@ -125,7 +129,14 @@ export default class DateInput extends React.Component {
           disabled={disabled}
           readOnly={this.isTouchDevice}
           required={required}
+          aria-describedby={screenReaderMessage && screenReaderMessageId}
         />
+
+        {screenReaderMessage &&
+          <p id={screenReaderMessageId} className="screen-reader-only">
+            {screenReaderMessage}
+          </p>
+        }
 
         <div
           className={cx('DateInput__display-text', {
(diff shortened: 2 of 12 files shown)


majapw:
@backwardok or @wyattdanger would you mind taking a look at this PR as our resident a11y experts?

test/components/DateInput_spec.jsx
```diff
@@ -58,6 +58,46 @@ describe('DateInput', () => {
       });
     });
 
+    describe('screen reader message', () => {
```
wyattdanger: Looks like we'll need similar specs for the other components touched in the PR

README.md
```diff
@@ -194,6 +194,11 @@ If the `required` prop is set to true, the input will have to be filled before t
   required: PropTypes.bool,
 ```
 
+The `screenReaderInputMessage` prop allows you to write a message for screen readers. When the prop is added a `aria-describedby` attribute will be added to the input box and the value will be read by screen readers. Take advantage of this to inform users about constraints your component has, such as the date format, minimum nights and blocked out dates etc.
```
wyattdanger: Would like to see a few grammatical changes here, but overall looks like reasonable documentation! Thanks for including this. Maybe something a bit more in this direction?

> The `screenReaderInputMessage` prop accepts a contextual message for screen readers. When an input is focused, the `screenReaderInputMessage` prop value is read. This can inform users about constraints, such as the date format, minimum nights, blocked out dates, etc.

src/components/DateInput.jsx
```diff
@@ -128,8 +131,15 @@ export default class DateInput extends React.Component {
           disabled={disabled}
           readOnly={this.isTouchDevice}
           required={required}
+          aria-describedby={screenReaderMessage && `DateInput__screen-reader-message-${id}`}
```
wyattdanger: Would prefer to see this string interpolated only once per render, perhaps at the top of the render method

README.md
allanpope: No problem. Thanks that reads a lot better.

src/components/DateInput.jsx
allanpope: Ah good pickup, I didn't consider that.

test/components/DateInput_spec.jsx
allanpope: Just had another look at the tests. I'm unsure what other components I could test?

All the tests files use enzymes shallow rendering, so I can only test that level of the component. The changes I've added to them are pretty simple. All they do is take a screenReaderMessage prop and pass it to a child component. This prop is passed all the way down until it gets to the `DateInput` component, which the above tests cover.

Happy to add more tests but need some further instruction, thanks.

src/components/SingleDatePicker.jsx
```diff
@@ -333,6 +333,7 @@ export default class SingleDatePicker extends React.Component {
       phrases,
       withPortal,
       withFullScreenPortal,
+      screenReaderInputMessage,
```
ljharb: Please add this prop to propTypes/defaultProps

stories/DateRangePicker.js
```diff
@@ -258,4 +258,9 @@ storiesOf('DateRangePicker', module)
     <DateRangePickerWrapper
       customInputIcon={<TestCustomInputIcon />}
     />
+  ))
+  .addWithInfo('with screen reader message', () => (
+    <DateRangePickerWrapper
+      screenReaderInputMessage='Here you could inform screen reader users of the date format, minimum nights, blocked out dates etc'
```
ljharb: Please use an Oxford comma after "dates" :-)

stories/SingleDatePicker.js
```diff
@@ -136,4 +136,9 @@ storiesOf('SingleDatePicker', module)
       numberOfMonths={1}
       enableOutsideDays
     />
+  ))
+  .addWithInfo('with screen reader message', () => (
+    <SingleDatePickerWrapper
+      screenReaderInputMessage='Here you could inform screen reader users of the date format, minimum nights, blocked out dates etc'
```
ljharb: Same here

test/components/DateInput_spec.jsx
```diff
@@ -69,6 +69,47 @@ describe('DateInput', () => {
       );
     });
 
+    describe('screen reader message', () => {
+      const id = 'date';
+      const screenReaderMessage = 'My screen reader message';
+      const screenReaderMessageName = `DateInput__screen-reader-message-${id}`;
+      const screenReaderMessageId = `#${screenReaderMessageName}`;
```
ljharb: Technically the ID is line 75, and this is the "selector" - could we improve the var names a bit?

src/components/SingleDatePicker.jsx
allanpope: Completely missed this one, have added it for this and DateRangePicker.

test/components/DateInput_spec.jsx
allanpope: Yeah your correct, I've updated the names of these variables. 

I also moved the creation of the wrappers into a before loop since they were all the same, which reduces the duplication and makes it easier to read.

stories/DateRangePicker.js
allanpope: Added.

wyattdanger:
Great! Please rebase, squash, write a solid commit message, and then we'll be good to ship. Thank you Allan

allanpope:
@wyattdanger @ljharb Thanks for taking a look at it.  

I've just rebased, squashed & committed.

PR #275, opened 2017-01-31
Remove mouse{down,up} and touch{start,end,tap}

moonboots:
to: @majapw 

Replaces mouse and touch handlers in favor of `onClick`/`onDayClick`. I think these mouse/touch handlers support functionality that's no longer used. This change should improve scroll performance on mobile because the browser doesn't need to interrupt scrolling to handle touchstart/end.

### DIFF
--- a/README.md
+++ b/README.md
@@ -457,18 +457,13 @@ Then, every Friday in the visible calendar would have the class `CalendarMonth__
 
 **Day interaction callbacks:**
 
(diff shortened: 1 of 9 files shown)


majapw:
Oh! I didn't realize you wanted to remove this completely. I suppose have fewer handlers floating around might help with performance as well. This seems reasonable to me (but is def breaking).

PR #279, opened 2017-01-31
Fix duplicate months

moonboots:
to: @majapw @ljharb 

This PR fixes two issues related to duplicate months in `CalendarMonthGrid`:

* When increasing `numberOfMonths`, the last month was [duplicated in the first iteration of this loop](https://github.com/airbnb/react-dates/compare/master...moonboots:fix-month-conflict?expand=1#diff-ed0d63b6cfc2e8bdb08f69f4867c47a8L117). I simplified to reuse the `getMonths` logic instead of updating piecemeal. This should only be called when `numberOfMonths` changes, and it shouldn't affect the original perf improvements.
* Include year in `CalendarMonth` key to allow more than 12 months. Otherwise, React will only render the first occurrence and print `Warning: flattenChildren(...): Encountered two children with the same key`.

### DIFF
--- a/src/components/CalendarMonthGrid.jsx
+++ b/src/components/CalendarMonthGrid.jsx
@@ -111,14 +111,7 @@ export default class CalendarMonthGrid extends React.Component {
     }
 
     if (hasNumberOfMonthsChanged) {
-      if (this.props.numberOfMonths < numberOfMonths) {
-        newMonths = months.slice();
-        for (let i = 0; i < numberOfMonths - this.props.numberOfMonths; i += 1) {
-          newMonths.push(months[months.length - 1].clone().add(i, 'month'));
-        }
-      } else {
-        newMonths = months.slice(0, numberOfMonths + 3);
-      }
+      newMonths = getMonths(initialMonth, numberOfMonths);
     }
 
     this.setState({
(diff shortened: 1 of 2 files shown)


test/components/CalendarMonthGrid_spec.jsx
```diff
@@ -51,4 +52,24 @@ describe('CalendarMonthGrid', () => {
...
+      initialMonth,
+      numberOfMonths: 24,
+    });
+
+    const months = wrapper.state().months;
+
+    const collisions = months
+      .map(m => m.format('YYYY-MM'))
+      .reduce((acc, m) => Object.assign(
+        {}, acc, { [m]: true }
```
ljharb: if the invocation parens and every argument can't fit on one line, then each argument must be on a line by itself.

src/components/CalendarMonthGrid.jsx
```diff
@@ -111,14 +111,7 @@ export default class CalendarMonthGrid extends React.Component {
...
     if (hasNumberOfMonthsChanged) {
-      if (this.props.numberOfMonths < numberOfMonths) {
-        newMonths = months.slice();
-        for (let i = 0; i < numberOfMonths - this.props.numberOfMonths; i += 1) {
-          newMonths.push(months[months.length - 1].clone().add(i, 'month'));
-        }
-      } else {
-        newMonths = months.slice(0, numberOfMonths + 3);
-      }
+      newMonths = getMonths(initialMonth, numberOfMonths);
```
majapw: Technically, I think the reason for this was to not create hella new moment objects when you changed the number of months which I think getMonths does, but I suppose that's a fairly rare edgecase.

moonboots:
Thanks, fixed the lint and spec issues, ptal

PR #280, opened 2017-02-01
Use `aria-label` instead of visually hidden label

wyattdanger:
Instead of using a `label` element with CSS to visually hide it, use the
`aria-label` property on the `input` element itself. Having hidden text for
screen readers could cause the label value to be read aloud twice by the screen
reader: first when navigating nodes in the form, and then again when focusing
the input.

### DIFF
--- a/src/components/DateInput.jsx
+++ b/src/components/DateInput.jsx
@@ -109,11 +109,8 @@ export default class DateInput extends React.Component {
           'DateInput--disabled': disabled,
         })}
       >
-        <label className="DateInput__label" htmlFor={id}>
-          {placeholder}
-        </label>
-
         <input
+          aria-label={placeholder}
           className="DateInput__input"
           type="text"
           id={id}
--- a/css/DateInput.scss
+++ b/css/DateInput.scss
@@ -42,17 +42,6 @@ $caret-top: $react-dates-spacing-vertical-picker - $react-dates-width-tooltip-ar
   background: $react-dates-color-gray-lighter;
 }
 
-.DateInput__label {
-  border: 0;
-  clip: rect(0, 0, 0, 0);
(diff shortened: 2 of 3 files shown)


src/components/DateInput.jsx
```diff
@@ -109,11 +109,8 @@ export default class DateInput extends React.Component {
           'DateInput--disabled': disabled,
         })}
       >
-        <label className="DateInput__label" htmlFor={id}>
-          {placeholder}
-        </label>
-
         <input
+          aria-label={placeholder}
```
ljharb: note that this prop defaults to the empty string, so i think you might want `{placeholder || null}` here?

ljharb:
Technically I think this is a breaking change, because somebody might have been using the CSS class to style it.

src/components/DateInput.jsx
lencioni: `placeholder` defaults to `'Select Date'`

ljharb: whoops, my eye looked at the wrong line

wyattdanger:
@ljharb I wouldn't expect anybody to use `.DateInput__label` but I suppose better safe than sorry?

majapw:
This looks reasonable to me. I agree with @ljharb that there is definitely a possibility that someone was overriding the class to create a visible label.

.... In other news, can we just make a visible label and ship that with an option to hide it for our own designs when we need to? I think that would be nice.

backwardok:
@majapw or even better, we make a visible label and you can't hide it and it just always shows up for everyone!

majapw:
I suppose the design on airbnb.com now has an actual visible label so we could ostensibly sneak that in. I think that mobile designs will not be so friendly though.

Issue #127, opened 2016-10-16
Make date cell render configurable


evgenTraytyak:
<img width="301" alt="screen shot 2016-10-16 at 17 13 40" src="https://cloud.githubusercontent.com/assets/6050297/19418019/06154366-93c4-11e6-9e08-29937ea493f4.png">

I want to render some additional information under the date. Price for example.
Does anybody have any ideas how to do it?

senseluo:
voting for this feature

ljharb:
@senseluo a +1 reaction on the original post is always sufficient to "vote" for a feature; +1 comments and similar don't add any value. Since you've already done that, I'll delete your comment as well as my reply, so as to keep the thread clean. Thanks!

majapw:
This has been been fixed by #283 (I must have missed this issue) but you can pass in a function to the `renderDay` prop to accomplish this. The function will receive the moment object representing the day as an argument. 

Check out https://github.com/airbnb/react-dates/blob/master/stories/DateRangePicker_day.js#L72 for an example of how to use it.

sunlee-newyork:
I tried to return a React element but it only works by returning a string. Is there a way to return a React element instead?

PR #283, opened 2017-02-01
[Issue 127] Render arbitrary content in CalendarDay

evandavis:
Re: #127

This PR adds a `renderDetails` method which, if provided, inserts a `<span className="CalendarDay__details"/>` into the `CalendarDay` square. 

I'm not married to the naming or the implementation; any/all criticism is welcome.

```js
/**
 * @param day - the Moment object representing the day
 * @returns {string|element} - returns anything that is valid in a JSX node: string, React element, null
 */
 // example
 // print a happy cat on Fridays
 <CalendarDay
   renderDetails={day => day.day() % 6 === 5 && '😻'} />
```

screenshot 2017-02-01 17 36 44

### DIFF
--- a/src/components/CalendarDay.jsx
+++ b/src/components/CalendarDay.jsx
@@ -8,6 +8,7 @@ const propTypes = {
   onDayClick: PropTypes.func,
   onDayMouseEnter: PropTypes.func,
   onDayMouseLeave: PropTypes.func,
+  renderDetails: PropTypes.func,
 };
 
 const defaultProps = {
@@ -15,6 +16,7 @@ const defaultProps = {
   onDayClick() {},
   onDayMouseEnter() {},
   onDayMouseLeave() {},
+  renderDetails: null,
 };
 
 export default class CalendarDay extends React.Component {
@@ -38,7 +40,7 @@ export default class CalendarDay extends React.Component {
   }
 
   render() {
-    const { day } = this.props;
+    const { day, renderDetails } = this.props;
 
     return (
       <div
@@ -48,6 +50,9 @@ export default class CalendarDay extends React.Component {
         onClick={e => this.onDayClick(day, e)}
       >
         <span className="CalendarDay__day">{day.format('D')}</span>
+        {renderDetails && (
+          <span className="CalendarDay__details">{renderDetails(day)}</span>
+        )}
       </div>
     );
   }
--- a/src/components/CalendarDay.jsx
+++ b/src/components/CalendarDay.jsx
@@ -39,14 +39,9 @@ export default class CalendarDay extends React.Component {
     onDayMouseLeave(day, e);
   }
 
-  maybeRenderDetails(day) {
-    const { renderDetails } = this.props;
-    return renderDetails(day);
-  }
-
   render() {
-    const { day } = this.props;
-    const details = this.maybeRenderDetails(day);
+    const { day, renderDetails } = this.props;
+    const details = renderDetails(day);
     return (
       <div
         className="CalendarDay"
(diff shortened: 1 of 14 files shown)


evandavis:
It occurs to me that this is rendering the detail `span` even if there is no content; I'll update.

src/components/CalendarDay.jsx
```diff
@@ -48,6 +55,9 @@ export default class CalendarDay extends React.Component {
         onClick={e => this.onDayClick(day, e)}
       >
         <span className="CalendarDay__day">{day.format('D')}</span>
+        {details && (
+          <span className="CalendarDay__details">{details}</span>
```
majapw: You could probably simplify this a bit and just have this be 
```
<span className="CalendarDay__details">{renderDetails(day)}</span>
```
and deconstruct `renderDetails` from the props. I'm not sure that there's a need for the function or to separate out the variable.

evandavis: @majapw if you look through the individual commits, I added this in 0531054 so that the `details` span wouldn't render if it was empty; maybe that wasn't needed?

majapw: I see. Because details defaults to a noOp which is truthy. I guess we could just set `const details = renderDetails(day);` and get rid of the `maybeRenderDetails` function but leave this as is.

evandavis: Indeed. In re:handing off render entirely, the reason I didn't go that way is because if the user still wants the default formatting for the number, they would have to know to wrap in `.CalendarDay__day`; I didn't think the API should expose implementation.

Could you say more about `React.Children.only`? The documentation for that method doesn't really tell me _why_ I would use it. 😕

evandavis: Also, should I add a modifier class to the whole cell when details are present to make styling the content easier?

evandavis:
In light of 22e231c getting merged, it seems like the better option now is to hand over rendering to the prop. What do you think @majapw? Is this still viable?

majapw:
@evandavis I was actually thinking the same thing yesterday. :) Let's keep this as is and merge this in asap after conflicts have been resolved.

evandavis:
@majapw I actually have the other implementation ready to go; I think we should try to get that in instead. I just wasn't sure if I should update this branch or open a new PR.

https://github.com/airbnb/react-dates/compare/master...relayrides:issue-127_renderDay

zachfeldman:
@evandavis not trying to clog up this PR, but I tested out your branch because I badly need this feature with whatever datepicker we end up going with. I went with the code from https://github.com/airbnb/react-dates/blob/master/examples/DateRangePickerWrapper.jsx almost exactly and got the following error in the JS console when I went to click the "Start Date" input:

`Warning: Failed prop type: Invalid input type: `day` of type `boolean` supplied to `CalendarDay`, expected `object`.`

Just passing along in case it's helpful! Really appreciate both of your work on this amazing component.

majapw:
Hi @zachfeldman I think that issue is unrelated and should be fixed in https://github.com/airbnb/react-dates/pull/305

majapw:
@evandavis Probably a new PR is the best way to go! Two comments on the diff you posted 

- you'd need to update the README to say `renderDay` instead of `renderDetails`
- I think this is where we'd want to wrap https://github.com/airbnb/react-dates/compare/master...relayrides:issue-127_renderDay#diff-bfcddc5b7aece6519bd248744069628bR74 in `React.Children.only`. The primary reason for this is to enforce passing down one parent DOM node (it will throw if you try to pass in like, two sibling spans or something like that). This is to help enforce what `renderDay` returns in a way that we can't do with proptypes.  @ljharb can probably say a little more as well.

Thanks again for this PR!

evandavis:
@majapw thanks, I'll make those updates and open the new PR, linking to this one for history.

evandavis:
Closing in favor of #307

PR #286, opened 2017-02-01
Avoid rendering DayPicker when not visible

lencioni:
I was looking at some markup that we generate during server rendering
and noticed a whole bunch of markup (~32 KiB of HTML) belonging to
react-dates for elements that were not visible. With a couple of minor
modifications, we can avoid rendering most of these nodes until the
input is actually focused and we want to open the DayPicker.

---

To @airbnb/webinfra @majapw 

If you don't see any problems with this approach, I'll go ahead and update the tests.

### DIFF
--- a/css/DateRangePicker.scss
+++ b/css/DateRangePicker.scss
@@ -23,14 +23,6 @@
   top: $react-dates-spacing-vertical-picker;
 }
 
-.DateRangePicker__picker--show {
-  visibility: visible;
-}
-
-.DateRangePicker__picker--invisible {
-  visibility: hidden;
-}
-
 .DateRangePicker__picker--direction-left {
     left: 0;
 }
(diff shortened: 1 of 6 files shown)


majapw:
I'll check out the branch and play around with it.

majapw:
I'm seeing a proptype warning:
```
Warning: Failed prop type: The prop `children` is marked as required in `Portal`, but its value is `null`.
    in Portal (created by SingleDatePicker)
    in SingleDatePicker (created by SingleDatePickerWrapper)
    in SingleDatePickerWrapper
    in div (created by Story)
    in div (created by Story)
    in Story
```

I think this is likely because due to your change, `this.renderDayPicker()` sometimes returns `null`. Maybe that logic should be moved into `this.maybeRenderDayPickerWithPortal()` so as not to render the portal either in the case that the picker is hidden.

lencioni:
@majapw Everything is updated. PTAL!

PR #288, opened 2017-02-02
copy-pasta caused the outsideclickhandler to always be undefined in attached SDPs

majapw:
WOOPS.

@airbnb/webinfra @ljharb halp me pls.


lencioni:
It would be nice to have a test or eslint rule to prevent this kind of thing from happening again

majapw:
the problem is that sometimes we do want it to be undefined, so I'm unsure
what we would test against.

PR #289, opened 2017-02-03
Avoid re-rendering CalendarMonth unless dates overlap

moonboots:
to: @majapw @ljharb 

Currently, each calendar month is re-rendered when users select or hover over any date. This PR avoids updating unaffected months by checking if start, end, and hover date overlap the current month. The extra renders become problematic with the new mobile datepicker I'm working on that has 12 scrollable months.

I believe the current `shouldComponentUpdate` methods don't prevent month re-renders because this [`modifiers` prop is redefined with every render](https://github.com/airbnb/react-dates/blob/master/src/components/DayPickerRangeController.jsx#L241), forcing child `shallowCompare`'s to consistently return true.


moonboots:
Thanks, updated with `shouldComponentUpdate` specs

moonboots:
@majapw I tested these optimizations with `react-addons-perf`, and they dropped the number of `CalendarMonth` renders by 4x (the storybook example I tested had 4 `CalendarMonths`). The changes also cut total render time by 2-3x from 140ms to 60ms.

Here's the test procedure I used for the before/after performance comparison:

* Load a [DateRangePicker story](http://airbnb.io/react-dates/?selectedKind=DateRangePicker&selectedStory=does%20not%20autoclose%20the%20DayPicker%20on%20date%20selection&full=0&down=1&left=1&panelRight=0&downPanel=kadirahq%2Fstorybook-addon-actions%2Factions-panel) with 2 horizontal months (4 total months with the transitional padding)
* Open the DayPicker
* `Perf.start()`
* Hover over 2/28 and click. I was careful to avoid hovering over other days because these will trigger more rendering.
* `Perf.stop();`
* `Perf.printInclusive(Perf.getLastMeasurements());`

Screenshot of optimized shouldComponentUpdate`Perf.printInclusive(...)`:
with optimizations

Screenshot of unoptimized shouldComponentUpdate`Perf.printInclusive(...)`:
without optimizations

I expect these performance improvements to be more dramatic on mobile because of the slower processing and larger number of months I'm attempting to render.

src/components/CalendarMonth.jsx
```diff
@@ -1,14 +1,16 @@
...
+import omit from 'lodash.omit';
+import shallowequal from 'shallowequal';
 import momentPropTypes from 'react-moment-proptypes';
 import moment from 'moment';
 import cx from 'classnames';
 
 import CalendarDay from './CalendarDay';
 
 import getCalendarMonthWeeks from '../utils/getCalendarMonthWeeks';
+import doDateRangesOverlap from '../utils/doDateRangesOverlap';
```
majapw: This is much clearer. Thanks!

PR #291, opened 2017-02-03
Simplify CalendarDay dom

moonboots:
to: @majapw @ljharb 

Simplifies the dom/styles for calendar days from `<td><div><span>31</span></div></td>` to `<td>31</td>`. The appearance remains unchanged.


src/components/CalendarDay.jsx
```diff
@@ -2,15 +2,18 @@ import React, { PropTypes } from 'react';
 import shallowCompare from 'react-addons-shallow-compare';
 import momentPropTypes from 'react-moment-proptypes';
 import moment from 'moment';
+import cx from 'classnames';
 
 const propTypes = {
+  className: PropTypes.string.isRequired,
```
majapw: Don't want to be passing down `className`

majapw:
A lot of the motivation for the previous dom structure was to keep all the table stuff in one file, but that may not be worth it. 

I really don't think we should expose `className` on any component since that's generally an antipattern to everything that we do.

moonboots:
Thanks @majapw, I've removed the `className` prop and instead pushed the `modifier` logic into `CalendarDay`, including the modifier css styles, `getModifiersForDay`, and specs. I think this refactor makes sense even independent of the dom changes because modifiers operate on the day level, not the month.

timhwang21:
Can styling changes be considered 'breaking?' Just a note that this might break styling for anyone who has styled the day cells (e.g. me).

For what it's worth it's an easy fix, but I was *very* confused for a second as to why none of my styles worked any more (package had automatically updated from `6.0.1` to `6.0.2` without me noticing).

Cheers!

majapw:
Whoops! You know the thought had crossed my mind, but you are probably right in that this should ahve been marked as breaking. @ljharb do you have any advice?

ljharb:
If that's the case, we should revert it in the v6 line and publish a v6.0.3, and then re-revert it and publish a v7 (or, find a way to make it not breaking)

JaKXz:
FWIW for `nyc` we use a strategy of publishing [`--tag next`](https://docs.npmjs.com/cli/publish) before going to `@latest` to help mitigate these risks by giving early adopters / contributors a chance to smoke test :) like @timhwang21 this was a surprise breakage for my team.

Thanks for getting back on it quickly though! 👍

majapw:
Hi @JaKXz @timhwang21 I just released v6.1.0 which includes a revert of this change. I'll release v7.0.0 shortly!

PR #295, opened 2017-02-04
Allow modifiers for SingleDatePicker, DateRangePicker

brentmulligan:
Hi, 

At present a modifier prop is not accepted by SingleDatePicker and DateRangePicker components. These components instead rely on a hard coded modifier object. This is problematic if you prefer to use either component but also need to apply a modifier.

This pull request makes the following core changes:

1.  Adds a modifier prop to SingleDatePicker, DateRangePicker and DayPickerRangeController.
2. Preserves the current hard coded modifier values in SingleDatePicker & DayPickerRangeController as defaultModifiers. 
4. Merges defaultModifiers and the new modifier prop into a finalModifier object
5. Passes finalModifier object to DayPicker

I also added two stories to illustrate the changes to DateRangePicker & SingleDatePicker.

Thanks!
Brent

### DIFF
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -61,6 +61,7 @@ const defaultProps = {
   onFocusChange() {},
   onPrevMonthClick() {},
   onNextMonthClick() {},
+  modifiers: {},
 
   // i18n
   displayFormat: () => moment.localeData().longDateFormat('L'),
@@ -190,6 +191,7 @@ export default class DateRangePicker extends React.Component {
       endDate,
       minimumNights,
       keepOpenOnDateSelect,
+      modifiers,
     } = this.props;
     const { dayPickerContainerStyles } = this.state;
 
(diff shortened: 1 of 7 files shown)


majapw:
I am not sure that this is something we want to expose. What modifiers are you missing in the `DateRangePicker`/`SingleDatePicker` that mandate this change?

brentmulligan:
I'm using this during checkout for users to select a "need by date" for t-shirts being screen printed. I want to make a distinction between delivery dates that require a rush shipping option (modified w/ color) and those that would require a non-rush option (not modified). 

I think that there are a lot of other use cases though such as needing to make a distinction for weekends or any specific date where the behavior around that date would be different from others.

majapw:
@brentmulligan we provide a `isDayHighlighted` method for the exact scenarios that you describe. We haven't found that many people seem to need more customizability beyond that and `isDayBlocked` and `isOutsideRange`, especially for the kind of date picking we're aiming to cover. If someone did need a lot of different modifiers, I'd probably expect them to use the `DayPicker` and `DateRangePickerInput`/`SingleDatePickerInput` building blocks to build their own custom interaction.

brentmulligan:
@majapw Ah yes, That is exactly what I need! I don't know how I overlooked it, thank you for pointing it out and sorry for the time waster :)

majapw:
No problem! I hope to... make the documentation slightly less terrible this week, so hopefully that will help you out faster next time! :D

PR #301, opened 2017-02-07
Keyboard Accessibility

majapw:
This is an MVP of keyboard accessibility! I imagine there will still be some polish to take care of, especially when it comes to managing focus and aria roles throughout. BUT I would really like to get this merged so that we can iterate more on it (and add better messaging/validation/accessibility to the inputs themselves as well).

HAVE A GIF!
react-dates

- [x] Fix issue where you navigate from April 2017 to May 2017 with keyboard with outside days enabled
- [X] second param in `onNextMonthClick`/`onPrevMonthClick`???
- [x] Update SingleDatePicker 
- [x] if a CalendarDay has been blurred in the DOM, (ie going to the prev month/next month buttons) but the month transition doesn't trigger a different focusedDate... then the day doesn't get updates and therefore doesn't get refocused. :/  
- [x] unavailable messaging on days
- [X] keyboard shortcuts explanation
- [x] hide keyboard shortcuts menu on touch devices
- [x] write some tests!
- [x] fix vertical orientation keyboard shortcuts panel
- [x] fix safari `?` shortcut
- [x] enable keyboard on `DayPickerRangeController`
- [x] return focus appropriately after the keyboard shortcuts dialog is closed
- [x] adjust phrase props for final copy

@airbnb/webinfra plz give me some love. #

### DIFF
--- a/css/DayPickerKeyboardShortcuts.scss
+++ b/css/DayPickerKeyboardShortcuts.scss
@@ -1,0 +1,161 @@
+.DayPickerKeyboardShortcuts__show,
+.DayPickerKeyboardShortcuts__close {
+  background: none;
+  border: 0;
+  color: inherit;
+  font: inherit;
+  line-height: normal;
+  overflow: visible;
+  padding: 0;
+  cursor: pointer;
+
+  &:active {
+    outline: none;
+  }
+}
+
+.DayPickerKeyboardShortcuts__show {
+  width: 22px;
+  position: absolute;
+}
+
+.DayPickerKeyboardShortcuts__show--bottom-right {
+  border-top: 26px solid $react-dates-color-white;
+  border-right: 33px solid $react-dates-color-primary;
+  bottom: 0;
+  right: 0;
+
+  &:hover {
+    border-right: 33px solid #008489;
+  }
+
+  .DayPickerKeyboardShortcuts__show_span {
+    bottom: 0;
+    right: -28px;
+  }
+}
+
+.DayPickerKeyboardShortcuts__show--top-right {
+  border-bottom: 26px solid $react-dates-color-white;
+  border-right: 33px solid $react-dates-color-primary;
+  top: 0;
+  right: 0;
+
+  &:hover {
+    border-right: 33px solid #008489;
+  }
+
+  .DayPickerKeyboardShortcuts__show_span {
+    top: 1px;
+    right: -28px;
+  }
+}
+
+.DayPickerKeyboardShortcuts__show--top-left {
+  border-bottom: 26px solid $react-dates-color-white;
+  border-left: 33px solid $react-dates-color-primary;
+  top: 0;
+  left: 0;
+
+  &:hover {
+    border-left: 33px solid #008489;
+  }
+
+  .DayPickerKeyboardShortcuts__show_span {
+    top: 1px;
+    left: -28px;
+  }
+}
+
+.DayPickerKeyboardShortcuts__show_span {
+  color: $react-dates-color-white;
+  position: absolute;
+}
+
+.DayPickerKeyboardShortcuts__panel {
+  background: $react-dates-color-white;
+  border: 1px solid $react-dates-color-border;
+  border-radius: 2px;
+  position: absolute;
+  top: 0;
+  bottom: 0;
+  right: 0;
+  left: 0;
+  z-index: 2;
+  padding: 22px;
+  margin: 33px;
+}
+
+.DayPickerKeyboardShortcuts__title {
+  font-size: 16px;
+  font-weight: bold;
+  margin: 0;
+}
+
+.DayPickerKeyboardShortcuts__list {
+  list-style: none;
+  padding: 0;
+
+}
+
+.DayPickerKeyboardShortcuts__close {
+  position: absolute;
+  right: 22px;
+  top: 22px;
+  z-index: 2;
+
+  svg {
+    height: 15px;
+    width: 15px;
+    fill: $react-dates-color-gray-lighter;
+
+    &:hover,
+    &:focus {
+      fill: $react-dates-color-gray-light;
+    }
+  }
+
+  &:active {
+    outline: none;
+  }
+}
+
+.KeyboardShortcutRow {
+  margin: 6px 0;
+}
+
+.KeyboardShortcutRow__key-container {
+  display: inline-block;
+  width: 15%;
+  white-space: nowrap;
+  text-align: right;
+  margin-right: 6px;
+}
+
+.KeyboardShortcutRow__key {
+  font-family: monospace;
+  font-size: 12px;
+  text-transform: uppercase;
+  background: $react-dates-color-gray-lightest;
+  padding: 2px 6px;
+}
+
+.KeyboardShortcutRow__action {
+  display: inline-block;
+}
+
+.DayPickerKeyboardShortcuts__panel--block {
+  .KeyboardShortcutRow {
+    margin-bottom: 16px;
+  }
+
+  .KeyboardShortcutRow__key-container {
+    width: auto;
+    text-align: left;
+    display: inline;
+  }
+
+  .KeyboardShortcutRow__action {
+    display: inline;
+  }
+}
--- a/css/variables.scss
+++ b/css/variables.scss
@@ -5,6 +5,7 @@ $react-dates-spacing-vertical-picker: 72px !default;
 
 $react-dates-color-primary: #00a699 !default;
 $react-dates-color-primary-dark: #00514a !default;
+$react-date-color-primary-dark-1: #008489 !default;
 $react-dates-color-primary-shade-1: #33dacd !default;
 $react-dates-color-primary-shade-2: #66e2da !default;
 $react-dates-color-primary-shade-3: #80e8e0 !default;
@@ -15,6 +16,7 @@ $react-dates-color-gray: #565a5c !default;
 $react-dates-color-gray-dark: darken($react-dates-color-gray, 10.5%) !default;
 $react-dates-color-gray-light: lighten($react-dates-color-gray, 17.8%) !default;   // #82888a
 $react-dates-color-gray-lighter: lighten($react-dates-color-gray, 45%) !default;     // #cacccd
+$react-dates-color-gray-lightest: lighten($react-dates-color-gray, 60%) !default;
 $react-dates-color-highlighted: #ffe8bc !default;
 
 $react-dates-color-border: #dbdbdb !default;
--- a/examples/DateRangePickerWrapper.jsx
+++ b/examples/DateRangePickerWrapper.jsx
@@ -5,6 +5,7 @@ import omit from 'lodash.omit';
 
 import DateRangePicker from '../src/components/DateRangePicker';
 
+import { DateRangePickerPhrases } from '../src/defaultPhrases';
 import DateRangePickerShape from '../src/shapes/DateRangePickerShape';
 import { START_DATE, END_DATE, HORIZONTAL_ORIENTATION, ANCHOR_LEFT } from '../constants';
 import isInclusivelyAfterDay from '../src/utils/isInclusivelyAfterDay';
@@ -74,10 +75,7 @@ const defaultProps = {
   // internationalization
   displayFormat: () => moment.localeData().longDateFormat('L'),
   monthFormat: 'MMMM YYYY',
-  phrases: {
-    closeDatePicker: 'Close',
-    clearDates: 'Clear Dates',
-  },
+  phrases: DateRangePickerPhrases,
 };
 
 class DateRangePickerWrapper extends React.Component {
@@ -112,6 +110,9 @@ class DateRangePickerWrapper extends React.Component {
   render() {
     const { focusedInput, startDate, endDate } = this.state;
 
+    // autoFocus, autoFocusEndDate, initialStartDate and initialEndDate are helper props for the
+    // example wrapper but are not props on the SingleDatePicker itself and
+    // thus, have to be omitted.
     const props = omit(this.props, [
       'autoFocus',
       'autoFocusEndDate',
--- a/examples/SingleDatePickerWrapper.jsx
+++ b/examples/SingleDatePickerWrapper.jsx
@@ -5,6 +5,7 @@ import omit from 'lodash.omit';
 
 import SingleDatePicker from '../src/components/SingleDatePicker';
 
+import { SingleDatePickerPhrases } from '../src/defaultPhrases';
 import SingleDatePickerShape from '../src/shapes/SingleDatePickerShape';
 import { HORIZONTAL_ORIENTATION, ANCHOR_LEFT } from '../constants';
 import isInclusivelyAfterDay from '../src/utils/isInclusivelyAfterDay';
@@ -13,6 +14,7 @@ const propTypes = {
   // example props for the demo
   autoFocus: PropTypes.bool,
   initialDate: momentPropTypes.momentObj,
+
   ...omit(SingleDatePickerShape, [
     'date',
     'onDateChange',
@@ -61,10 +63,7 @@ const defaultProps = {
   // internationalization props
   displayFormat: () => moment.localeData().longDateFormat('L'),
   monthFormat: 'MMMM YYYY',
-  phrases: {
-    closeDatePicker: 'Close',
-    clearDate: 'Clear Date',
-  },
+  phrases: SingleDatePickerPhrases,
 };
 
 class SingleDatePickerWrapper extends React.Component {
@@ -90,9 +89,16 @@ class SingleDatePickerWrapper extends React.Component {
   render() {
     const { focused, date } = this.state;
 
+    // autoFocus and initialDate are helper props for the example wrapper but are not
+    // props on the SingleDatePicker itself and thus, have to be omitted.
+    const props = omit(this.props, [
+      'autoFocus',
+      'initialDate',
+    ]);
+
     return (
       <SingleDatePicker
-        {...this.props}
+        {...props}
         id="date_input"
         date={date}
         focused={focused}
--- a/src/components/CalendarDay.jsx
+++ b/src/components/CalendarDay.jsx
@@ -5,17 +5,25 @@ import { forbidExtraProps, nonNegativeInteger } from 'airbnb-prop-types';
 import moment from 'moment';
 import cx from 'classnames';
 
-import { DAY_SIZE } from '../../constants';
+import { CalendarDayPhrases } from '../defaultPhrases';
+import getPhrasePropTypes from '../utils/getPhrasePropTypes';
+
+import { BLOCKED_MODIFIER, DAY_SIZE } from '../../constants';
 
 const propTypes = forbidExtraProps({
   day: momentPropTypes.momentObj,
   daySize: nonNegativeInteger,
   isOutsideDay: PropTypes.bool,
   modifiers: PropTypes.object,
+  isFocused: PropTypes.bool,
+  tabIndex: PropTypes.oneOf([0, -1]),
   onDayClick: PropTypes.func,
   onDayMouseEnter: PropTypes.func,
   onDayMouseLeave: PropTypes.func,
   renderDay: PropTypes.func,
+
+  // internationalization
+  phrases: PropTypes.shape(getPhrasePropTypes(CalendarDayPhrases)),
 });
 
 const defaultProps = {
@@ -23,10 +31,15 @@ const defaultProps = {
   daySize: DAY_SIZE,
   isOutsideDay: false,
   modifiers: {},
+  isFocused: false,
+  tabIndex: -1,
   onDayClick() {},
   onDayMouseEnter() {},
   onDayMouseLeave() {},
   renderDay: null,
+
+  // internationalization
+  phrases: CalendarDayPhrases,
 };
 
 export function getModifiersForDay(modifiers, day) {
@@ -36,6 +49,15 @@ export function getModifiersForDay(modifiers, day) {
 export default class CalendarDay extends React.Component {
   shouldComponentUpdate(nextProps, nextState) {
     return shallowCompare(this, nextProps, nextState);
+  }
+
+  componentDidUpdate(prevProps) {
+    const { isFocused, tabIndex } = this.props;
+    if (tabIndex === 0) {
+      if (isFocused || tabIndex !== prevProps.tabIndex) {
+        this.buttonRef.focus();
+      }
+    }
   }
 
   onDayClick(day, e) {
@@ -60,29 +82,47 @@ export default class CalendarDay extends React.Component {
       isOutsideDay,
       modifiers,
       renderDay,
+      tabIndex,
+      phrases: { unavailable, available },
     } = this.props;
 
-    const className = cx('CalendarDay', {
-      'CalendarDay--outside': !day || isOutsideDay,
-    }, getModifiersForDay(modifiers, day).map(mod => `CalendarDay--${mod}`));
+    if (!day) return <td />;
 
+    const modifiersForDay = getModifiersForDay(modifiers, day);
+
+    const className = cx('CalendarDay', {
+      'CalendarDay--outside': isOutsideDay,
+    }, modifiersForDay.map(mod => `CalendarDay--${mod}`));
+
+
+    let availabilityText = '';
+    if (BLOCKED_MODIFIER in modifiers) {
+      availabilityText = modifiers[BLOCKED_MODIFIER](day) ? unavailable : available;
+    }
+
+    const ariaLabel = `${availabilityText} ${day.format('dddd')}. ${day.format('LL')}`;
     const daySizeStyles = {
       width: daySize,
       height: daySize - 1,
     };
 
-    return (day ?
-      <td
-        className={className}
-        style={daySizeStyles}
-        onMouseEnter={e => this.onDayMouseEnter(day, e)}
-        onMouseLeave={e => this.onDayMouseLeave(day, e)}
-        onClick={e => this.onDayClick(day, e)}
-      >
-        {renderDay ? renderDay(day) : day.format('D')}
+    return (
+      <td className={className}>
+        <button
+          type="button"
+          ref={(ref) => { this.buttonRef = ref; }}
+          className="CalendarDay__button"
+          aria-label={ariaLabel}
+          style={daySizeStyles}
+          onMouseEnter={(e) => { this.onDayMouseEnter(day, e); }}
+          onMouseLeave={(e) => { this.onDayMouseLeave(day, e); }}
+          onMouseUp={(e) => { e.currentTarget.blur(); }}
+          onClick={(e) => { this.onDayClick(day, e); }}
+          tabIndex={tabIndex}
+        >
+          {renderDay ? renderDay(day) : day.format('D')}
+        </button>
       </td>
-      :
-      <td />
     );
   }
 }
--- a/src/components/CalendarDay.jsx
+++ b/src/components/CalendarDay.jsx
@@ -7,6 +7,7 @@ import cx from 'classnames';
 
 import { CalendarDayPhrases } from '../defaultPhrases';
 import getPhrasePropTypes from '../utils/getPhrasePropTypes';
+import getPhrase from '../utils/getPhrase';
 
 import { BLOCKED_MODIFIER, DAY_SIZE } from '../../constants';
 
@@ -83,7 +84,10 @@ export default class CalendarDay extends React.Component {
       modifiers,
       renderDay,
       tabIndex,
-      phrases: { unavailable, available },
+      phrases: {
+        chooseAvailableDate,
+        dateIsUnavailable,
+      },
     } = this.props;
 
     if (!day) return <td />;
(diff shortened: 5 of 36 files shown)


src/components/CalendarDay.jsx
```diff
@@ -55,20 +64,28 @@ export default class CalendarDay extends React.Component {
...
+          ref={(ref) => { this.buttonRef = ref; }}
+          className="CalendarDay__button"
+          aria-label={`${day.format('dddd')}. ${day.format('LL')}`}
+          onMouseEnter={e => this.onDayMouseEnter(day, e)}
+          onMouseLeave={e => this.onDayMouseLeave(day, e)}
+          onMouseUp={e => e.currentTarget.blur()}
+          onClick={e => this.onDayClick(day, e)}
+          tabIndex={isFocused ? 0 : -1}
+        >
+          {renderDay ? renderDay(day) : day.format('D')}
```
majapw: This may cause problems due to the fact that `renderDay` could return anything. Thoughts?

src/components/DateInput.jsx
```diff
@@ -59,10 +61,10 @@ export default class DateInput extends React.Component {
   }
 
   componentDidUpdate(prevProps) {
-    const { focused } = this.props;
-    if (prevProps.focused === focused) return;
+    const { focused, hasFocus } = this.props;
```
majapw: Need to come up with some better nomenclature to denote component "focus" (ie which is currently being selected) vs. DOM "focus" because this PR separates the two.

src/components/CalendarDay.jsx
ljharb: We originally decided my suggestion of using a component here like `<Day>{renderDay ? renderDay(day) : day.format('D')}</Day>` so that a propType could be applied to the return value - is that perhaps worth doing here?

src/components/DayPicker.jsx
```diff
@@ -174,21 +199,94 @@ export default class DayPicker extends React.Component {
...
+        newFocusedDate.endOf('week');
+        didTransitionMonth = this.maybeTransitionNextMonth(newFocusedDate);
+        break;
+      case 34: // Page Down
+        newFocusedDate.add(1, 'month');
+        didTransitionMonth = this.maybeTransitionNextMonth(newFocusedDate);
+        break;
+
+      case 27: // Escape
+        onBlur();
```
majapw: So this takes focus back to the input, so maybe this name is not quite right.

src/components/DayPicker.jsx
```diff
@@ -174,21 +199,94 @@ export default class DayPicker extends React.Component {
...
+        return;
+      default:
+        break;
+    }
+
+    // If there was a month transition, do not update the focused date until the transition has
+    // completed. Otherwise, attempting to focus on a DOM node may interrupt the CSS animation.
```
majapw: Incidentally, the way this behavior manifests itself is weird AF

src/components/DayPicker.jsx
```diff
@@ -174,21 +199,94 @@ export default class DayPicker extends React.Component {
...
+    if (!didTransitionMonth) {
+      this.setState({
+        focusedDate: newFocusedDate,
+      });
+    }
+  }
+
+  onPrevMonthClick(e, nextFocusedDate) {
```
majapw: So with the regular button click event... `nextFocusedDate` is actually like the native event object or something... which is weird.

majapw: We're passing this here so that we have access to what we want to focus on after the month transition.

ljharb: `e` should always be the last argument

src/components/CalendarDay.jsx
majapw: I'm still not super sure how that would work? What would the `Day` component look like?

ljharb:
```jsx
function Day({ children }) {
  return React.Children.only(children);
}
Day.propTypes = {
  children: PropTypes.node.isRequired,
};
```

and the propType can be as fancy as we want :-)

majapw:
Hi @backwardok @wyattdanger @ljharb @lencioni! I still need to write tests, but I'd love to get some feedback before I do. Would y'all be able to take a look at this PR? I am pushing real hard to get it in this week.

This is the MVP. I think there will likely be some polish work that still needs to be done.

css/DayPickerNavigation.scss
```diff
@@ -23,6 +23,7 @@
 
   &:active {
     background: darken($react-dates-color-white, 5%);
+    // outline: 0;
```
gdborton: nit: remove

css/variables.scss
```diff
@@ -16,6 +17,7 @@ $react-dates-color-gray: #565a5c !default;
 $react-dates-color-gray-dark: darken($react-dates-color-gray, 10.5%) !default;
 $react-dates-color-gray-light: lighten($react-dates-color-gray, 17.8%) !default;   // #82888a
 $react-dates-color-gray-lighter: lighten($react-dates-color-gray, 45%) !default;     // #cacccd
+$react-dates-color-gray-lightest: #eeeeee;
```
gdborton: Isn't this changing the shade of gray by a tiny bit? Why not stick with lighten?

src/components/DayPicker.jsx
```diff
@@ -174,21 +204,107 @@ export default class DayPicker extends React.Component {
...
+        newFocusedDate.endOf('week');
+        didTransitionMonth = this.maybeTransitionNextMonth(newFocusedDate);
+        break;
+      case 'PageDown':
+        newFocusedDate.add(1, 'month');
+        didTransitionMonth = this.maybeTransitionNextMonth(newFocusedDate);
+        break;
+
+      case '?':
+        this.toggleKeyboardShortcuts();
```
raejin: This is probably nitpicking, but can there be an easy way to detect if the shortcut panel was activated by DateInput, hence if we're closing the shortcut panel it would focus back on the DateInput? (I tried played with it but lost myself in managing props)

src/components/CalendarDay.jsx
```diff
@@ -55,23 +75,42 @@ export default class CalendarDay extends React.Component {
...
-        onMouseLeave={e => this.onDayMouseLeave(day, e)}
-        onClick={e => this.onDayClick(day, e)}
-      >
-        {renderDay ? renderDay(day) : day.format('D')}
+      'CalendarDay--outside': isOutsideDay,
+    }, modifiersForDay.map(mod => `CalendarDay--${mod}`));
+
+
+    let availabilityText = '';
+    if (BLOCKED_MODIFIER in modifiers) {
```
raejin: When will `modifiers` not include `BLOCKED_MODIFIER`?

src/components/DayPicker.jsx
majapw: Yeah I kind of wanted to do this... but it was somewhat painful! I will think more on it.

src/components/CalendarDay.jsx
majapw: if you roll your own `DayPicker` wrapper and define your own modifiers

src/components/DayPickerRangeController.jsx
```diff
@@ -157,9 +174,37 @@ export default class DayPickerRangeController extends React.Component {
     });
   }
 
+  getFirstFocusableDay(newMonth) {
```
raejin: Did you plan to add test for this?

examples/DateRangePickerWrapper.jsx
```diff
@@ -73,43 +74,40 @@ const defaultProps = {
...
-  },
+  phrases: DateRangePickerPhrases,
 };
 
 class DateRangePickerWrapper extends React.Component {
   constructor(props) {
     super(props);
 
-    let focusedInput = null;
+    let selectedInput = null;
```
moonboots: Renaming from `focused` to `selected` creates a pretty big diff that might make merging/rebasing other PRs hairy. If `selected` is required in a11y props, can the renaming be limited to external facing html props?

css/DayPickerKeyboardShortcuts.scss
```diff
@@ -0,0 +1,106 @@
...
+  z-index: 2;
+
+  svg {
+    height: 15px;
+    width: 15px;
+    fill: $react-dates-color-gray-lighter;
+
+    &:hover,
+    &:focus {
+      fill: $react-dates-color-gray-light;
```
backwardok: perhaps something to do in a future update, but one thing we came across during a11y testing is that users who have ie9 and enable a setting that essentially turns off background colors and text colors will end up not being able to notice the difference here. We'll likely need to do something where we change the border color or add an outline to the element for more visibility.

examples/SingleDatePickerWrapper.jsx
```diff
@@ -61,43 +63,45 @@ const defaultProps = {
...
+  onSelectChange(selected) {
+    this.setState({ selected });
   }
 
   render() {
-    const { focused, date } = this.state;
+    const { selected, date } = this.state;
+
+    const props = omit(this.props, [
+      'autoFocus',
```
backwardok: Why do these need to be removed?

src/components/CalendarDay.jsx
```diff
@@ -55,23 +75,42 @@ export default class CalendarDay extends React.Component {
...
+      availabilityText = modifiers[BLOCKED_MODIFIER](day) ? unavailable : available;
+    }
+
+    const ariaLabel = `${availabilityText} ${day.format('dddd')}. ${day.format('LL')}`;
```
backwardok: I'm curious whether or not it's more valuable for screen reader users to hear the date first or the availability first. Might be something we want to run through user research at some point

src/components/DateInput.jsx
```diff
@@ -19,14 +20,17 @@ const propTypes = forbidExtraProps({
...
+  onKeyDownQuestionMark: PropTypes.func,
 });
 
 const defaultProps = {
   placeholder: 'Select Date',
   displayValue: '',
   inputValue: '',
   screenReaderMessage: '',
-  focused: false,
+  focused: false, // handles actual DOM focus
```
backwardok: feels like these comments shouldn't need to be duplicated within the `defaultProps` definition

src/components/DateInput.jsx
```diff
@@ -81,13 +87,29 @@ export default class DateInput extends React.Component {
   }
 
   onKeyDown(e) {
```
backwardok: it might be valuable to debounce this event so that it doesn't get triggered as frequently when the user is pressing down a key, especially if they're doing a key combination

src/components/DayPicker.jsx
```diff
@@ -174,21 +204,107 @@ export default class DayPicker extends React.Component {
...
-    if (this.state.monthTransition || !this.state.currentMonth.isSame(prevState.currentMonth)) {
+    const { monthTransition, currentMonth } = this.state;
+    if (monthTransition || !currentMonth.isSame(prevState.currentMonth)) {
       if (this.isHorizontal()) {
         this.adjustDayPickerHeight();
       }
     }
+
+    if (!prevProps.showKeyboardShortcuts && this.props.showKeyboardShortcuts) {
+      // not sure what the right approach is here
```
backwardok: what's the behavior you're trying to accomplish?

src/components/DayPicker.jsx
```diff
@@ -174,21 +204,107 @@ export default class DayPicker extends React.Component {
...
+      this.container.focus();
+    }
+  }
+
+  onKeyDown(e) {
```
backwardok: similar comment here about debouncing

src/components/DayPicker.jsx
```diff
@@ -436,38 +631,63 @@ export default class DayPicker extends React.Component {
...
+    const shouldFocusDate = !isCalendarMonthGridAnimating && !showKeyboardShortcuts;
+
     return (
-      <div className={dayPickerClassNames} style={dayPickerStyle} >
+      <div className={dayPickerClassNames} style={dayPickerStyle}>
         <OutsideClickHandler onOutsideClick={onOutsideClick}>
-          {!verticalScrollable && this.renderNavigation()}
-
-          <div className="DayPicker__week-headers">
+          <div
```
backwardok: I still think these should be part of the table itself. Seems weird that they're essentially the table headers but not actually part of that markup. It also makes it kind of weird for sighted screen reader users

src/components/DayPickerKeyboardShortcuts.jsx
```diff
@@ -0,0 +1,141 @@
...
+            {phrases.keyboardShortcuts}
+          </div>
+
+          <ul className="DayPickerKeyboardShortcuts__list">
+            {keyboardShortcuts.map(({ unicode, label, action }) => (
+              <KeyboardShortcutRow key={label} unicode={unicode} label={label} action={action} />
+            ))}
+          </ul>
+
+          <button
```
backwardok: Since this is visually one of the first things in the dialog, this should probably equivalently be one of the first things in the markup

src/defaultPhrases.js
```diff
@@ -0,0 +1,176 @@
+const closeDatePicker = 'Close';
+const focusStartDate = 'Focus on start date';
```
backwardok: this phrase feels a little off to me - without running this, I have a hard time understanding what it is this is describing. If I were to hear this phrase being said, I wouldn't understand why I would need to focus on the start date. Seems like this should be something more like "Select start date", but I'm not sure if that's exactly what this is accomplishing.

src/defaultPhrases.js
```diff
@@ -0,0 +1,176 @@
...
+const clearDate = 'Clear Date';
+const clearDates = 'Clear Dates';
+const jumpToPrevMonth = 'Jump to previous month';
+const jumpToNextMonth = 'Jump to next month';
+const keyboardShortcuts = 'Keyboard Shortcuts';
+const showKeyboardShortcutsPanel = 'Show the keyboard shortcuts panel';
+const hideKeyboardShortcutsPanel = 'Hide the keyboard shortcuts panel';
+const toggleKeyboardShortcutsPanel = 'Toggle the keyboard shortcuts panel';
```
backwardok: Is the shortcut panel being shown non modal? Seems like you shouldn't be able to access this button when the shortcut panel is shown, in which case it should always be the "show" version.

src/defaultPhrases.js
```diff
@@ -0,0 +1,176 @@
...
+const enterKey = 'Enter key';
+const leftArrowRightArrow = 'Left Arrow/Right Arrow';
+const upArrowDownArrow = 'Up Arrow/Down Arrow';
+const pageUpPageDown = 'Page Up/Page Down';
+const homeEnd = 'Home/End';
+const escape = 'Escape';
+const questionMark = 'Question mark';
+const selectFocusedDate = 'Select the currently focused date';
+const moveFocusByOneDay = 'Decrement/Increment currently focused day by 1 day';
```
backwardok: maybe easier wording as "move forward or backward one day"

backwardok: or "go back or advance one day"

backwardok: (Decrement and increment feel unusual for something that's not a strict number)

src/defaultPhrases.js
```diff
@@ -0,0 +1,176 @@
...
+const moveFocusByOneWeek = 'Decrement/Increment currently focused day by 1 week';
+const moveFocusByOneMonth = 'Decrement/Increment currently focused day by 1 month';
+const moveFocustoStartAndEndOfWeek = 'Navigate to the beginning or end of the currently focused week';
+const returnFocusToInput = 'Return focus to the input field';
```
backwardok: This should probably be "Close date picker", since returning focus isn't really what you're trying to do when you want to close the date picker

src/components/DateInput.jsx
```diff
@@ -81,13 +87,29 @@ export default class DateInput extends React.Component {
...
-      }
+    const {
+      onKeyDownShiftTab,
+      onKeyDownTab,
+      onKeyDownArrowDown,
+      onKeyDownQuestionMark,
+    } = this.props;
+
+    switch (e.key) {
+      case 'Tab':
```
ljharb: this would be shorter and imo cleaner as if/else instead of a switch/case

src/components/DayPicker.jsx
```diff
@@ -311,27 +493,35 @@ export default class DayPicker extends React.Component {
...
       orientation,
+      phrases,
     } = this.props;
 
     let onNextMonthClick;
     if (orientation === VERTICAL_SCROLLABLE) {
       onNextMonthClick = this.multiplyScrollableMonths;
     } else {
-      onNextMonthClick = this.onNextMonthClick;
+      onNextMonthClick = e => this.onNextMonthClick(null, e);
```
ljharb: explicit return?

src/utils/getPhrasePropTypes.js
```diff
@@ -0,0 +1,9 @@
+import { PropTypes } from 'react';
+
+export default function getPhrasePropTypes(defaultPhrases) {
+  return Object.keys(defaultPhrases).reduce((phrases, key) => {
+    const phrasesWithNewKey = { ...phrases };
+    phrasesWithNewKey[key] = PropTypes.node;
+    return phrasesWithNewKey;
```
ljharb: `return { ...phrases, [key]: PropTypes.node };`

src/components/DayPicker.jsx
majapw: Basically, my thought was that if you type `?` inside the date input, probably focus should move to the `DayPicker` instead of the input... but that's pretty hard and also maybe a bad pattern?

src/components/DayPicker.jsx
majapw: Can we put a pin on it until we get design to look at it (slash do it in a follow up PR)? I feel like this is not critical here.

src/components/DayPickerRangeController.jsx
majapw: SOON. first I wanted comments in case I was horribly off the mark.

src/defaultPhrases.js
majapw: We're gonna run this through a copy person before we upgrade react-dates in Airbnb, so maybe let's leave it til then?

src/defaultPhrases.js
majapw: > Seems like you shouldn't be able to access this button when the shortcut panel is shown
I am not sure how to make that happen, but that's certainly not the behavior rn. Could you maybe run the code and lemme know what you think?

src/defaultPhrases.js
majapw: I feel like we can nit the text, but it would probs make sense until someone can look at the copy and give us recs.

examples/SingleDatePickerWrapper.jsx
backwardok: Maybe a comment before that line so it's understood?

src/components/CalendarDay.jsx
backwardok: You can do that, but you'll need to make sure that you add all the related keyboard handling to that element

src/components/DateInput.jsx
backwardok: Is there already a dependency on lodash for this library? Otherwise I'd say it might be worth just creating a util function for that to not have to load an entire library

ljharb: `lodash.debounce` is just that one function, not an "entire library" - imo it's always better to add a dep rather than reimplementing something yourself.

majapw: Debouncing doesn't actually make very much sense here, because the `onKeyDown` listener will still be called for every keystroke, it'll just be called on a delay and there's no way to ostensibly cancel the previous keydown listener. Maybe `throttle is more reasonable?

backwardok: Yep throttle that's the one I wanted :P

majapw:
@airbnb/webinfra @backwardok @ljharb I split out some of the work and broke up the PR into two commits (one for the breaking name change and one for all the a11y work). Would appreciate another set of eyes.

Once again, I still have to write tests! but that is incoming.

src/components/CalendarDay.jsx
```diff
@@ -55,23 +75,42 @@ export default class CalendarDay extends React.Component {
...
+    return (
+      <td className={className}>
+        <button
+          type="button"
+          ref={(ref) => { this.buttonRef = ref; }}
+          className="CalendarDay__button"
+          aria-label={ariaLabel}
+          onMouseEnter={(e) => { this.onDayMouseEnter(day, e); }}
+          onMouseLeave={(e) => { this.onDayMouseLeave(day, e); }}
+          onMouseUp={(e) => { e.currentTarget.blur(); }}
```
backwardok: note for the future - we should just have a nice looking focus state such that doing this isn't necessary. Really don't like that we have this pattern :/

src/components/DateInput.jsx
```diff
@@ -136,14 +157,12 @@ export default class DateInput extends React.Component {
...
-          aria-describedby={screenReaderMessage && screenReaderMessageId}
+          aria-describedby={screenReaderMessageId}
         />
 
-        {screenReaderMessage &&
-          <p id={screenReaderMessageId} className="screen-reader-only">
-            {screenReaderMessage}
-          </p>
-        }
+        <p id={screenReaderMessageId} className="screen-reader-only">
```
backwardok: since `screenReaderMessage` isn't required, it seems like you should only render this if it exists.

src/components/DateInput.jsx
```diff
@@ -136,14 +157,12 @@ export default class DateInput extends React.Component {
           disabled={disabled}
           readOnly={isTouch}
           required={required}
-          aria-describedby={screenReaderMessage && screenReaderMessageId}
```
backwardok: if you were running into issues before where this was coming back as '' and still adding the attribute, then you might consider having the default value of `screenReaderMessage` as `null` or `undefined` so that this works

src/components/DateRangePicker.jsx
```diff
@@ -123,9 +136,45 @@ export default class DateRangePicker extends React.Component {
...
+
+  onDayPickerFocus() {
+    this.setState({
+      isDateRangePickerInputFocused: false,
+      isDayPickerFocused: true,
+      showKeyboardShortcuts: false,
+    });
+  }
+
+  onDayPickerBlur() {
```
backwardok: is there any potential for this to be triggered as a result of triggering the keyboard shortcut modal? or can you not trigger it from the input itself?

src/components/DateRangePicker.jsx
```diff
@@ -180,6 +229,14 @@ export default class DateRangePicker extends React.Component {
     }
   }
 
+  showKeyboardShortcutsPanel() {
+    this.setState({
+      isDateRangePickerInputFocused: false,
+      isDayPickerFocused: true,
```
backwardok: why does showing the keyboard shortcuts panel focus the day picker?

src/components/DateRangePickerInput.jsx
```diff
@@ -146,8 +161,8 @@ export default class DateRangePickerInput extends React.Component {
         {(showDefaultInputIcon || customInputIcon !== null) &&
           <button
             type="button"
-            aria-label={phrases.focusStartDate}
             className="DateRangePickerInput__calendar-icon"
+            aria-label={phrases.focusStartDate}
```
backwardok: not that it matters that much, but curious why you decided to move this

src/components/DateRangePickerInput.jsx
```diff
@@ -132,11 +144,14 @@ export default class DateRangePickerInput extends React.Component {
       customInputIcon,
       customArrowIcon,
       phrases,
+      isFocused,
     } = this.props;
 
     const inputIcon = customInputIcon || (<CalendarIcon />);
     const arrowIcon = customArrowIcon || (<RightArrow />);
 
+    const screenReaderText = screenReaderMessage || phrases.keyboardNavigationInstructions;
```
backwardok: kinda curious about this - if someone were to add in a custom screen reader message here, then the user wouldn't hear about the keyboard navigation instructions, even if they're not mutually exclusive. Since the keyboard controls are not customizable, it seems like the keyboard instructions should always be present.

src/components/DayPicker.jsx
```diff
@@ -179,21 +205,107 @@ export default class DayPicker extends React.Component {
...
+
+    if (isFocused !== this.props.isFocused) {
+      if (isFocused) {
+        const focusedDate = this.getFocusedDay(currentMonth);
+        this.setState({
+          showKeyboardShortcuts,
+          focusedDate,
+        });
+      } else {
+        this.setState({ focusedDate: null });
```
backwardok: having a little bit of a hard time tracking focus state and shortcut panel status but I'm curious why we wouldn't update the `showKeyboardShortcuts` state if it changed. Can't easily tell if it's just always controlled in one place so you don't have to worry about it, but from an API perspective on this component itself, it's a little confusing.

src/components/DayPicker.jsx
backwardok: Whatever had focus prior to triggering the keyboard shortcut menu should be what gets focus when it's closed. When the shortcut menu is open, it should have focus.

src/components/DayPicker.jsx
```diff
@@ -179,21 +205,107 @@ export default class DayPicker extends React.Component {
...
+      case 'PageDown':
+        newFocusedDate.add(1, 'month');
+        didTransitionMonth = this.maybeTransitionNextMonth(newFocusedDate);
+        break;
+
+      case '?':
+        this.toggleKeyboardShortcuts();
+        break;
+
+      case 'Escape':
```
backwardok: one thing to be careful with here - I would expect that if the keyboard shortcut modal was open on top of the date picker modal, that hitting escape would close the shortcut modal and hitting escape again would close the date picker modal. It seems like hitting escape once here does both?

src/components/DayPicker.jsx
backwardok: where does the focus happen after?

backwardok: (and maybe add a comment with a `see [insert function name here]`)

src/components/DayPicker.jsx
```diff
@@ -244,6 +402,15 @@ export default class DayPicker extends React.Component {
     });
   }
 
+  isDayVisible(day, newMonth) {
+    const { numberOfMonths } = this.props;
+    const { currentMonth } = this.state;
+
+    const month = newMonth || currentMonth;
+    return !day.isBefore(month.clone().startOf('month')) &&
```
backwardok: not necessary to fix, but it might make this line a lot easier to read and understand if you pull out the `month.clone()....` pieces into their own `const`s

src/components/SingleDatePickerInput.jsx
```diff
@@ -85,33 +89,39 @@ export default class SingleDatePickerInput extends React.Component {
...
       onClearDate,
       onChange,
       onFocus,
       onKeyDownShiftTab,
       onKeyDownTab,
+      onKeyDownArrowDown,
       screenReaderMessage,
     } = this.props;
 
+    const screenReaderText = screenReaderMessage || phrases.keyboardNavigationInstructions;
```
backwardok: similar comment here as in other file

wyattdanger:
is the `DayPickerRangeController` supposed to work with keyboard in the storybook?

wyattdanger:
Looks like the Keyboard Shortcuts view could use some love in vertical orientation

screenshot 2017-03-09 11 36 14

src/components/DateInput.jsx
```diff
@@ -81,13 +87,24 @@ export default class DateInput extends React.Component {
   }
 
   onKeyDown(e) {
-    const { onKeyDownShiftTab, onKeyDownTab } = this.props;
-    if (e.key === 'Tab') {
```
majapw: Should stop propagation here to prevent shenanigans

src/components/DayPickerKeyboardShortcuts.jsx
```diff
@@ -0,0 +1,141 @@
...
+  phrases: DayPickerKeyboardShortcutsPhrases,
+};
+
+function KeyboardShortcutRow({ unicode, label, action }) {
+  return (
+    <li className="KeyboardShortcutRow">
+      <div
+        className="KeyboardShortcutRow__key-container"
+      >
+        <span className="KeyboardShortcutRow__key" aria-label={label}>
```
lencioni: Should this be `<kbd>` instead of span? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/kbd

wyattdanger: TIL ^

sontek:
@majapw I love this PR!! 

Can I ask a slightly unrelated question?   What were you using to test the accessibility?  What is the black box showing the text?

majapw:
@sontek Thanks! Hopefully this should land soon (I'm waiting on help from a copy person here at Airbnb).

That black box is voiceover on OSX. It's the default voiceover app. :)

src/components/DateInput.jsx
majapw: Figured it out!

majapw:
Hey @airbnb/webinfra, this PR is ready to go! Please take a look and give me your feedback!

src/defaultPhrases.js
```diff
@@ -1,9 +1,36 @@
...
+const moveFocustoStartAndEndOfWeek = 'Go to the first or last day of a week';
+const returnFocusToInput = 'Return to the date input field';
+const keyboardNavigationInstructions = `Press the down arrow key to interact with the calendar and
+  select a date. Press the question mark key to get the keyboard shortcuts for changing dates.`;
+
+// eslint-disable-next-line camelcase
+const chooseAvailableStartDate = ({ checkin_date }) => `Choose ${checkin_date} as your check-in date. It's available.`;
+
+// eslint-disable-next-line camelcase
+const chooseAvailableEndDate = ({ checkout_date }) => `Choose ${checkout_date} as your check-out date. It's available.`;
```
moonboots: What's the plan for integrating these with the internal translation setup?

majapw: So when you look at `getPhrase`, you'll see that each phrase accepts a function, node, or string and behaves accordingly. My plan is to add code to our wrappers that automatically pulls in our translations appropriately and create a phrase bundle that can easily be spread onto any page.

We might want to do this for the more base-level components (ie DayPickerRangeController) we're using as well.

src/components/CalendarDay.jsx
DBjelovuk: This is breaking functionality in IE. https://github.com/airbnb/react-dates/issues/1606

PR #303, opened 2017-02-07
Avoid transition padding months for scrollable

moonboots:
to: @majapw @ljharb 

Another perf PR that avoids rendering the hidden leading and trailing months for the scrollable version of DayPicker. These additional months are used to improve sliding animation for the horizontal and vertical orientation daypickers. The new scrollable daypicker doesn't use this animation, and removing will save some rendering.

Along the way, I refactored and add specs for `getMonths` and `componentWillReceiveProps`.

### DIFF
--- a/src/components/CalendarMonthGrid.jsx
+++ b/src/components/CalendarMonthGrid.jsx
@@ -4,6 +4,7 @@ import shallowCompare from 'react-addons-shallow-compare';
 import momentPropTypes from 'react-moment-proptypes';
 import moment from 'moment';
 import cx from 'classnames';
+import range from 'lodash.range';
 
 import CalendarMonth from './CalendarMonth';
 
@@ -19,6 +20,7 @@ import {
 } from '../../constants';
 
 const propTypes = {
+  addTransitionMonths: PropTypes.bool,
   enableOutsideDays: PropTypes.bool,
   firstVisibleMonthIndex: PropTypes.number,
   initialMonth: momentPropTypes.momentObj,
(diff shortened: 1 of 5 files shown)


src/components/CalendarMonthGrid.jsx
```diff
@@ -84,29 +90,19 @@ export default class CalendarMonthGrid extends React.Component {
...
-        newMonths.push(months[months.length - 1].clone().add(1, 'month'));
-      } else {
-        newMonths = months.slice(0, months.length - 1);
-        newMonths.unshift(months[0].clone().subtract(1, 'month'));
-      }
-    }
 
-    if (hasNumberOfMonthsChanged) {
-      newMonths = getMonths(initialMonth, numberOfMonths);
+    if (hasMonthChanged || hasNumberOfMonthsChanged) {
```
majapw: Why are you removing the performance gains from not creating a new moment object for every month on every transition? Like there was a reason these were separated and that change was v. recent.

src/components/CalendarMonthGrid.jsx
```diff
@@ -54,23 +57,26 @@ const defaultProps = {
...
 };
 
-function getMonths(initialMonth, numberOfMonths) {
-  let month = initialMonth.clone().subtract(1, 'month');
-
-  const months = [];
-  for (let i = 0; i < numberOfMonths + 2; i++) {
-    months.push(month);
-    month = month.clone().add(1, 'month');
+export function getMonths({ initialMonth, numberOfMonths, addTransitionMonths }) {
```
majapw: prefer if you're exporting a function for tests that you separate this out into its own file.

Issue #253, opened 2017-01-17
Blinking cursor in mobile


narmadham:
Hi,

I am using SingleDatePicker with portal option. Once the portal is closed the focused sets back to the input again. It doesnt create a problem in desktop but in mobile I am gettting a blinking cursor. I am able to replicate in the examples as well.

Is there any way to avoid the blinking cursor ?

jonathanong:
I am seeing the same issue and it is messing up my vertical mobile layout. here's what happens immediately after clicking a date input:

![img_0218](https://cloud.githubusercontent.com/assets/643505/22051727/11932e54-dcfa-11e6-9104-38c5a618d0ec.PNG)

I can only get the correct layout when pressing "done"

jonathanong:
this may relevant to https://github.com/airbnb/react-dates/issues/246

narmadham:
Hi @jonathanong  Ya it is almost similar. I tried blurring out after date change but no luck. Reconsidering the usage of this plugin since there are lot of minor issues which are creating an impact.

amlivingamliving:
Hi @jonathanong and @narmadham  - I'm having the same issue. It does look similar to #246 . This is with `withPortal` set to `true` on my `SingleDatePickerWrapper`.

![img_9606](https://cloud.githubusercontent.com/assets/8516645/22128103/eece5a5a-de53-11e6-99f9-6bb891efff4c.PNG)

jurekbarth:
I think its a bug with the `readonly` attribute on an input field. Safari iOS still triggers focus even if `readonly` is set. 
You can test yourself here: http://codepen.io/jurekbarth/pen/PWpEQv

jurekbarth:
For a temporary workaround you could set the disabled attribute like so `disabled={disabled || this.isTouchDevice}` in `DateInput.jsx`

dominicchapman:
@jurekbarth it looks like a change was made back in November having been tested on Surface Tablet, not iOS https://github.com/airbnb/react-dates/pull/174

narmadham:
thanks @jurekbarth this workaround really helped me.

Nehlin:
@jurekbarth Not getting the workaround to work. It's setting the disabled flag to true permanently for ios, right? If so how does that help?

jurekbarth:
@Nehlin how are you trying to solve it? Did you change the file in your `node_modules` folder? If so thats not the way to go, because your bundler catch up the already processed file in the `lib` folder. Hope this helps :)

[...]

PR #304, opened 2017-02-07
Fix blinking cursor on iOS

moonboots:
to: @majapw @ljharb 

Removes the blinking cursor appearing in DateInput on iOS by adding `user-select: none;` ([recommended as a workaround in another project](https://github.com/Dogfalo/materialize/issues/2448#issuecomment-242662049)). This PR should resolve the ["Blinking cursor in mobile" issue](https://github.com/airbnb/react-dates/issues/253).

Here's a gif of the blinking cursor and "Done" bottom bar:
ios-cursor-flash

### DIFF
--- a/css/DateInput.scss
+++ b/css/DateInput.scss
@@ -50,6 +50,13 @@ $caret-top: $react-dates-spacing-vertical-picker - $react-dates-width-tooltip-ar
   border: 0;
   height: 100%;
   width: 100%;
+
+  &[readonly] {
+    -moz-user-select: none;
+    -webkit-user-select: none;
+    -ms-user-select: none;
+    user-select: none;
+  }
 }
 
 .DateInput__display-text {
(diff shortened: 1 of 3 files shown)


css/DateInput.scss
```diff
@@ -50,6 +50,13 @@ $caret-top: $react-dates-spacing-vertical-picker - $react-dates-width-tooltip-ar
   border: 0;
   height: 100%;
   width: 100%;
+
+  &[readonly] {
+    user-select: none;
```
ljharb: the standard property should always go last, so that it takes precedence in browsers that support both.

majapw:
LGTM after @ljharb's comment is addressed. Thanks for this, @moonboots!

majapw:
Cool. Can you squash the commits into one before we merge tho?

PR #305, opened 2017-02-07
Avoid passing false as day

moonboots:
to: @majapw @ljharb 

Small PropType fix for my [previous CalendarDay PR](https://github.com/airbnb/react-dates/pull/291). In `CalendarMonth`, `day` is `false` instead of a moment object for outside days, which fails PropType checks.

screen shot 2017-02-07 at 12 11 17 pm

### DIFF
--- a/test/utils/getCalendarMonthWeeks_spec.js
+++ b/test/utils/getCalendarMonthWeeks_spec.js
@@ -38,6 +38,32 @@ describe('getCalendarMonthWeeks', () => {
     });
   });
 
+  describe('padding when enableOutsideDays is false', () => {
+    let weeksWithPadding;
+
+    before(() => {
+      // using specific month Feb 2017 to manually compare with calendar
+      weeksWithPadding = getCalendarMonthWeeks(moment('2017-02-01'), false);
+    });
+
+    it('null pads leading days', () => {
+      const firstWeek = weeksWithPadding[0];
+      expect(firstWeek[0]).to.equal(null); // Sun Jan 29
+      expect(firstWeek[1]).to.equal(null); // Mon Jan 30
+      expect(firstWeek[2]).to.equal(null); // Tue Jan 31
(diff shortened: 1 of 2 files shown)


src/components/CalendarMonth.jsx
```diff
@@ -100,7 +100,7 @@ export default class CalendarMonth extends React.Component {
               <tr key={i}>
                 {week.map((day, dayOfWeek) => (
                   <CalendarDay
-                    day={day}
+                    day={day ? day : null}
```
majapw: Doesn't it make more sense to return null in `getCalendarMonthWeeks`?

moonboots: Makes sense, moved to `getCalendarMonthWeeks` and added specs

test/utils/getCalendarMonthWeeks_spec.js
```diff
@@ -38,6 +38,32 @@ describe('getCalendarMonthWeeks', () => {
     });
   });
 
+  describe('padding when enableOutsideDays is false', () => {
+    let weeksWithPadding;
+
+    before(() => {
```
ljharb: this should use `beforeEach` - `before` means "before all" so it only runs once.

moonboots: `beforeEach` would work too, but here I intended the before all behavior because I'm not mutating this variable. I wrapped in a before block instead of [defining up top like the other variables](https://github.com/moonboots/react-dates/blob/f99af7d544bf83af652b0135fa6b5e16439499e0/test/utils/getCalendarMonthWeeks_spec.js#L8) as a minor perf improvement to avoid running if `.only(...)` was used elsewhere.

ljharb: We shouldn't worry about perf in tests; `before` should always be avoided.

PR #307, opened 2017-02-07
Custom content in CalendarDay with #renderDay

evandavis:
Replaces #283

```js
/**
 * @param day - the Moment object representing the day
 * @returns {string|element} - a single text or JSX node
 */
 // example
 // print a happy cat on Fridays
 <CalendarDay
   renderDay={day => (day.day() % 6 === 5 ? '😻' : day.format('D'))}
 />
```

screenshot 2017-02-07 13 41 54

### DIFF
--- a/src/components/CalendarDay.jsx
+++ b/src/components/CalendarDay.jsx
@@ -7,19 +7,21 @@ import cx from 'classnames';
 const propTypes = {
   day: momentPropTypes.momentObj,
   isOutsideDay: PropTypes.bool,
+  modifiers: PropTypes.object,
   onDayClick: PropTypes.func,
   onDayMouseEnter: PropTypes.func,
   onDayMouseLeave: PropTypes.func,
-  modifiers: PropTypes.object,
+  renderDay: PropTypes.func,
 };
 
 const defaultProps = {
   day: moment(),
   isOutsideDay: false,
+  modifiers: {},
   onDayClick() {},
   onDayMouseEnter() {},
   onDayMouseLeave() {},
-  modifiers: {},
+  renderDay: null,
 };
 
 export function getModifiersForDay(modifiers, day) {
@@ -51,6 +53,7 @@ export default class CalendarDay extends React.Component {
       day,
       isOutsideDay,
       modifiers,
+      renderDay,
     } = this.props;
 
     const className = cx('CalendarDay', {
@@ -64,7 +67,7 @@ export default class CalendarDay extends React.Component {
         onMouseLeave={e => this.onDayMouseLeave(day, e)}
         onClick={e => this.onDayClick(day, e)}
       >
-        {day.format('D')}
+        {renderDay ? renderDay(day) : day.format('D')}
       </td>
       :
       <td />
--- a/README.md
+++ b/README.md
@@ -212,6 +212,13 @@ The `navPrev` and `navNext` props are used to assign custom icons to the "Next",
   navNext: PropTypes.node,
 ```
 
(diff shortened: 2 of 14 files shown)


src/components/CalendarDay.jsx
```diff
@@ -46,6 +48,16 @@ export default class CalendarDay extends React.Component {
     onDayMouseLeave(day, e);
   }
 
+  renderDay(day) {
+    const { renderDay } = this.props;
+
+    if (renderDay) {
+      const result = renderDay(day);
+      return typeof result === 'string' ? result : Children.only(result);
```
evandavis: @majapw @ljharb `Children.only` doesn't accept strings, so I had to work around it a bit.

majapw: Okay, after more exploration, we kind of realized that we should probably just return the result directly because `Children.only` breaks on `null`, `string`, `number`, `undefined`... etc. X_x I thought this function did something slightly different then it does.

Basically we should just have {renderDay ? renderDay(day) : day.format('D')}` directly in the `<td />` instead.

src/components/CalendarDay.jsx
```diff
@@ -64,7 +76,7 @@ export default class CalendarDay extends React.Component {
         onMouseLeave={e => this.onDayMouseLeave(day, e)}
         onClick={e => this.onDayClick(day, e)}
       >
-        {day.format('D')}
+        {this.renderDay(day)}
```
ljharb: an alternative would be to get rid of the instance method entirely, and just define the defaultProp value for `renderDay` to be `day => day.format('D')`, and do `{renderDay(day)}` inline here?

evandavis: @ljharb the linter requires that we have default props at every level, so we would have to use `renderDay: undefined` instead of `null` in the wrapping components, and it seems like that pattern is discouraged. I think the instance method is easier to reason about given that requirement.

evandavis:
@majapw updated, rebased, and squashed

src/components/CalendarDay.jsx
ljharb: That's a good point, altho defaulting to undefined elsewhere would be fine too.

zachfeldman:
@evandavis I'm trying this out and returning the day, then some content in a span. Works pretty decently but the span gets rendered as plaintext (&lt;, etc). Not sure if this is intentional but once again just throwing in my two cents!

evandavis:
@zachfeldman can you post your `renderDay` function? It sounds like you are returning a string instead of a React node.

zachfeldman:
@evandavis you're right :). Fixed it to a React node and now we're good! Thanks!

evandavis:
@zachfeldman 👍 return an array if you need adjacent nodes: `[day.format('D'), <span />]`

src/components/CalendarDay.jsx
```diff
@@ -64,7 +67,7 @@ export default class CalendarDay extends React.Component {
         onMouseLeave={e => this.onDayMouseLeave(day, e)}
         onClick={e => this.onDayClick(day, e)}
       >
-        {day.format('D')}
+        {renderDay ? renderDay(day) : day.format('D')}
```
Bamieh: it would be better to have the default renderDay function as `(day) => day.format('D')` and just call it always, no need for the if statement

evandavis: @bamieh we already discussed that here: https://github.com/airbnb/react-dates/pull/307#pullrequestreview-20665548

PR #309, opened 2017-02-07
Prevent outside day boxes from resizing on hover

### LINKED ISSUE #308: BUX: hovering over outside days (when enabled) slightly resizes box
feb-07-2017 18-13-30

Note the other outside cells slightly shift in the process of hovering.

timhwang21:
Fixes #308.

Right now because outside days have no border and hovering does, the box for outside days slightly resizes when hovered over.

Adding a 1px invisible border is the easiest way to fix this.

Comparison:

Old: 
feb-07-2017 18-13-30

New: 
feb-07-2017 18-13-36

### DIFF
--- a/css/CalendarDay.scss
+++ b/css/CalendarDay.scss
@@ -26,7 +26,7 @@
 }
 
 .CalendarDay--outside {
-  border: 0;
+  border: 1px solid rgba(0, 0, 0, 0);
   cursor: default;
 
   &:active {


timhwang21:
Ah @majapw sorry, could you roll this back...? My fault for only quickly verifying the fix on the initial screen. On calendars with leading outside dates, you get this:

screen shot 2017-02-08 at 1 22 45 am

I poked around for a bit for a better solution, but admittedly CSS isn't really my thing and I couldn't find any simple, perfect solution.

majapw:
Right. Forgot about the collapsed table border thing. I'll dig into this a bit more, see if I can't find a similarly elegant solution.

Issue #306, opened 2017-02-07
Bug / regression?: incorrect "invalid" dates when startDate/endDate moment object's time is not 12:00:00


timhwang21:
I am using this component together with a TimePicker component. Both are controlled components, so changing the time changes the moments that are currently in the DatePicker.

I upgraded the version of react-dates to major version 5 this week, and noticed a weird interaction when using the DateRangePicker:

* When the time is after `12:00:00`, the date immediately after the start date is marked as out of range.
* When the time is before `12:00:00`, the date immediately before the start date is marked as out of range.

Hopefully the images make this more clear. If you have any questions, please let me know. Thanks!

Case 1: Time before `12:00:00`:
<img width="431" alt="screen shot 2017-02-07 at 4 10 50 pm" src="https://cloud.githubusercontent.com/assets/5831434/22711825/94b4ab9e-ed50-11e6-94ea-b5d6c8cfbf53.png">
<img width="401" alt="screen shot 2017-02-07 at 4 10 56 pm" src="https://cloud.githubusercontent.com/assets/5831434/22711827/94b5c1e6-ed50-11e6-898e-a637e8a31fdf.png">

Case 2: Time after `12:00:00`:
<img width="394" alt="screen shot 2017-02-07 at 4 11 04 pm" src="https://cloud.githubusercontent.com/assets/5831434/22711828/94b5f490-ed50-11e6-8c5b-ebd3ee566fed.png">
<img width="332" alt="screen shot 2017-02-07 at 4 11 08 pm" src="https://cloud.githubusercontent.com/assets/5831434/22711826/94b5888e-ed50-11e6-8c59-93a60c8600b0.png">

majapw:
Huh, I have no idea what would cause this regression. @timhwang21 what version did you upgrade from and are you now on v5.0.0 or v5.2.0?

This is probably coming from the minimum nights calculation I imagine. Can you share your prop values?

timhwang21:
I'll test later today if it's a regression by rolling back the version, didn't have time earlier today

Will also reply with props, I overwrote the blocked dates function but realize I didn't write any minimum nights logic so I'll check that out as well.

Thanks for the quick response!

Tim Hwang
Software Engineer
http://timhwang21.github.io

timhwang21:
Sorry, I actually upgraded from `4.3.2` to `6.0.1`. Problem is still there after rolling back, so it's not a regression. 

[...]

PR #310, opened 2017-02-08
Fix blocked days when moment obj time isn't noon

timhwang21:
References #306.

Currently, the logic for checking if days are blocked by looking at number of days difference in `doesNotMeetMinimumNights()`. However, the day difference rounds down to zero if the time difference is anything < 24hr.

This patch checks if `minimumNights === 1` and if so uses `isSameDay()` instead.

I looked through the specs and couldn't figure out a good place to add a check for this base case, and would appreciate suggestions. 

**Before**
screen shot 2017-02-08 at 2 53 52 am
screen shot 2017-02-08 at 2 54 01 am
(Note that after adding 5 minutes, the next day is now blocked)

**After**
screen shot 2017-02-08 at 2 56 52 am
screen shot 2017-02-08 at 2 56 59 am
(Day is not blocked after adding 5 minutes)
screen shot 2017-02-08 at 2 57 10 am
(Looking @ the console output, we see that the date has successfully been selected)

### DIFF
--- a/src/components/DayPickerRangeController.jsx
+++ b/src/components/DayPickerRangeController.jsx
@@ -163,7 +163,7 @@ export default class DayPickerRangeController extends React.Component {
     if (focusedInput !== END_DATE) return false;
 
     if (startDate) {
-      const dayDiff = day.diff(startDate, 'days');
+      const dayDiff = day.diff(startDate.clone().startOf('day').hour(12), 'days');
       return dayDiff < minimumNights && dayDiff >= 0;
     }
     return isOutsideRange(moment(day).subtract(minimumNights, 'days'));
--- a/test/components/DayPickerRangeController_spec.jsx
+++ b/test/components/DayPickerRangeController_spec.jsx
@@ -12,7 +12,8 @@ import isInclusivelyAfterDay from '../../src/utils/isInclusivelyAfterDay';
 
 import { START_DATE, END_DATE } from '../../constants';
 
-const today = moment().startOf('day');
+// Set to noon to mimic how days in the picker are configured internally
+const today = moment().startOf('day').hours(12);
 
 describe('DayPickerRangeController', () => {
   describe('#render()', () => {
@@ -298,6 +299,19 @@ describe('DayPickerRangeController', () => {
               <DayPickerRangeController
                 focusedInput={END_DATE}
                 startDate={startDate}
+                minimumNights={MIN_NIGHTS}
+              />,
+            );
+            expect(wrapper.instance().doesNotMeetMinimumNights(testDate)).to.equal(false);
+          });
+
+          it('handles time differences of less than 1 full day properly', () => {
+            const partialDate = moment(startDate).add(5, 'minutes');
+            const testDate = moment(startDate).add(MIN_NIGHTS, 'days');
+            const wrapper = shallow(
+              <DayPickerRangeController
+                focusedInput={END_DATE}
+                startDate={partialDate}
                 minimumNights={MIN_NIGHTS}
               />,
             );
--- a/test/components/DateRangePickerInputController_spec.jsx
+++ b/test/components/DateRangePickerInputController_spec.jsx
@@ -17,7 +17,8 @@ import {
   END_DATE,
 } from '../../constants';
 
-const today = moment().startOf('day');
+// Set to noon to mimic how days in the picker are configured internally
+const today = moment().startOf('day').hours(12);
 
 describe('DateRangePickerInputController', () => {
   describe('#render', () => {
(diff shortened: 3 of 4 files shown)


src/components/DayPickerRangeController.jsx
```diff
@@ -160,6 +160,8 @@ export default class DayPickerRangeController extends React.Component {
     if (focusedInput !== END_DATE) return false;
 
     if (startDate) {
+      if (minimumNights === 1) return isSameDay(day, startDate);
```
majapw: This doesn't seem to be an issue that would just be relegated to `minimumNights` not equaling 1... It seems like it would cause problems for any minimumNights value, no?

timhwang21: Oh you're right. I just tested this with my use case, which was without a minimumNights prop (so default 1). Will look into a more robust fix

majapw: So I think the helper methods (`isSameDay`, etc.) all compare both dates at midnight. If we were to change the `dayDiff` line to:
```
      const dayDiff = day.clone().startOf('day').diff(startDate.clone().startOf('day'), 'days');
```
that should work.

timhwang21: I was trying a similar approach just now, but figured I'd describe it here first. I was worried about `clone()`'s performance, so I was doing `dayDiff = day.day() - startDate.day()`. Does this look better to you?

timhwang21: screen shot 2017-02-09 at 7 39 51 pm

That did it, also works for other values of minimum nights.

test/components/DayPickerRangeController_spec.jsx
```diff
@@ -303,6 +303,19 @@ describe('DayPickerRangeController', () => {
             );
             expect(wrapper.instance().doesNotMeetMinimumNights(testDate)).to.equal(false);
           });
+
```
timhwang21: New test fails without above change

src/components/DayPickerRangeController.jsx
```diff
@@ -163,7 +163,7 @@ export default class DayPickerRangeController extends React.Component {
     if (focusedInput !== END_DATE) return false;
 
     if (startDate) {
-      const dayDiff = day.diff(startDate, 'days');
+      const dayDiff = day.day() - startDate.day();
```
majapw: This is not going to work over week borders as it will only return 0-6 and `.date()` and `.dayOfYear()` have similar problems over month and year boundaries respectively. I think unfortunately, something like I recommended earlier with comparing the `startOf` both days seems the most reasonable. Alternatively, because you are guaranteed that `day` will always be at noon, you could set `startDate.hour('12)` and compare those, although that is more brittle.

timhwang21: Ugh, I meant `date()` which gets date of month and not `day()`. But that has the same issue with year boundaries. I then tried comparing both `.year()` and then comparing `.dayOfYear()`, but it seems like this is a lot less performant than the clone strategy, which surprised me. So looks like yes, that is the way to go. 

For what it's worth, I also tested skipping setting `day`, and it's quite a bit better. What do you think?

I can go ahead with whichever strategy is the most ideal.

Thanks!

screen shot 2017-02-10 at 5 11 41 pm

majapw: Nice. Yeah, let's skip doing it for day, but because of https://github.com/airbnb/react-dates/blob/master/src/utils/getCalendarMonthWeeks.js#L5, let's actually do 
```
day.diff(startDate.clone().startOf('day').hour(12), 'days');
```
just to cover all possible edge cases

timhwang21:
So it looks like this is failing because the spec has this definition of 'today':

`const today = moment().startOf('day');`

Changing it to reflect the internal values in the actual picker (noon) fixes the issue. However, I was just hoping for a sanity check on that test change.

test/components/DayPickerRangeController_spec.jsx
```diff
@@ -12,7 +12,7 @@ import isInclusivelyAfterDay from '../../src/utils/isInclusivelyAfterDay';
 
 import { START_DATE, END_DATE } from '../../constants';
 
-const today = moment().startOf('day');
+const today = moment().startOf('day').hours(12);
```
majapw: This makes sense. Maybe add a comment about how this is due to the fact that all moment objects passed to interaction handlers from `CalendarDay` are set to noon. Not necessary but could be nice.

timhwang21: That 'today' is in three other specs. For consistency I think it would be better to bring them all in line with this. I can comment on each.

majapw: SGTM

thecotne:
just spend about hour and half finding why my app was broken

and i need to say this is stupid (and i don't see this documented anywhere)

so if for example i take values from api/server i need to set hours to 12 in order to get startDate correctly highlighted in UI

PR #313, opened 2017-02-08
Replace react-addons-shallow-compare by React.PureComponent

7rulnik:
[shallow-compare](https://facebook.github.io/react/docs/shallow-compare.html) is a legacy since react@15.3.0. It should be major update.

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Ensure packages are installed with correct version numbers by running:
   Which produces and runs a command like:
 
   ```sh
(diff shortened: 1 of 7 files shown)


ljharb:
Why would we want to drop 0.14 support?

Since this package won't be eliminated in v15, there's zero reason to drop it until v16 comes out.

7rulnik:
Sure, we should merge it when v16 comes out

ljharb:
Technically we'd only be able to merge it once we drop v15 support, which won't be at least until v17 comes out.

Closing this for the time being.

PR #316, opened 2017-02-10
chore: use nyc over istanbul-cli

JaKXz:
👋 Hi! `istanbuljs` member here -

Thought I'd help out a little by simplifying your coverage commands and configuration :) 

Otherwise, the main win of bringing you over to the much better supported `nyc` + `babel-plugin-istanbul` is that you're now getting _more accurate coverage reports and stats!_

Please let me know if anything concerns you - very happy to answer any questions you might have. Thanks again for this wonderful library!


ljharb:
Can you elaborate more on why we should change our coverage method? What's different about `nyc` and why should we prefer it over `istanbul`?

.editorconfig
```diff
@@ -0,0 +1,16 @@
+root = true
```
ljharb: unless this file can be kept in sync with CI (ie, changes to it will cause misformatted files to fail) i'd prefer the file not be included.

test/components/CalendarDay_spec.jsx
```diff
@@ -30,7 +30,7 @@ describe('CalendarDay', () => {
       expect(wrapper.text()).to.equal(lastOfMonth.format('D'));
     });
 
-    it('contains arbitrary content if renderDay is provided', () => {
+    it.skip('contains arbitrary content if renderDay is provided', () => {
```
ljharb: just noting this line will need reverting before merging

test/mocha.opts
```diff
@@ -0,0 +1,5 @@
+--compilers js:babel-register,jsx:babel-register
```
ljharb: this is a good change regardless; i'd be stoked to merge it in a separate PR :-)

test/mocha.opts
```diff
@@ -0,0 +1,5 @@
...
+--require airbnb-js-shims
+--require ./test/_helpers/ignoreSVGStrings.jsx
+--recursive
+test/**/*_spec.js test/**/*_spec.jsx test/_helpers/restoreSinonStubs.js
```
ljharb: however, the glob part of this configuration should remain in package.json, not in the mocha opts.

JaKXz:
> What's different about nyc and why should we prefer it over istanbul?

`nyc` is the successor to the istanbul cli. `nyc` uses istanbul's APIs under the hood and there have been some features & fixes since `istanbul@1.x-alpha` was published.

It's preferred over the istanbul cli because 
- the istanbul cli is no longer really maintained, or, better stated: future development is happening in `nyc`. 
- `babel-plugin-istanbul` + `nyc` provides [more accurate coverage of ES2015+ code](https://github.com/dtinth/babel-plugin-__coverage__#is-the-resulting-coverage-differ-from-istanbulisparta) [hence coverage numbers dropping, relatively signficantly].

.editorconfig
JaKXz: I have found [EditorConfig](http://editorconfig.org) to be very helpful as a maintainer because it takes care of nuisances with whitespace and line endings and newlines etc. behind the scenes. e.g. the IntelliJ platform comes with support built in and almost all popular editors have a simple installer. It's worth noting that [GitHub also has native support](https://github.com/editorconfig/editorconfig.github.com/pull/48).

If you still prefer it not be added, I will remove it.

ljharb: I'd love to have it, but something in CI would need to enforce it (the linter typically does this for us). Do you know of anything that can do that?

package.json
```diff
@@ -9,14 +9,11 @@
...
-    "cover:clean": "rimraf coverage",
-    "cover:check": "babel-node ./node_modules/istanbul/lib/cli check-coverage",
-    "cover:mocha": "babel-node ./node_modules/istanbul/lib/cli cover node_modules/mocha/bin/_mocha -- --compilers js:babel-register,jsx:babel-register --require airbnb-js-shims --recursive ./test/_helpers test/**/*",
+    "precover": "rimraf coverage",
+    "cover": "cross-env NODE_ENV=test nyc mocha",
     "lint": "eslint --ext .js,.jsx src test",
-    "mocha": "mocha --compilers js:babel-register,jsx:babel-register --require airbnb-js-shims --recursive ./test/_helpers",
     "pretest": "npm run --silent lint",
-    "tests-only": "npm run mocha --silent test",
+    "tests-only": "mocha --silent",
```
ljharb: I think the file path needs to be restored here?

.editorconfig
JaKXz: https://github.com/jedmao/eclint looks promising!

JaKXz: But I don't feel like it's a strong enough win for you, if it must be enforced like this.

I will remove it from this PR since it's not really the main focus.

package.json
JaKXz: This was a typo on my part - the `--recursive` flag handles the globs for us.

JaKXz:
@ljharb mind elaborating on

> lockfiles are harmful in non-top-level-apps.

If you publish with a `files` array in `package.json` you're not adding any weight to the package as a dependency. I notice this package uses `.npmignore` and I missed adding it there, so that was my bad. Is there something I'm missing?

ljharb:
`files` arrays are incredibly dangerous, for one, but npmignore is fine.

If I publish a shrinkwrap file, yarn ignores it. If I publish a yarn lock file, yarn and npm ignore it. In no case am I ensuring my users get a deterministic dep tree. However, I am ensuring that if an in-range update is published that breaks me, my own tests will merrily pass but my users will merrily break.

Only an app is in the position to truly know that dependencies should be frozen.

JaKXz:
Curiouser and curiouser

> files arrays are incredibly dangerous, for one

Mind elaborating on that?

ljharb:
@JaKXz i'm on a number of common slacks and on freenode IRC, but here's fine too.

Basically, the absolute worst case of forgetting to include a new file in `npmignore` is that "your package is a tiny bit too large". The worst case of forgetting to include a new file in "files" is that "your package is subtly broken, in a way that's almost impossible for you, the author, (or your CI tests) to determine". Overarching package size just *isn't a big problem* anyways, and the worst case is *so* much worse with an "inclusion list" approach, that the only safe approach is an "exclusion list".

JaKXz:
I see. That's a reasonable approach to safety. 

Anyway, are there any other changes you want to see in this PR?

majapw:
@ljharb what do we need to do to get this merged in?

test/mocha.opts
```diff
@@ -0,0 +1,4 @@
+--compilers js:babel-register,jsx:babel-register
+--require airbnb-js-shims
+--require ./test/_helpers/ignoreSVGStrings.jsx
```
ljharb: This fails to include the file that restores sinon stubs; please revert this part and restore `npm run mocha`, and have that pass the file there. `mocha` doesn't work with global before/after handlers if they're `--require`d in.

ljharb:
@JaKXz i've rebased this branch; but it seems like your previous command (`nyc mocha`) is a bit wonky; it doesn't seem to take the same path arguments as `mocha` does; and we prefer to use `npm run mocha` as our central source of truth for the mocha command. Separately, adding the path args to `mocha.opts` doesn't seem to work either.

Is this not something `nyc` supports?

JaKXz:
@ljharb sorry for the long turnaround on this! I think I've addressed your concerns now. :)

package.json
```diff
@@ -9,14 +9,12 @@
...
     "clean": "rimraf lib",
-    "cover": "npm run cover:clean && npm run cover:mocha && npm run cover:check",
-    "cover:clean": "rimraf coverage",
-    "cover:check": "babel-node ./node_modules/istanbul/lib/cli check-coverage",
-    "cover:mocha": "babel-node ./node_modules/istanbul/lib/cli cover node_modules/mocha/bin/_mocha -- --compilers js:babel-register,jsx:babel-register --require airbnb-js-shims --recursive ./test/_helpers test/**/*",
+    "precover": "rimraf coverage",
+    "cover": "cross-env NODE_ENV=test nyc npm run mocha",
     "lint": "eslint --ext .js,.jsx src test",
-    "mocha": "mocha --compilers js:babel-register,jsx:babel-register --require airbnb-js-shims --recursive ./test/_helpers",
+    "mocha": "mocha ./test/_helpers test/**/*_spec.js*",
```
ljharb: is the `_spec.js` part needed? Every JS file inside `test` should be considered a test, and double-running the helpers shouldn't be a problem :-/

package.json
```diff
@@ -9,14 +9,12 @@
...
     "pretest": "npm run --silent lint",
-    "tests-only": "npm run mocha --silent test",
+    "tests-only": "npm run mocha -- --silent",
```
ljharb: the `--silent` was intended for `npm`, not `mocha` - this change passes it to `mocha` instead. Can this line be reverted?

.nycrc
```diff
@@ -0,0 +1,24 @@
...
+  "require": [
+    "babel-register"
+  ],
+  "reporter": [
+    "text",
+    "lcov"
+  ],
+  "all": true,
+  "check-coverage": true,
+  "statements": 89,
```
ljharb: just one final re-confirmation - are these numbers exactly in line with our current coverage? If not, can we bump them to be? :-)

JaKXz:
@ljharb here are the original watermarks, and I've dropped the numbers in `.nycrc` to reflect the current coverage.

screen shot 2017-03-31 at 8 58 35 pm

package.json
JaKXz: fixed.

package.json
JaKXz: fixed.

majapw:
Party! Why did coveralls coverage go down so significantly though?

JaKXz:
Because the instrumentation is much more accurate now, and we are measuring against _all_ JS and jsx files :) which I don't think was happening before.

majapw:
Exciting! Also, I think I need to write more tests. :P @ljharb is this good? Can we merge it in? I'll lower the coveralls check for now and admin priv merge it, if that's the case.

Thanks for all the hard work @JaKXz! We super appreciate it. :D

ljharb:
I've just re-rebased and pushed up; if travis-ci passes (even if coveralls drops) then this is good to merge 🎉

ljharb:
Coveralls passed which probably means the thresholds aren't high enough; but I'm going to merge this now!

ljharb:
(I filed #417 to combine coverage across all test runs, instead of just on node 4.6)

ljharb:
Thanks @JaKXz! 🎉

majapw:
Thanks y'all! @ljharb, I lowered the threshold so all following prs
wouldn't be breaking, but yes, i agree that they are too low.

PR #319, opened 2017-02-10
Use consolidated-events for adding event listeners

lencioni:
This is the same package that react-waypoint uses to efficiently add
event listeners to nodes. It has two important features:

1. It will try to create as few event listeners as possible by
   delegating to a single event listener on the node. This should
   improve performance, especially when combined with other modules that
   are using consolidated-events.

2. It allows the third argument to be safely used by browsers that
   support it. In other words, we will be able to use safely passive
   event listeners in browsers that support them. This should improve
   performance.

---

Use passive event listeners for window resize

By default, any event listener has the option to call
`e.preventDefault()` to prevent the default event from happening. This
means that the browser has to wait for the event handler to finish
before it can perform the default. For events like scrolling, this can
have a dramatically negative impact on the scrolling animation, which
needs to run at 60fps in order for it to feel good.

Thankfully, modern browsers offer us an escape hatch: passive event
listeners. By setting this flag, we are saying that `e.preventDefault()`
will never be called by the event handler. This makes it so the browser
is free to run the default.

In the case of a scroll event, this is the scroll animation. In the case
of window resize, I don't think this will have much of an effect. But, I
don't see any harm in doing it.

To: @majapw @mikefowler

### DIFF
--- a/src/components/OutsideClickHandler.jsx
+++ b/src/components/OutsideClickHandler.jsx
@@ -1,4 +1,5 @@
 import React, { PropTypes } from 'react';
+import { addEventListener, removeEventListener } from 'consolidated-events';
 
 const propTypes = {
   children: PropTypes.node,
@@ -17,21 +18,18 @@ export default class OutsideClickHandler extends React.Component {
   }
 
   componentDidMount() {
-    if (document.addEventListener) {
-      // `useCapture` flag is set to true so that a `stopPropagation` in the children will
-      // not prevent all outside click handlers from firing - maja
-      document.addEventListener('click', this.onOutsideClick, true);
-    } else {
-      document.attachEvent('onclick', this.onOutsideClick);
-    }
+    // `capture` flag is set to true so that a `stopPropagation` in the children
+    // will not prevent all outside click handlers from firing - maja
+    this.clickHandle = addEventListener(
+      document,
+      'click',
+      this.onOutsideClick,
+      { capture: true },
+    );
   }
 
   componentWillUnmount() {
-    if (document.removeEventListener) {
-      document.removeEventListener('click', this.onOutsideClick, true);
-    } else {
-      document.detachEvent('onclick', this.onOutsideClick);
-    }
+    removeEventListener(this.clickHandle);
   }
 
   onOutsideClick(e) {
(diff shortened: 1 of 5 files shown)


src/components/OutsideClickHandler.jsx
```diff
@@ -18,21 +19,18 @@ export default class OutsideClickHandler extends React.Component {
...
-    if (document.addEventListener) {
-      // `useCapture` flag is set to true so that a `stopPropagation` in the children will
-      // not prevent all outside click handlers from firing - maja
-      document.addEventListener('click', this.onOutsideClick, true);
-    } else {
-      document.attachEvent('onclick', this.onOutsideClick);
-    }
+    // `capture` flag is set to true so that a `stopPropagation` in the children
+    // will not prevent all outside click handlers from firing - maja
+    this.clickHandle = addEventListener(
```
ljharb: can you confirm that consolidated-events has an attachEvent fallback?

lencioni: It does not. We can add it if it is important.

lencioni: Here's where it would go: https://github.com/lencioni/consolidated-events/blob/603e9c9c54f3367cf4ca401672447e66d0731fcd/src/TargetEventHandlers.js#L44

lencioni: However, since the fallback wasn't used in all of the places react-dates was adding event listeners, I don't think this should be a blocker for using consolidated-events in this project.

ljharb: By not having it, this PR is a breaking change.

I think it'd be good to add, but if we want this to be semver-major, then that's fine too.

lencioni: I'm not convinced that this PR is a breaking change, since the fallback was only in one place.

ljharb: For anyone importing OutsideClickHandler directly, and using only that component, it would be a breaking change.

lencioni: Only if we consider OutsideClickHandler part of the public API. I don't think we do.

majapw: Yeah we don't actually export it in `index.js`. Also `attachEvent` is a relic of back when we cared about IE8.

lencioni:
Merge conflicts have been resolved. PTAL

PR #322, opened 2017-02-10
Fix lint warnings where possible

majapw:
This is the more up to date version of https://github.com/airbnb/react-dates/pull/99.

Basically, remove all the lint warnings that we can. There are a few places where `ReactDOM.findDOMNode` is still in place with disabling, because that's necessary to return the actual dom node representing react components, but those should be removed by https://github.com/airbnb/react-dates/pull/137 eventually. Similarly, I left the `jsx-a11y/no-static-element-interactions` rule ignored because those should all be fixed by https://github.com/airbnb/react-dates/pull/301.

Woo!

to: @ljharb @airbnb/webinfra

### DIFF
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -132,7 +132,7 @@ export default class DateRangePicker extends React.Component {
   }
 
   getDayPickerDOMNode() {
-    return ReactDOM.findDOMNode(this.dayPicker);
+    return ReactDOM.findDOMNode(this.dayPicker); // eslint-disable-line react/no-find-dom-node
   }
 
   isOpened() {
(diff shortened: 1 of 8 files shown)


src/components/DateRangePicker.jsx
```diff
@@ -132,7 +132,8 @@ export default class DateRangePicker extends React.Component {
   }
 
   getDayPickerDOMNode() {
-    return ReactDOM.findDOMNode(this.dayPicker);
+    // eslint-ignore-next-line
```
ljharb: is this comment needed if the next one ignores too?

majapw: oh woops

PR #324, opened 2017-02-11
Typescript definitions

Artur-A:
Added typescript definitions for react controls, similar to `index.js`.

Definitions will be taken automatically by `tsc` from `index.d.ts`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,0 +1,358 @@
+import moment from "moment";
+import * as React from "react";
+
+export = ReactDates;
+
+declare namespace momentPropTypes{
+    type momentObj = any;
+    type momentString = any;
+    type momentDurationObj = any;
+}
+
+    
+declare namespace ReactDates{
+    type AnchorDirectionShape = 'left' | 'right';
+    type FocusedInputShape = 'startDate' | 'endDate';    
+    type OrientationShape = 'horizontal' | 'vertical';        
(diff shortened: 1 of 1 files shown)


Artur-A:
Jordan, Thank you for your answer.

Microsoft still faces a huge amount of problems with Typings, [here](https://github.com/Microsoft/TypeScript/issues/9184) is the current discussion.

Making TS Types apart from the project rises two inconvenient things:
1) somebody outside from the project should constantly review the changes and update `.d.ts`, which occurs not so good for many reasons
2) `.d.ts` starts to having its own versions different from the modern targeting repo, it's always obsolete
3) there appears a need to download additional `/@types/` npm package

So Microsoft recommends to put definition types as part of the main project, which will be up to date and ready for a use.

There are a lot of TypeScript developers and `.d.ts` will make them able to use `react-dates` controls.  

Please write if you are still cannot put `.d.ts` inside the root of project for some reasons then I will add types to a seperate repo `@types/react-dates`.

ljharb:
If it lives in this repo, we're responsible for maintaining their accuracy - and we don't use typescript and don't wish to.

If they're in a separate place, nobody has any expectation that the typing so will be accurate. I think a separate repo is best, pending the outcome of the discussions you referenced (which I am personally a part of).

Artur-A:
Alright, then it seems reasonable to move types to a seperate repo.

PR #332, opened 2017-02-16
Clean up prop types definitions for top-level components

majapw:
This change does the following
- Make `onDate(s)Change` and `onFocusChange` required
- Sort `DateRangePicker`, `SingleDatePicker`, `DayPicker` by category with comments
- Use `forbidExtraProps` on all components
- Add propTypes and defaultProps to the examples so that information shows up in the info addon

I would like ideally to have `focusedInput`/`startDate`/`endDate` and `focused`/`date` to be required as well, but since those can take a value of `null` that's a bit trickier. I might make a validator that allows for this (something like `isDefined`).

to: @airbnb/webinfra

### DIFF
--- a/examples/DateRangePickerWrapper.jsx
+++ b/examples/DateRangePickerWrapper.jsx
@@ -1,6 +1,68 @@
 import React from 'react';
+import moment from 'moment';
+import omit from 'lodash.omit';
 
 import DateRangePicker from '../src/components/DateRangePicker';
+
+import DateRangePickerShape from '../src/shapes/DateRangePickerShape';
+import { START_DATE, END_DATE, HORIZONTAL_ORIENTATION, ANCHOR_LEFT } from '../constants';
+import isInclusivelyAfterDay from '../src/utils/isInclusivelyAfterDay';
+
+const propTypes = omit(DateRangePickerShape, [
+  'startDate',
+  'endDate',
+  'onDatesChange',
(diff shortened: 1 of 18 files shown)


examples/DateRangePickerWrapper.jsx
```diff
@@ -1,7 +1,69 @@
...
+import moment from 'moment';
+import omit from 'lodash.omit';
 
 import DateRangePicker from '../src/components/DateRangePicker';
 
+import DateRangePickerShape from '../src/shapes/DateRangePickerShape';
+import { START_DATE, END_DATE, HORIZONTAL_ORIENTATION, ANCHOR_LEFT } from '../constants';
+import isInclusivelyAfterDay from '../src/utils/isInclusivelyAfterDay';
+
+const propTypes = omit(DateRangePickerShape, [
```
lencioni: You could use object rest here instead of omit and avoid the dependency,

majapw: It's a dev dependency so I didn't think it was a big deal. Doesn't using object rest require unused vars to be ignored?

lencioni: Yep

ljharb:
`airbnb-prop-types` now contains `explicitNull`, if that's useful here?

majapw:
Sweet. I'll add that in. :)

PR #336, opened 2017-02-20
[Fix] `DateInput`: `isTouchDevice` should only be determined in `componentDidMount`

### LINKED ISSUE #335: Server Rendering Checksum Error
When server-rendering we get a checksum error because `DateInput` is using `isTouchDevice` to set the `readOnly` attribute on the input. `isTouchDevice` is always rendering `false` on the server (since `window` isn't defined) regardless of whether the client device is actually a touchDevice. What's worse, this causes checksum errors even if the device is not a touch device anyway:

```
       Warning: React attempted to reuse markup in a container but the checksum was invalid. This generally means that you are using server rendering and the markup generated on the server was not what the client was expecting. React injected new markup to compensate which works but you have lost many of the benefits of server rendering. Instead, figure out why the markup being generated is different on the client or server:
        (client)  autocomplete="off" readonly="" aria-des
        (server)  autocomplete="off" aria-describedby=""
```

One way to fix this might be to move the device type check to run only after the component mounts and just use a constant value when on the server and on the initial render. Or, alternatively, you could allow the user to specify the value themselves if they so choose and thus figure out their own fix, such as guessing the device type on the server via the request's user agent header (similar to this discussion for media queries in react-responsive: https://github.com/contra/react-responsive/issues/43).

ljharb:
Fixes #335.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -9,5 +9,6 @@
   "rules": {
     "no-mixed-operators": [2, { "allowSamePrecedence": true }],
     "jsx-a11y/no-static-element-interactions": 1, // TODO: enable
+    "react/no-did-mount-set-state": 0, // necessary for server-rendering
   }
 }
--- a/src/components/DateInput.jsx
+++ b/src/components/DateInput.jsx
@@ -42,12 +42,15 @@ export default class DateInput extends React.Component {
     super(props);
     this.state = {
       dateString: '',
+      isTouchDevice: false,
     };
 
     this.onChange = this.onChange.bind(this);
     this.onKeyDown = this.onKeyDown.bind(this);
+  }
 
-    this.isTouchDevice = isTouchDevice();
+  componentDidMount() {
+    this.setState({ isTouchDevice: isTouchDevice() });
   }
 
   componentWillReceiveProps(nextProps) {
@@ -89,7 +92,10 @@ export default class DateInput extends React.Component {
   }
 
   render() {
-    const { dateString } = this.state;
+    const {
+      dateString,
+      isTouchDevice: isTouch,
+    } = this.state;
     const {
       id,
       placeholder,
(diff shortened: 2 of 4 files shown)


src/components/DateInput.jsx
```diff
@@ -89,7 +92,10 @@ export default class DateInput extends React.Component {
   }
 
   render() {
-    const { dateString } = this.state;
+    const {
+      dateString,
+      isTouchDevice: isTouch,
```
majapw: I'm confused about the need for this name change. Can't we just keep this as `isTouchDevice`?

.eslintrc
```diff
@@ -9,5 +9,6 @@
   "rules": {
     "no-mixed-operators": [2, { "allowSamePrecedence": true }],
     "jsx-a11y/no-static-element-interactions": 1, // TODO: enable
+    "react/no-did-mount-set-state": 0, // necessary for server-rendering
```
majapw: Can we exclusively put this on the `DateInput` instead of as a universal exception?

ljharb: I can, but it should be a universal exception, because it's necessary for *any* component. I need to disable this rule upstream in the airbnb config too - `setState` in `componentDidMount` is a good thing.

src/components/DateInput.jsx
ljharb: That identifier name is already used on [line 5](https://github.com/airbnb/react-dates/pull/336/files#diff-f090ce99cecab0a8a3afc65f42d45527R5)

.eslintrc
majapw: Really?

src/components/DateInput.jsx
majapw: I see

.eslintrc
ljharb: Every single one of airbnb's hypernova components works like that already. We can definitely strive to avoid it, but it's not possible to avoid for client-required functionality.

majapw: I guess that's fair, but given that this repo is small, I would prefer to turn it off on a case by case basis.

ljharb: If you prefer that I can do it - but we might need to make the same change in any component that has touch detection.

Issue #278, opened 2017-01-31
Option for input icon for end date


JemarJones:
The input icon add in #222 is awesome and definitely useful. However its only available in relation to the start date. I have requirements to have the exact same thing done right next to the EndDate as well, like so:
![screen shot 2017-01-31 at 3 58 41 pm](https://cloud.githubusercontent.com/assets/6874590/22484234/6a5841ee-e7ce-11e6-9086-c860cb1ff02f.png)
 I figure that others may also have similar requirements, would this project be willing to add such a feature? If that is the case, i am willing to put up a PR for it!

majapw:
Yeah I think we could move this logic in the `DateInput` component perhaps. :) Happy to accept a PR.

nezed:
@majapw The feature in #338 may be universal solution for same cases 😉
@JemarJones currently you can make some workaround with `customArrowIcon` prop

JemarJones:
@nezed Hah good point, the option wasn't quite [around just yet](https://github.com/airbnb/react-dates/issues/277) when i made this 😜.

JemarJones:
In regards to #338 resolving this, i think it may not be the best solution to have to fully visually reimplement (and keep up to date with any future changes to) DateInput just to add the input icon.

nezed:
@JemarJones i have made this as common solution for cases with appending and prepending elements, and complex formatting cases.
Its deal between producing a few more props and partial content overriding by user.
Whereas in terms of the fact that the main components has already a lot of props, and may require more in future, i think the proposed renderer is optimal compromise.

JemarJones:
@nezed Those are good points. What if alternatively, a prop was very similar to yours was added called something like `wrapInput` that received the node (i haven't checked which node that is) that currently renders the input, and could simply use that as a child if it chooses? I think that could be a nice solution because it makes things like this issue a case of adding only a small thing instead of recreating the whole input display.

nezed:
@JemarJones Sounds good, but have not ability to work with date format (e.g. append label with weekday).
Can you copy and post your proposition described above into #338? (It will be great to discuss and improve solution of exiting PR!)

Clcll:
I resolve it this way
`
                        <DateRangePicker
                            startDate={this.state.startDate} // momentPropTypes.momentObj or null,
                            endDate={this.state.endDate} // momentPropTypes.momentObj or null,
                            onDatesChange={({ startDate, endDate }) => this.setState({ startDate, endDate })} 
[...]

PR #338, opened 2017-02-20
Custom input text and calendar caption renders

### LINKED ISSUE #331: Custom input text and calendar caption
Hi guys!
We are planning to create a PR which introduces two new props:
```js
// Returns text that will be displayed inside .DateInput__display-text
renderInputText: moment => string | React.element<*>,
// Returns element that will be added at bottom of CalendarMonthGrid (in both cases, vertical and horizontal)
renderCalendarCaption: () => React.element<*>,
```
May it be interesting and useful?

nezed:
Features introduced:

- Custom input text renderer:
 <br/>
- Calendar caption renderer:

These props are universal, and available in `DateRangePicker`, `SingleDatePicker` and `DayPicker`.
This PR resolves #331 

/сс @majapw @ljharb


ljharb:
Thanks for the PR, but since we still haven't decided nor indicated that this is a feature we want in #331, this work might have been premature.

Separately, I think the "caption" feature and the "custom input text" feature need to be considered separately and in separate PRs.

nezed:
@ljharb Thanks for your attention!
As you proposed, i have moved "caption" feature into separate PR (#341).
Unfortunately i have not seed any members activity in base issue #331.
I hope that your team may decide this features useful or may propose changes, because this features are very important for our team and required for using `react-dates` on our projects.

JemarJones:
I think `renderInputText` may be an inappropriate name for a prop that even in your example is used to render something that isn't simply text

nezed:
@JemarJones It makes sense, but `renderInputValue` may cause confusion, It could be perceived as input value.
We also have internal `inputValue` prop, and other entry with same name may cause the erroneous association.

JemarJones:
Mentioned this in #278 but i suppose its more relevant here 

> Those are good points. What if alternatively, a prop was very similar to #338 was added called something like wrapInput that received the node (i haven't checked which node that is) that currently renders the input as an argument, and could simply use that as a child if it chooses? I think that could be a nice solution because it makes things like this issue a case of adding only a small thing instead of recreating the whole input display.

iamandrewluca:
Any updates?

nezed:
@iamandrewluca We still using `@kupibilet/react-dates` fork for this functionality

PR #341, opened 2017-02-21
Calendar caption with renderCalendarCaption

nezed:
#### Usage example:
```js
  const CustomInfo = ({day, format}) => (
    <div>
      ❕ Some useful info here
    </div>
  );

  <SingleDatePickerWrapper
    renderCalendarInfo={() => (
      <CustomInfo />
    )}
  />
```
#### Example screenshot:

The `renderCalendarInfo` prop are universal, and available in `DateRangePicker`, `SingleDatePicker` and `DayPicker`.

> As a part of #331

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,7 +138,7 @@ By default, we do not show days from the previous month and the next month in th
 
 Optionally you can provide calendar caption block renderer.
 ```js
-   renderCalendarCaption: PropTypes.func,
+   renderCalendarInfo: PropTypes.func,
 ```
 
 **DayPicker presentation:**
--- a/README.md
+++ b/README.md
@@ -136,6 +136,11 @@ By default, we do not show days from the previous month and the next month in th
    initialVisibleMonth: PropTypes.func,
 ```
 
+Optionally you can provide calendar caption block renderer.
+```js
+   renderCalendarCaption: PropTypes.func,
+```
+
 **DayPicker presentation:**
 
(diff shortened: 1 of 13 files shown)


README.md
```diff
@@ -136,6 +136,11 @@ By default, we do not show days from the previous month and the next month in th
    initialVisibleMonth: PropTypes.func,
 ```
 
+Optionally you can provide calendar caption block renderer.
+```js
+   renderCalendarCaption: PropTypes.func,
```
majapw: I'm concerned with the nomenclature of using `caption` here as we already have a `<caption>` dom element on each table inside each `CalendarMonth` component. I think using `Calendar` is also somewhat problematic as this is rendered on the `DayPicker` level not on the `CalendarMonth` level. Perhaps renaming it to `renderDayPickerCaption` or `renderDayPickerAdditionalInfo` might be a good idea? What do you think?

majapw: Oh this is totes my fault by having the `Calendar props` story isn't it. What if we called this `renderCalendarAdditionalInfo` or something? Maybe even just `renderAdditionalInfo`? God, I'm verbose.

test/components/DayPicker_spec.jsx
```diff
@@ -64,6 +64,15 @@ describe('DayPicker', () => {
       });
     });
 
+    describe('renderCaption', () => {
+      it('caption exists', () => {
+        const testCaptionClass = 'test-cation-container';
```
majapw: nit: `test-caption-container`

README.md
nezed: Glad to see your response!
Maybe its better to this props `renderCalendarInfo`?

majapw: Yes. Let's do that instead!

PR #351, opened 2017-03-02
Moves phrases out into one shared defaultPhrases file

majapw:
to: @ljharb @airbnb/webinfra 

This is to split up some of the code in https://github.com/airbnb/react-dates/pull/301, which I admit, got away from me somewhat.

### DIFF
--- a/.istanbul.yml
+++ b/.istanbul.yml
@@ -7,6 +7,7 @@ instrumentation:
     "public/**",
     "test/**",
     "coverage/**",
+    "src/defaultPhrases.js"
   ]
 check:
   global:
--- a/src/components/DateRangePickerInput.jsx
+++ b/src/components/DateRangePickerInput.jsx
@@ -2,6 +2,9 @@ import React, { PropTypes } from 'react';
 import { forbidExtraProps } from 'airbnb-prop-types';
 import cx from 'classnames';
 
+import { DateRangePickerInputPhrases } from '../defaultPhrases';
+import getPhrasePropTypes from '../utils/getPhrasePropTypes';
+
 import DateInput from './DateInput';
 import RightArrow from '../svg/arrow-right.svg';
 import CloseButton from '../svg/close.svg';
@@ -41,9 +44,7 @@ const propTypes = forbidExtraProps({
   customArrowIcon: PropTypes.node,
 
   // i18n
-  phrases: PropTypes.shape({
-    clearDates: PropTypes.node,
-  }),
+  phrases: PropTypes.shape(getPhrasePropTypes(DateRangePickerInputPhrases)),
 });
 
 const defaultProps = {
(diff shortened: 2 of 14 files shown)


.istanbul.yml
```diff
@@ -7,6 +7,7 @@ instrumentation:
     "public/**",
     "test/**",
     "coverage/**",
+    "src/defaultPhrases.js"
```
ljharb: trailing comma?

ljharb: also why does this need to be ignored, if it's exercised on every import?

src/components/DateRangePickerInput.jsx
```diff
@@ -145,13 +144,16 @@ export default class DateRangePickerInput extends React.Component {
         })}
       >
         {(showDefaultInputIcon || customInputIcon !== null) &&
-          <span
+          <button
```
ljharb: was this change meant to be included in this PR?

src/components/DayPickerNavigation.jsx
```diff
@@ -80,6 +91,7 @@ export default function DayPickerNavigation(props) {
       }
 
       <span
+        aria-label={phrases.jumpToNextMonth}
```
ljharb: ?

src/components/DayPickerNavigation.jsx
```diff
@@ -72,6 +82,7 @@ export default function DayPickerNavigation(props) {
     <div className={navClassNames}>
       {!isVerticalScrollable &&
         <span
+          aria-label={phrases.jumpToPrevMonth}
```
ljharb: ?

src/components/SingleDatePickerInput.jsx
```diff
@@ -122,13 +121,11 @@ export default class SingleDatePickerInput extends React.Component {
               'SingleDatePickerInput__clear-date--hide': !displayValue,
               'SingleDatePickerInput__clear-date--hover': isClearDateHovered,
             })}
+            aria-label={phrases.clearDate}
```
ljharb: ?

.istanbul.yml
majapw: Fair. I was thinking that we would ignore because writing tests for the file itself seems v. silly

src/components/DateRangePickerInput.jsx
majapw: Yes. I figured we'd do some of the small aria-labeling/add those phrases unrelated to keyboard nav here.

src/components/DayPickerNavigation.jsx
majapw: Yes. I figured we'd do some of the small aria-labeling/add those phrases unrelated to keyboard nav here.

src/components/DayPickerNavigation.jsx
majapw: Yes. I figured we'd do some of the small aria-labeling/add those phrases unrelated to keyboard nav here.

src/components/SingleDatePickerInput.jsx
majapw: Yes. I figured we'd do some of the small aria-labeling/add those phrases unrelated to keyboard nav here.

.istanbul.yml
ljharb: It's useful to ensure that it can be imported without an error :-)

PR #352, opened 2017-03-03
Remove override of day picker width

jerrylow:
The `$react-dates-width-day-picker: 300px;` defined in `CalendarMonthGrid.scss` overrides any custom variables provided.


majapw:
Thank you! This was def a bug leftover.

PR #356, opened 2017-03-06
Add customCloseIcon prop

moonboots:
@majapw @ljharb 

Adds a `customCloseIcon` prop to allow overriding the default `<CloseButton />`. I matched the behavior of `customArrowIcon`: default prop of null, fallback to original component inside `render`, and a wrapping css class.

### DIFF
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -50,6 +50,7 @@ const defaultProps = {
   showDefaultInputIcon: false,
   customInputIcon: null,
   customArrowIcon: null,
+  customCloseIcon: null,
 
   // calendar presentation and interaction related props
   orientation: HORIZONTAL_ORIENTATION,
@@ -222,6 +223,7 @@ export default class DateRangePicker extends React.Component {
       keepOpenOnDateSelect,
       renderDay,
       initialVisibleMonth,
+      customCloseIcon,
     } = this.props;
     const { dayPickerContainerStyles } = this.state;
 
@@ -230,6 +232,8 @@ export default class DateRangePicker extends React.Component {
       : undefined;
     const initialVisibleMonthThunk =
       initialVisibleMonth || (() => (startDate || endDate || moment()));
+
+    const closeIcon = customCloseIcon || (<CloseButton />);
 
     return (
       <div
@@ -272,7 +276,9 @@ export default class DateRangePicker extends React.Component {
             <span className="screen-reader-only">
               {this.props.phrases.closeDatePicker}
             </span>
-            <CloseButton />
+            <div className="DateRangePicker__close">
+              {closeIcon}
+            </div>
           </button>
         }
       </div>
@@ -293,6 +299,7 @@ export default class DateRangePicker extends React.Component {
       showDefaultInputIcon,
       customInputIcon,
       customArrowIcon,
+      customCloseIcon,
       disabled,
       required,
       phrases,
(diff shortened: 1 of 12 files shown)


src/components/DateRangePicker.jsx
```diff
@@ -273,7 +277,9 @@ export default class DateRangePicker extends React.Component {
             <span className="screen-reader-only">
               {this.props.phrases.closeDatePicker}
             </span>
-            <CloseButton />
+            <div className="DateRangePickerInput__close">
```
majapw: This class name is wrong. You're not inside the input.

stories/DateRangePicker_input.js
```diff
@@ -30,6 +30,19 @@ const TestCustomArrowIcon = () => (
...
+const TestCustomCloseIcon = () => (
+  <span
+    style={{
+      border: '1px solid #dce0e0',
+      backgroundColor: '#fff',
+      color: '#484848',
+      padding: '3px',
+    }}
+  >
+    {'X'}
```
majapw: This can just be `X` no?

test/components/DateRangePickerInput_spec.jsx
```diff
@@ -117,6 +117,17 @@ describe('DateRangePickerInput', () => {
...
   });
 
+  describe('props.customCloseIcon is a React Element', () => {
+    it('has custom icon', () => {
+      const wrapper = shallow(
+        <DateRangePickerInput
+          showClearDates
+          customCloseIcon={<span className="custom-close-icon" />}
+        />);
+      expect(wrapper.find('.DateRangePickerInput__calendar-icon .custom-close-icon'));
```
majapw: This should fail... why would it be under `.DateRangePickerInput__calendar-icon`? Why isn't this failing?

moonboots:
@majapw I've added a corresponding spec for SingleDatePicker. The spec you pointed out wasn't failing because it accidentally didn't have any assertions, only a dangling `expect(..);`, fixed now.

test/components/SingleDatePickerInput_spec.jsx
```diff
@@ -54,6 +54,17 @@ describe('SingleDatePickerInput', () => {
...
     });
+
+    describe('props.customCloseIcon is a React Element', () => {
+      it('has custom icon', () => {
+        const wrapper = shallow(
+          <SingleDatePickerInput
+            showClearDates
+            customCloseIcon={<span className="custom-close-icon" />}
+          />);
+        expect(wrapper.find('.DateRangePickerInput__calendar-icon .custom-close-icon'));
```
majapw: You now have the same problem here in this spec...

moonboots: Oops, fixed

zeemawn:
Correct me if I'm wrong, but shouldn't this whole thing rather be called "custom CLEAR icon"?
The 'X' might look like a close-icon, but its function is to clear the dates, isn't it?
(Besides that, it would be nice to have an actual close-button)

ljharb:
It closes the datepicker, it doesn't clear the dates.

majapw:
It actually does both... We overload the icon in use, so uh, I get the Q

ljharb:
Lol well there you go

ljharb:
@majapw can we unoverload them?

interactivejohn:
I ran into the same issue as @zeemawn. I spent a good couple hours thinking that I could add a custom close icon to the default state, only to find out that it was only implemented for the `showClearDate` and `withFullScreenPortal` props. Maybe something can be added to the README that explains where the `customCloseIcon` is implemented? Meanwhile I will fork this package and start working on implementing a close functionality for the default `DatePicker` as requested above (I also need it for a project).

PR #367, opened 2017-03-09
Add missing customCloseIcon propTypes

moonboots:
@majapw 

I forgot a few propTypes in my last PR, which is triggering some extra prop warnings.


majapw:
Jack

Issue #363, opened 2017-03-08
Feature: prop `firstDayOfTheWeek`


daltones:
This is related to #60. That issue was closed with a `wontfix` tag. It was concluded that the issue is a matter of localization and is already handled by Moment.

I am proposing here a way to force a day to be the first of the week _to be displayed_.

I have a use case for `DayPicker` that the user selects only Mondays, for real events that need to be from Monday to Sunday. It's very important for the user experience that this _specific_ calendar shows Monday as the first day.

I have to emphasize that this is not a localization issue. It's an specific type of event that needs an specific interface.

Actually this same app has other parts that require simple date selection. And in these cases it's expected that the calendar normally shows Sunday as first day. 

I can't mess up with Moment's global locale and all date operations based on weeks. So that would be a feature for `react-dates`.

If we all agree on this, I'd be happy to try a PR.

ljharb:
I think this is a valid use case. The concern I have is that all of our *internal* moment week calculations might be off if we start the week on a different day than the locale does.

I can't promise we'd accept it, of course, but a proof of concept PR for all the primitive components might be very instructive.

sfernandezsl:
Any update on this?

daltones:
Hi @ljharb, sorry for not leaving a comment before.

This issue was closed by https://github.com/react-dates/react-dates/pull/371.

ljharb:
aha, thanks :-)

PR #371, opened 2017-03-13
Feature: add firstDayOfWeek prop

daltones:
This PR tries to provide a solution to #363.

We urgently needed this feature in our app. Right now a forked `react-dates` with this implementation is running in production, all good. We use just `<DayPicker />` components in two different pages.

~For now it's missing passing along prop `firstDayOfWeek` from `DateRangePicker` and `SingleDatePicker` and write some new tests for this functionality.~

The big change here is the [almost] rewrite of `getCalendarMonthWeeks`. The logic now is that every operation involving weekdays needs the `firstDayOfWeek` parameter (0-Sun...6-Sat locale independent). And then `firstDayOfWeek` prop from components gets the default value according to Moment's global locale (so we're not breaking anything).

I'd be happy if I receive feedback of this implementation. I'll update the PR very soon with the rest.


daltones:
Oh, I forgot to mention. This implementation also fixes a bug in an edge case I found:

```jsx
// We have Moment's global locale starting the week on Sundays
moment.locale('some-sunday-based-locale');

// For some reason in initialVisibleMonth we return a moment object set
// with another locale that starts the week on Mondays
<DayPicker initialVisibleMonth={() => moment().locale('some-monday-based-locale')} />
```

In `react-dates` v8.1.1 we'll have a calendar that the shows weekdays header starting on Monday, but the actual days arranged as starting on Sunday.

src/components/CalendarMonth.jsx
```diff
@@ -44,6 +45,7 @@ const defaultProps = {
   onDayMouseEnter() {},
   onDayMouseLeave() {},
   renderDay: null,
+  firstDayOfWeek: moment.localeData().firstDayOfWeek(),
```
ljharb: this will freeze the default at parse time (and the locale could be changed at any time) - i think this default needs to be calculated in the render path, and have it default to `-1` or `null` here.

src/components/CalendarMonthGrid.jsx
```diff
@@ -52,6 +53,7 @@ const defaultProps = {
   onMonthTransitionEnd() {},
   renderDay: null,
   transformValue: 'none',
+  firstDayOfWeek: moment.localeData().firstDayOfWeek(),
```
ljharb: same here

src/components/DayPicker.jsx
```diff
@@ -65,6 +66,7 @@ const defaultProps = {
   onOutsideClick() {},
   hidden: false,
   initialVisibleMonth: () => moment(),
+  firstDayOfWeek: moment.localeData().firstDayOfWeek(),
```
ljharb: and here

src/utils/getCalendarMonthWeeks.js
```diff
@@ -1,39 +1,39 @@
-export default function getCalendarMonthWeeks(month, enableOutsideDays) {
+import moment from 'moment';
+
+export default function getCalendarMonthWeeks(
+  month,
+  enableOutsideDays,
+  firstDayOfWeek = moment.localeData().firstDayOfWeek(),
```
ljharb: can something validate that this value is an integer 0-6, and throw a TypeError if not?

daltones: I thought about that, but I didn't see any other example of this kind of validation. Even here we call `month.clone()` without knowing if `month` is in fact a moment object. Moreover, like all these parameters of `utils` functions, values came from component props that already pass by some `propTypes` validation in dev.

ljharb: The propType should ensure that. I think that in general the utils functions should not assume safe inputs.

ljharb:
@daltones would you please rebase this branch to ensure there's no merge commits in it?

daltones:
@ljharb done!

src/components/CalendarMonth.jsx
```diff
@@ -52,16 +54,27 @@ const defaultProps = {
 export default class CalendarMonth extends React.Component {
   constructor(props) {
     super(props);
+
     this.state = {
-      weeks: getCalendarMonthWeeks(props.month, props.enableOutsideDays),
+      weeks: getCalendarMonthWeeks(
+        props.month,
+        props.enableOutsideDays,
+        props.firstDayOfWeek === null ? moment.localeData().firstDayOfWeek() : props.firstDayOfWeek,
```
ljharb: This should probably use `== null` to avoid distinguishing from undefined

src/components/CalendarMonth.jsx
```diff
@@ -52,16 +54,27 @@ const defaultProps = {
...
+    const { month, enableOutsideDays, firstDayOfWeek } = nextProps;
+    if (!month.isSame(this.props.month)
+        || enableOutsideDays !== this.props.enableOutsideDays
+        || firstDayOfWeek !== this.props.firstDayOfWeek) {
       this.setState({
-        weeks: getCalendarMonthWeeks(month, enableOutsideDays),
+        weeks: getCalendarMonthWeeks(
+          month,
+          enableOutsideDays,
+          firstDayOfWeek === null ? moment.localeData().firstDayOfWeek() : firstDayOfWeek,
```
ljharb: Also here

src/components/DayPicker.jsx
```diff
@@ -350,11 +352,16 @@ export default class DayPicker extends React.Component {
 
     const style = this.isHorizontal() ? horizontalStyle : {};
 
+    let { firstDayOfWeek } = this.props;
+    if (firstDayOfWeek === null) {
```
ljharb: And here

src/components/DayPickerRangeController.jsx
```diff
@@ -52,6 +52,7 @@ const propTypes = forbidExtraProps({
   onOutsideClick: PropTypes.func,
   renderDay: PropTypes.func,
   renderCalendarInfo: PropTypes.func,
+  firstDayOfWeek: PropTypes.oneOf([0, 1, 2, 3, 4, 5, 6]),
```
ljharb: I think this propType should be shared and exported up from the components that need it - ie, this component shouldn't du-locate the definition, it should just import the propType from DayPicker and delegate the authority to it.

Same applies to any component where the prop is merely proxied down.

Effectively I'd prefer the oneOf to only appear in one place.

src/utils/getCalendarMonthWeeks.js
```diff
@@ -1,39 +1,48 @@
...
-  // set utc offset to get correct dates in future (when timezone changes)
-  const baseDate = month.clone();
-  const firstOfMonth = baseDate.clone().startOf('month').hour(12);
-  const lastOfMonth = baseDate.clone().endOf('month').hour(12);
+import moment from 'moment';
 
-  const currentDay = firstOfMonth.clone();
-  let currentWeek = [];
-  const weeksInMonth = [];
+const WEEKDAYS = [0, 1, 2, 3, 4, 5, 6];
```
ljharb: Could this array be shared, and used inside the oneOf?

src/components/CalendarMonth.jsx
ljharb: I do think you're right; that a value of `undefined` would trigger the default, and a value of `null` would not, so === null would work fine - but in general for style, yes. Not a huge deal, ofc.

CarlRosell:
What's the current status of this PR? 
I'm glad to see someone else also having this issue. I've made the same changes for other "date-picker"-libraries as this is a major blocker for the company where I work. 👍

majapw:
Hi @CarlRosell, I can take a look at this tomorrow/early next week.

daspete:
Hi there, is this feature in the todo pipeline? It would be great to see this one working, because our company need to set the first day to monday.... or is there another way to set this?

majapw:
@ljharb since I updated this PR with my own changes/rebase/etc. can you please sanity-check it for me before I merge it in?

test/utils/getCalendarMonthWeeks_spec.js
```diff
@@ -157,4 +187,86 @@ describe('getCalendarMonthWeeks', () => {
...
+      it('month ends at [n][2] if last day is Tuesday', () => {
+        const weeksInJanuary2017 = getCalendarMonthWeeks(january2017Start);
+        const lastDayOfMonth = weeksInJanuary2017[weeksInJanuary2017.length - 1][2];
+        const rightPosition = january2017End.isSame(lastDayOfMonth, 'day');
+        expect(rightPosition).to.equal(true);
+      });
+    });
+
+    describe('locale with Monday as first day of week', () => {
+      before(() => {
```
ljharb: we need to use `beforeEach` here; `before` is beforeAll, and we don't want that.

test/utils/getCalendarMonthWeeks_spec.js
```diff
@@ -157,4 +187,86 @@ describe('getCalendarMonthWeeks', () => {
...
+  });
+
+  describe('firstDayOfWeek default value locale-aware', () => {
+    // using these known dates to see if they appear at the right position of the calendar
+    // trying different firstDayOfWeek values
+    const january2017Start = moment('2017-01-01'); // Sunday
+    const january2017End = january2017Start.clone().endOf('month'); // Tuesday
+
+    describe('locale with Sunday as first day of week', () => {
+      before(() => {
```
ljharb: we need to use `beforeEach` here; `before` is beforeAll, and we don't want that.

majapw:
@ljharb addressed your comments!

src/shapes/DayOfWeekShape.js
```diff
@@ -0,0 +1,5 @@
+import { PropTypes } from 'react';
```
ljharb: This needs to use the prop-types package

majapw:
Naw this is good! You're catching all the things I've missed. :) I addressed the prop types issue @ljharb, can you take another look?

CarlRosell:
🎉

cesalberca:
Nice! Just in time for a component I'm doing in spanish ^^ Any ideas when this patch will be released?

majapw:
Should be live in v12.3.0! :)

PR #377, opened 2017-03-17
Disable fastclick

moonboots:
to: @majapw 

This is kind of a terrible way to work around a fastclick issue affecting our implementation. Let me know if you think it would be less terrible as a prop.

### DIFF
--- a/src/components/DateInput.jsx
+++ b/src/components/DateInput.jsx
@@ -122,7 +122,7 @@ export default class DateInput extends React.Component {
       >
         <input
           aria-label={placeholder}
-          className="DateInput__input"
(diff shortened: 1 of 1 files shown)


majapw:
@moonboots is this adding a class that does something base on some CSS existing ... in our codebase? Why not add the actual CSS here? This seems v. strange.

moonboots:
Unfortunately, [disabling fastclick on parts of the page requires adding this `needsclick` class](https://github.com/ftlabs/fastclick/blob/3db9f899c25b7b2e1517dc5cc17494ec9094bc43/lib/fastclick.js#L253). We don't reference this class in css.

Issue #378, opened 2017-03-17
Weird border overlapping behavior when hovering a calendar day within a range


ConAntonakos:
First of all, I want to greatly thank all contributors for this beautiful component. 👏 

Problem
---
When playing around with the colors, I noticed this odd quirk:

![screen shot 2017-03-17 at 1 11 00 pm](https://cloud.githubusercontent.com/assets/1504302/24054683/7061d074-0b13-11e7-8e89-bc03c94a4a70.png)
_Taken from Airbnb.com_

Apparently, hovering a calendar day adjacent to other colored calendar days doesn't prioritize changing the `border-color` for the hovered element.
___
What would be the best way to address this? Is this a limitation of using `<table>` and `<td>` HTML elements?

I'm happy to contribute to this if it is indeed deemed an unintended result. Thanks!

majapw:
Hi! This is definitely unintentional but has been fairly low priority recently and thus been somewhat ignored!

I think it is related to the collapsed border of the `<table>`/`<td>` but think it would be possible to address with a change to the specificity of the CSS classes. If you'd be willing to explore this and try to take it on, I'd be super excited to accept a PR for the fix! :) I'll assign it to you for now, and if you don't have the time, feel free to pass it back to me. ✌️

ConAntonakos:
Awesome! Thank you, @majapw. I'm very happy to help; the least I can do.

I had a very small amount of time to play around with it, but what is the intended scope, so I can be consistent with your design decisions? Do you still want the same border width or same styling as it is currently? For example, I played around with `border-collapse: separate` in `.CalendarMonth table`, but then you get this unevenness in border widths when comparing `td:first-child` & `td:last-child` to `<td>`s in the middle.

Open to suggestions! 😃

PR #380, opened 2017-03-17
Fix weird border styling behavior for hover date (#378)

ConAntonakos:
My solution for #378! I'm still noticing this weird curviness to the top-<left|right> border styles. I'm not sure how to solve that.

Let me know what you think. Thanks!

### DIFF
--- a/css/CalendarDay.scss
+++ b/css/CalendarDay.scss
@@ -62,6 +62,22 @@
     border: 1px double $react-dates-color-primary;
   }
 
+  &.CalendarDay--hovered-previous-day {
+    border-right: 1px double $react-dates-color-primary;
+  }
+
+  &.CalendarDay--hovered-next-day {
+    border-left: 1px double $react-dates-color-primary;
+  }
+
+  &.CalendarDay--hovered-previous-week {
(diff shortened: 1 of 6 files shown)


majapw:
Sorry for taking so long to get to this! I think a problem is that in adding these modifiers (to be fair, this is largely due to the way we use modifiers right now), we're going to take a performance hit. It would be cool if it were possible to do this mostly in CSS. :/ I will investigate it a bit more.

ConAntonakos:
No problem, @majapw! I think I agree with you regarding perf, though I haven't tested it thoroughly. I simply wanted to get this solution out at least since it was the first one that came to mind. I also agree about possibly changing this via styling only. Do you think it's possible with the current `<table> <td>` implementation? Do you think that might have to change? I'd be more than happy to rewrite the PR.

majapw:
This still needs to be addressed, but because tables are hard I'm going to close it for now because I do not believe this is the correct solution.

ConAntonakos:
No worries. Thanks for checking in. I'd be happy to update the PR once a better solution is figured out.

PR #387, opened 2017-03-20
Simplify the README

majapw:
Describing each prop in detail feels unwieldy, especially since it is ultimately better to just go to the storybook and see what the heck they actually do. It also fails to adequately explain what you actually need to get a functional datepicker. In this change, I try to focus on the required props and point the viewer to the storybook for the optional props.

To view the change, I would recommend ignoring the diff and just looking at the new README on its own.

to: @airbnb/webinfra

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ To run that demo on your own computer:
 * Visit http://localhost:9001/
 
 ## Getting Started
-#### Install dependencies
+### Install dependencies
 Ensure packages are installed with correct version numbers by running:
   ```sh
   (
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -40,494 +40,196 @@ Ensure packages are installed with correct version numbers by running:
...
+import { DateRangePicker, SingleDatePicker, DayPickerRangeController } from 'react-dates';
 ```
 
-#### Include CSS
+### Include CSS
 When you use Webpack with CSS loader:
 ```js
 import 'react-dates/lib/css/_datepicker.css';
 ```
 Otherwise create a CSS file with the contents of `require.resolve('react-dates/lib/css/_datepicker.css')` and include it in the html `<head>` section.
```
raejin: Maybe specify the css file needs to be compiled? Also I'd like to suggest separting the different usages using bullet points. (webpack vs everything else)

README.md
```diff
@@ -40,494 +40,196 @@ Ensure packages are installed with correct version numbers by running:
...
-  endDate: momentPropTypes.momentObj,
-```
-
-`onDatesChange` is the callback necessary to update the date state being held in the parent component and pass that back down to the `DateRangePicker` as props. `onDatesChange` receives an object of the form
-`{ startDate: momentPropTypes.momentObj, endDate: momentPropTypes.momentObj }`
-as an argument.
-```js
-  onDatesChange: PropTypes.func,
-```
+We provide a handful of components for your use. Each of these have a handful of props required for a fully functional interactive date picker as well as a large number of props for customizing the look and feel of the inputs, calendar, and
```
raejin: > Each of these have a handful of props required for a fully functional interactive date picker as well as a large number of props for customizing the look and feel of the inputs, calendar, and even individual days. 

It feels like a very long sentence with little added information to me (Disclaimer: I'm a really bad & lazy reader with longer sentence). Breaking up the sentences a little bit may help with readability, and make it relatable to the user. 
i.e. 
> If you supply essential props with each component, you'll get a full featured interactive date picker. With additional optional props, you can customize the look and feel of the inputs, calendar, etc..

README.md
```diff
@@ -40,494 +40,196 @@ Ensure packages are installed with correct version numbers by running:
...
+even individual days. You can see what each of the props do in the [live demo](http://airbnb.io/react-dates/) or explore
+how to properly wrap the pickers in the [examples folder](https://github.com/airbnb/react-dates/tree/master/examples).
 
-**Focus:**
+#### DateRangePicker
+The `DateRangePicker` is a fully controlled component that allows users to select a date range. You can control the selected
+dates using the `startDate`, `endDate`, and `onDatesChange` props as shown below. The `DateRangePicker` also manages internal
+state for partial dates entered by typing (although `onDatesChange` will not trigger until a date has been entered
+completely in that case). Similarly, you can control which input is focused as well as calendar visibility (the calendar is
+only visible if `focusedInput` is defined) with the `focusedInput` and `onFocusChange` props as shown below.
```
Reaverart: @majapw yes, thank you, now its much clearer to understand what it stands for 👍

Reaverart:
Thank you for this update, looks it will be clearer to understand minimal setup, but I cant agree that storybook is replacement of a good readme, its more like addition. 
The main reason I was messed up is still here: storybook uses custom `DateRangePickerWrapper` from examples folder which isnt in lib export, so first time I thought that storybook is outdated and there was breaking changes.. And there is still no connection between samples in readme and storybook. Just compare minimal setup from storybook:
```jsx
<DateRangePickerWrapper/>
```
vs real configuration:
```jsx
// at first write your own HOC to manage focusedInput, then your render function will be
render() {
  <DateRangePicker
    endDate={this.state.endDate}
    onDatesChange={({ startDate, endDate }) => this.setState({ startDate, endDate })}
    onFocusChange={focusedInput => this.setState({ focusedInput })}
  />
}
```
So probably at least notice on top of storybook might help others. Something like "All examples built around custom wrapper which is not in export, lookup for minimal setup in readme".

README.md
majapw: What do you mean by needs to be compiled? `react-dates/lib/css/_datepicker.css` has already been run through a SASS compiler.

README.md
majapw: Seems legit. I'll make the change.

README.md
```diff
@@ -40,494 +40,196 @@ Ensure packages are installed with correct version numbers by running:
...
+<DateRangePicker
+  startDate={this.state.startDate} // momentPropTypes.momentObj or null,
+  endDate={this.state.endDate} // momentPropTypes.momentObj or null,
+  onDatesChange={({ startDate, endDate }) => this.setState({ startDate, endDate })} // PropTypes.func.isRequired,
+  focusedInput={this.state.focusedInput} // PropTypes.oneOf([START_DATE, END_DATE]) or null,
+  onFocusChange={focusedInput => this.setState({ focusedInput })} // PropTypes.func.isRequired,
+/>
 ```
 
-`isOutsideRange` indicates which days are out of selectable range.
```
Reaverart: Just to notice, I think better to not cut existing docs unless storybook lacks in some parts. for instance in storybook not clear what this function should return

majapw: I think it'd be better to add this information to the storybook rather than leave the readme as noisy and out of date as it is right now. You can see the source of each story in the info (?) tab, which should ameliorate this somewhat.

Issue #264, opened 2017-01-25
[Feature Suggestion] - Add onClose callback


BrunoQuaresma:
Add a hook that is called when the picker is close.

majapw:
Out in v10.1.0

BrunoQuaresma:
Thanks!

PR #397, opened 2017-03-23
[Feature] - Add onClose callback

rafacianci:
This PR is related to #264 

We are having some troubles when the user close the picker on the `<DateRangePicker />` component, because it calls the `onDatesChange` and the `onFocusChange` at the same time, and when the `onFocusChange` is called without the focus there wasn't time to reset the state.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,6 +111,7 @@ navPrev: PropTypes.node,
 navNext: PropTypes.node,
 onPrevMonthClick: PropTypes.func,
 onNextMonthClick: PropTypes.func,
+onClose: PropTypes.func,
 
 // day presentation and interaction related props
 renderDay: PropTypes.func,
@@ -171,6 +172,7 @@ navPrev: PropTypes.node,
 navNext: PropTypes.node,
 onPrevMonthClick: PropTypes.func,
 onNextMonthClick: PropTypes.func,
+onClose: PropTypes.func,
 
 // day presentation and interaction related props
 renderDay: PropTypes.func,
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -74,6 +74,8 @@ const defaultProps = {
   onPrevMonthClick() {},
   onNextMonthClick() {},
 
+  onClose() {},
+
   // day presentation and interaction related props
   renderDay: null,
   minimumNights: 1,
@@ -143,7 +145,7 @@ export default class DateRangePicker extends React.Component {
   }
 
   onOutsideClick() {
-    const { onFocusChange } = this.props;
+    const { onFocusChange, onClose, startDate, endDate } = this.props;
     if (!this.isOpened()) return;
 
     this.setState({
@@ -153,6 +155,7 @@ export default class DateRangePicker extends React.Component {
     });
 
     onFocusChange(null);
+    onClose({ startDate, endDate });
   }
 
   onDateRangePickerInputFocus(focusedInput) {
@@ -296,6 +299,7 @@ export default class DateRangePicker extends React.Component {
       renderCalendarInfo,
       initialVisibleMonth,
       customCloseIcon,
+      onClose,
       phrases,
     } = this.props;
     const { dayPickerContainerStyles, isDayPickerFocused, showKeyboardShortcuts } = this.state;
@@ -324,6 +328,7 @@ export default class DateRangePicker extends React.Component {
           onNextMonthClick={onNextMonthClick}
           onDatesChange={onDatesChange}
           onFocusChange={onFocusChange}
+          onClose={onClose}
           focusedInput={focusedInput}
           startDate={startDate}
           endDate={endDate}
@@ -387,6 +392,7 @@ export default class DateRangePicker extends React.Component {
       reopenPickerOnClearDates,
       keepOpenOnDateSelect,
       onDatesChange,
+      onClose,
     } = this.props;
 
     const { isDateRangePickerInputFocused } = this.state;
@@ -422,6 +428,7 @@ export default class DateRangePicker extends React.Component {
             onFocusChange={this.onDateRangePickerInputFocus}
             onArrowDown={this.onDayPickerFocus}
             onQuestionMark={this.showKeyboardShortcutsPanel}
+            onClose={onClose}
             phrases={phrases}
             screenReaderMessage={screenReaderInputMessage}
             isFocused={isDateRangePickerInputFocused}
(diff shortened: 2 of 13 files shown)


test/components/DateRangePickerInputController_spec.jsx
```diff
@@ -90,6 +90,25 @@ describe('DateRangePickerInputController', () => {
       wrapper.instance().onClearFocus();
       expect(onFocusChangeStub.calledWith(null)).to.equal(true);
     });
+
+    it('calls props.onFocusChange with startDate and endDate args', () => {
```
viniciusdacal: I think here it was supposed to be `props.onClose`

README.md
```diff
@@ -171,6 +172,7 @@ navPrev: PropTypes.node,
 navNext: PropTypes.node,
 onPrevMonthClick: PropTypes.func,
 onNextMonthClick: PropTypes.func,
+onClose: ({ startDate, endDate }) => this.setState({ startDate, endDate }) // PropTypes.func,
```
viniciusdacal: shouldn't it  be `onclose: PropTypes.func` instead?

README.md
```diff
@@ -111,6 +111,7 @@ navPrev: PropTypes.node,
...
```
viniciusdacal: do we need the comment here and? Same case line `175`

src/components/DateRangePicker.jsx
```diff
@@ -370,6 +375,8 @@ export default class DateRangePicker extends React.Component {
       reopenPickerOnClearDates,
       keepOpenOnDateSelect,
       onDatesChange,
+      onFocusChange,
```
viniciusdacal: Are we using onFocusChange in this scope?

README.md
rafacianci: I put the function to let the user know how to use the callback... I didn't what would be the better way to show it. Any ideias?

ljharb: English text in the documentation. propTypes are not for a human, they're for React to issue warnings when it doesn't match.

viniciusdacal: I think we could set examples above, after line `77`

viniciusdacal: And this one, would go after line `142`

rafacianci:
@ljharb and @majapw I updated the Pull Request, can you review it again? please...
thanks.

src/components/DateRangePickerInputController.jsx
```diff
@@ -123,7 +129,10 @@ export default class DateRangePickerInputController extends React.Component {
       !isInclusivelyBeforeDay(endDate, startDate);
     if (isEndDateValid) {
       onDatesChange({ startDate, endDate });
-      if (!keepOpenOnDateSelect) onFocusChange(null);
+      if (!keepOpenOnDateSelect) {
+        onFocusChange(null);
```
ljharb: Maybe this should call `this.onClearFocus` instead? (To avoid repeating this pair of statements)

src/components/DayPickerRangeController.jsx
```diff
@@ -147,7 +149,10 @@ export default class DayPickerRangeController extends React.Component {
         this.props.onFocusChange(START_DATE);
       } else if (isInclusivelyAfterDay(day, firstAllowedEndDate)) {
         endDate = day;
-        if (!keepOpenOnDateSelect) this.props.onFocusChange(null);
+        if (!keepOpenOnDateSelect) {
+          this.props.onFocusChange(null);
```
ljharb: Let's destructure out these prop funcs so their `this` isn't the props object

src/components/DateRangePickerInputController.jsx
rafacianci: nice.. :D

src/components/DayPickerRangeController.jsx
rafacianci: ok.. 👍

rafacianci:
done @ljharb.. can you review it?

src/components/DayPickerRangeController.jsx
```diff
@@ -144,10 +146,13 @@ export default class DayPickerRangeController extends React.Component {
...
       if (!startDate) {
         endDate = day;
-        this.props.onFocusChange(START_DATE);
+        onFocusChange(START_DATE);
       } else if (isInclusivelyAfterDay(day, firstAllowedEndDate)) {
         endDate = day;
-        if (!keepOpenOnDateSelect) this.props.onFocusChange(null);
+        if (!keepOpenOnDateSelect) {
+          onFocusChange(null);
+          this.props.onClose({ startDate, endDate });
```
ljharb: Ideally this would be destructured out too

rafacianci: ahh ok.. I didnt because it was using just so I thought it should be better to use the props.. But I removing.. ;)

rafacianci: done. @ljharb :D

ljharb: Variables are free; passing the props object to user code is not :-)

rafacianci: nice.. :)
can you approve it?
and.. can I merge and squash?

ljharb: You can always rebase down to an atomic set of commits; please do so frequently :-)

rafacianci: done @ljharb :)

rafacianci:
@majapw can you review it?

rafacianci:
\o// thanks.. ☺️

majapw:
@ljharb The new travis tests are failing in a weird way. Specifically there seems to be an error with installing enzyme when testing with React 15... Can you take a look?

ljharb:
It's because React 15.5, due to new warnings, is a breaking change for us and for enzyme. The quick fix is to lock React down to < 15.5; pending the changes required to run the tests on both 15.4 and 15.5+.

majapw:
@ljharb Enzyme has a fix, I just need to add `react-test-renderer` to my deps according to @lelandrichardson, which I will do. :)

majapw:
PR open here - https://github.com/airbnb/react-dates/pull/442

senseluo:
just pull the latest code and got a warning message like: 
Warning: Unknown prop `onClose` on <input> tag. Remove this prop from the element.

ljharb:
@senseluo filed as #444; fix incoming.

haunguyen90:
great.... waiting for the update

majapw:
@senseluo can you give any more information? Does this happen in the storybook? Does this happen on a particular story? Is this happening when you are trying to use the prop yourself? Which component? Please clarify so that we can help you.

PR #406, opened 2017-03-27
DaySize update

moonboots:
to: @majapw @ljharb 

Continuation of Maja's [original PR](https://github.com/airbnb/react-dates/pull/249) to add customizable daySize's. I rebased and inlined a [`39 + 1` calculation](https://github.com/airbnb/react-dates/commit/6dfdab41c721f484fb10bf8d4e18a0e9efc431c8#diff-7b46ca2b2b498da14bfa494d3106eea5L4) to `40` daySize default for a [slightly cleaner number](https://github.com/airbnb/react-dates/pull/249#discussion_r95916020).

I punted on my request to make the [padding configurable as well](https://github.com/airbnb/react-dates/pull/249#discussion_r95908679). It's on of a few remaining hardcoded sizes, and I think we can revisit with a more general solution like sizing with pure css. On this topic, is there any sizing logic that strictly requires js? I think the transition animation should be doable without knowing pixel sizes in js, e.g. using translate `-100%`.

This current version should be sufficient for our use, so PTAL.

### DIFF
--- a/constants.js
+++ b/constants.js
@@ -11,4 +11,6 @@ module.exports = {
 
   ANCHOR_LEFT: 'left',
   ANCHOR_RIGHT: 'right',
+
+  DAY_SIZE: 39,
 };
--- a/src/components/CalendarMonth.jsx
+++ b/src/components/CalendarMonth.jsx
@@ -3,7 +3,7 @@
 import React, { PropTypes } from 'react';
 import shallowCompare from 'react-addons-shallow-compare';
 import momentPropTypes from 'react-moment-proptypes';
-import { forbidExtraProps } from 'airbnb-prop-types';
+import { forbidExtraProps, nonNegativeInteger } from 'airbnb-prop-types';
 import moment from 'moment';
 import cx from 'classnames';
 
@@ -17,6 +17,7 @@ import {
   HORIZONTAL_ORIENTATION,
   VERTICAL_ORIENTATION,
   VERTICAL_SCROLLABLE,
+  DAY_SIZE,
 } from '../../constants';
 
 const propTypes = forbidExtraProps({
@@ -25,6 +26,7 @@ const propTypes = forbidExtraProps({
   enableOutsideDays: PropTypes.bool,
   modifiers: PropTypes.object,
   orientation: ScrollableOrientationShape,
+  daySize: nonNegativeInteger,
   onDayClick: PropTypes.func,
   onDayMouseEnter: PropTypes.func,
   onDayMouseLeave: PropTypes.func,
@@ -40,6 +42,7 @@ const defaultProps = {
   enableOutsideDays: false,
   modifiers: {},
   orientation: HORIZONTAL_ORIENTATION,
+  daySize: DAY_SIZE,
   onDayClick() {},
   onDayMouseEnter() {},
   onDayMouseLeave() {},
@@ -81,6 +84,7 @@ export default class CalendarMonth extends React.Component {
       onDayMouseEnter,
       onDayMouseLeave,
       renderDay,
+      daySize,
     } = this.props;
 
     const { weeks } = this.state;
@@ -105,6 +109,7 @@ export default class CalendarMonth extends React.Component {
                 {week.map((day, dayOfWeek) => (
                   <CalendarDay
                     day={day}
+                    daySize={daySize}
                     isOutsideDay={!day || day.month() !== month.month()}
                     modifiers={modifiers}
                     key={dayOfWeek}
--- a/src/components/CalendarMonthGrid.jsx
+++ b/src/components/CalendarMonthGrid.jsx
@@ -1,7 +1,7 @@
 import React, { PropTypes } from 'react';
 import shallowCompare from 'react-addons-shallow-compare';
 import momentPropTypes from 'react-moment-proptypes';
-import { forbidExtraProps } from 'airbnb-prop-types';
+import { forbidExtraProps, nonNegativeInteger } from 'airbnb-prop-types';
 import moment from 'moment';
 import cx from 'classnames';
 import { addEventListener, removeEventListener } from 'consolidated-events';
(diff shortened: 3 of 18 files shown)


moonboots:
Thanks @majapw, extracted `DAY_SIZE` into a constant fixed the right padding calc

src/components/CalendarMonth.jsx
```diff
@@ -25,6 +26,7 @@ const propTypes = forbidExtraProps({
   enableOutsideDays: PropTypes.bool,
   modifiers: PropTypes.object,
   orientation: ScrollableOrientationShape,
+  daySize: PropTypes.number,
```
ljharb: Should this allow any number, or only a positive integer? If the latter, let's use airbnb-prop-types to restrict it.

src/components/CalendarMonthGrid.jsx
```diff
@@ -33,6 +35,7 @@ const propTypes = forbidExtraProps({
   onMonthTransitionEnd: PropTypes.func,
   renderDay: PropTypes.func,
   transformValue: PropTypes.string,
+  daySize: PropTypes.number,
```
ljharb: Since this propType definition appears in multiple places, let's use a shared one so the source of truth only lives in one place.

src/components/DayPicker.jsx
```diff
@@ -38,6 +39,7 @@ const propTypes = forbidExtraProps({
   hidden: PropTypes.bool,
   initialVisibleMonth: PropTypes.func,
   renderCalendarInfo: PropTypes.func,
+  daySize: PropTypes.number,
```

src/components/DayPickerRangeController.jsx
```diff
@@ -43,6 +44,7 @@ const propTypes = forbidExtraProps({
   orientation: ScrollableOrientationShape,
   withPortal: PropTypes.bool,
   initialVisibleMonth: PropTypes.func,
+  daySize: PropTypes.number,
```
ljharb: And here

src/shapes/SingleDatePickerShape.js
```diff
@@ -35,10 +35,12 @@ export default {
   keepOpenOnDateSelect: PropTypes.bool,
   reopenPickerOnClearDate: PropTypes.bool,
   renderCalendarInfo: PropTypes.func,
+  daySize: PropTypes.number,
```
ljharb: And here

src/components/CalendarMonthGrid.jsx
majapw: Isn't that true of like... 90% of react-dates proptypes though?

src/components/CalendarMonth.jsx
majapw: i concur. airbnb-prop-types is already available to you.

src/components/CalendarMonthGrid.jsx
ljharb: yes, and ideally they'd be imported all from one place too - my personal preference is to export it from the leaf-est component that needs it.

majapw: I think that ca be clean up that we could do all at once in a follow up PR.

moonboots:
@majapw @ljharb Updated with `nonNegativeNumber`, deferred the proptype crunch, ptal

constants.js
```diff
@@ -11,4 +11,6 @@ module.exports = {
 
   ANCHOR_LEFT: 'left',
   ANCHOR_RIGHT: 'right',
+
+  DAY_SIZE: 39,
```
majapw: Didn't you change this one to 40?

src/components/CalendarMonth.jsx
```diff
@@ -25,6 +26,7 @@ const propTypes = forbidExtraProps({
   enableOutsideDays: PropTypes.bool,
   modifiers: PropTypes.object,
   orientation: ScrollableOrientationShape,
+  daySize: nonNegativeInteger,
```
majapw: Can you double-check that this doesn't need to be `nonNegativeInteger()`?

moonboots: Tested with `daySize={-1}`, `nonNegativeInteger` without extra function invocation correctly prints to `console.error`.

constants.js
moonboots: `40` was a mistake when I misinterpreted the [`getCalendarMonthWidth` logic](https://github.com/airbnb/react-dates/pull/406/files#diff-7b46ca2b2b498da14bfa494d3106eea5R2). It was causing the missing right padding you pointed out.

src/components/CalendarMonth.jsx
majapw: Oh! Great. :)

constants.js
majapw: Wonderful.

src/components/CalendarMonth.jsx
ljharb: thats an omission in airbnb-prop-types, which i'll fix in the next semver-major :-)

palashkaria: I think this does need to be nonNegativeInteger(). This leads to a lot of warnings for me, of the form:

```
Warning: CalendarDay: type specification of prop `daySize` is invalid; the type checker function must return `null` or an `Error` but returned a function. You may have forgotten to pass an argument to the type checker creator (arrayOf, instanceOf, objectOf, oneOf, oneOfType, and shape all require an argument).

```

I have opened a PR to fix this issue: https://github.com/airbnb/react-dates/pull/664

ljharb: in fact nonNegativeInteger isn't thunkified, so this is correct as-is https://unpkg.com/airbnb-prop-types@2.8.0/src/nonNegativeInteger.js

ljharb: @palashkaria what version of react are you using?

PR #409, opened 2017-03-29
Fix input focus bottom bar on iOS

moonboots:
@majapw @backwardok 

There's an issue on  iOS>=8 where focusing on an `<input>` element brings up a native form navigation bottom bar. This covers UI at the bottom of the page and sometimes requires extra confirmation to dismiss. In this PR, I'm shifting focus to the DayPicker element where the tap interaction occurs. There's still a brief flash of that bar, and I'm open to other ideas. Would swapping the `<input type="text">` with an `<input type="button">` for touch devices check out from an accessibility perspective?

### DIFF
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -345,6 +345,11 @@ export default class DateRangePicker extends React.Component {
             withFullScreenPortal={withFullScreenPortal}
             onDatesChange={onDatesChange}
             onFocusChange={onFocusChange}
+            focusDayPicker={() => {
+              if (this.dayPicker) {
+                setTimeout(() => this.dayPicker.focus(), 0);
+              }
+            }}
             phrases={phrases}
             screenReaderMessage={screenReaderInputMessage}
           />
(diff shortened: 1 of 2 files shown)


src/components/DateRangePicker.jsx
```diff
@@ -345,6 +345,11 @@ export default class DateRangePicker extends React.Component {
             withFullScreenPortal={withFullScreenPortal}
             onDatesChange={onDatesChange}
             onFocusChange={onFocusChange}
+            focusDayPicker={() => {
```
majapw: I believe this behavior should only happen when withPortal or withFullscreenPortal is used. This is not the behavior I want for other situations. I also feel like this code belongs a layer down in the `DayPicker` itself, not here.

moonboots: Makes sense to move the focusing into `DayPicker`. I think this common parent will still need to coordinate between the input and daypicker though because they're in isolated subtrees.

majapw:
Closing in favor of https://github.com/airbnb/react-dates/pull/410

Issue #246, opened 2017-01-11
Mobile Date Picker Doesn't Open on first Click


iamkevingreen:
I'm experiencing an issue with various settings on the latest iOS with the portal and without the portal. In which a prompt on the device appears either via a Done or an X as shown here: 

![glass_and_iphone_6_ _ios_10_2__14c89_](https://cloud.githubusercontent.com/assets/2124358/21869594/50c0e45a-d826-11e6-887c-9e29b2eabcd1.png)

The user has to click done in order for the prompt to pop up. This is only an issue on iOS devices and not on android phones. 

```
      <SingleDatePicker
        {...this.props}
        id="date_input"
        date={date}
        numberOfMonths={months}
        placeholder="00/00/00"
        focused={focused}
        onDateChange={this.onDateChange}
        onFocusChange={this.onFocusChange}
      />
```

```
      <SingleDatePicker
        {...this.props}
        id="date_input"
        date={date}
        withFullScreenPortal={true}
        withPortal={true}
        numberOfMonths={months}
        placeholder="00/00/00"
        focused={focused}
        onDateChange={this.onDateChange}
        onFocusChange={this.onFocusChange}
      />
```

majapw:
Huh. That's really weird. I haven't seen that before. This is in Safari on IOS 8? I will investigate.

@moonboots have you been testing on any iOS devices recently?

iamkevingreen:
This is on Safari and Chrome it's specific to iOS handling of fields not just safari.

moonboots:
I haven't seen this issue before and unfortunately wasn't able to reproduce in the storybook examples on an iPhone. I think the bar with the "Done" button appears because `DateInput` contains a text field for keyboard input, but this shouldn't prevent the focus/appearance of the calendar. @iamkevingreen, could you provide the implementation of your `onFocusChange` callback? Does it interact with any other elements on the page? Thanks

iamkevingreen:
Sure @moonboots , let me provide my entire implementation: 

```
 updateStart(startDate) {
    this.setState({ startDate })
 }
render() {
   return (
       <SingleDatePickerWrapper startDatePlaceholderText="00/00/00" updateDate={this.updateStart} />
     )
}
```

```
class SingleDatePickerWrapper extends React.Component {
  constructor(props) {
    super(props);
    this.state = {
      focused: false,
      date: null,
      months: 1
    };

    this.onDateChange = this.onDateChange.bind(this)
[...]

PR #410, opened 2017-03-30
Separates out focusedInput from DOM Focus and focuses on the calendar for withPortal implementations

majapw:
This, as far as I can tell, should be a fix for https://github.com/airbnb/react-dates/issues/246.

I'm going to favor this PR over https://github.com/airbnb/react-dates/pull/409 because it paves the way better for https://github.com/airbnb/react-dates/pull/301.

@airbnb/webinfra @moonboots PTAL

### DIFF
--- a/src/components/DateInput.jsx
+++ b/src/components/DateInput.jsx
@@ -19,6 +19,8 @@ const propTypes = forbidExtraProps({
   onFocus: PropTypes.func,
   onKeyDownShiftTab: PropTypes.func,
   onKeyDownTab: PropTypes.func,
+
+  isFocused: PropTypes.bool, // describes actual DOM focus
 });
 
 const defaultProps = {
@@ -35,6 +37,8 @@ const defaultProps = {
   onFocus() {},
   onKeyDownShiftTab() {},
   onKeyDownTab() {},
+
+  isFocused: false,
 };
 
 export default class DateInput extends React.Component {
@@ -62,10 +66,10 @@ export default class DateInput extends React.Component {
   }
 
   componentDidUpdate(prevProps) {
-    const { focused } = this.props;
-    if (prevProps.focused === focused) return;
+    const { focused, isFocused } = this.props;
+    if (prevProps.focused === focused && prevProps.isFocused === isFocused) return;
 
-    if (focused) {
+    if (focused && isFocused) {
       this.inputRef.focus();
       this.inputRef.select();
     } else {
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -11,6 +11,7 @@ import { DateRangePickerPhrases } from '../defaultPhrases';
 
 import OutsideClickHandler from './OutsideClickHandler';
 import getResponsiveContainerStyles from '../utils/getResponsiveContainerStyles';
+import isTouchDevice from '../utils/isTouchDevice';
 
 import isInclusivelyAfterDay from '../utils/isInclusivelyAfterDay';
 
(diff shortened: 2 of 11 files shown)


src/components/DateRangePicker.jsx
```diff
@@ -125,9 +136,44 @@ export default class DateRangePicker extends React.Component {
...
+    });
+
     onFocusChange(null);
   }
 
+  onDateRangePickerInputFocus(focusedInput) {
+    const { onFocusChange, withPortal, withFullScreenPortal } = this.props;
+
+    if (focusedInput) {
+      const moveFocusToDayPicker = withPortal || withFullScreenPortal;
```
moonboots: Would it make more sense to target based on touch instead of portal presence? The mobile vertical scrollable date picker doesn't use these portal props, but it also requires daypicker focus.

src/components/DateRangePickerInputController.jsx
```diff
@@ -84,11 +86,13 @@ const defaultProps = {
...
   customCloseIcon: null,
 
+  isFocused: false,
+
   // i18n
   phrases: DateRangePickerInputPhrases,
 };
 
-export default class DateRangePickerInputWithHandlers extends React.Component {
+export default class DateRangePickerInputController extends React.Component {
```
moonboots: 💯 This has thrown a wrench in many of my greps

src/components/DateRangePicker.jsx
majapw: hmm, well my thought is that it only makes sense to do this when the input is not visible... but maybe on touch devices makes sense too. Is the input going to be visible at the top of the VERTICAL_SCROLLABLE DayPicker? I forget what the designs look like

src/components/DateRangePickerInputController.jsx
majapw: TYPOS. hooray!

src/components/DateRangePicker.jsx
moonboots: The input will always be visible with `VERTICAL_SCROLLABLE`, and the focus should stick back to the day picker after tapping an input.

majapw: Okay, let's have this occur always for portals and also always on touch devices.

majapw:
@moonboots PTAL

src/components/DateInput.jsx
```diff
@@ -19,6 +19,8 @@ const propTypes = forbidExtraProps({
   onFocus: PropTypes.func,
   onKeyDownShiftTab: PropTypes.func,
   onKeyDownTab: PropTypes.func,
+
+  isFocused: PropTypes.bool, // describes actual DOM focus
```
backwardok: A little confused about the distinction between this bool and `focused` by name alone. Seems like it'd be better as something like `focusInputDOM` or something that makes it more explicit.

src/components/DateRangePicker.jsx
```diff
@@ -125,9 +141,44 @@ export default class DateRangePicker extends React.Component {
     const { onFocusChange } = this.props;
     if (!this.isOpened()) return;
 
+    this.setState({
+      isDateRangePickerInputFocused: false,
+      isDayPickerFocused: false,
+      showKeyboardShortcuts: false,
```
backwardok: this doesn't exist yet does it? Isn't that in the a11y PR?

src/components/DayPicker.jsx
```diff
@@ -447,37 +461,49 @@ export default class DayPicker extends React.Component {
...
           </div>
 
           <div
-            className={transitionContainerClasses}
-            ref={(ref) => { this.transitionContainer = ref; }}
-            style={transitionContainerStyle}
+            className="DayPicker__focus-region"
+            ref={(ref) => { this.container = ref; }}
+            role="region"
+            tabIndex={0}
```
backwardok: is this something you specifically want users to be able to tab to or just to focus? Seems like something that should be more for focusing purposes, in which case you'd want a tabindex of -1

src/components/DateInput.jsx
majapw: I wanted parity with focusing the DayPicker as well.

src/components/DateRangePicker.jsx
majapw: oops!

PR #414, opened 2017-03-31
Tap performance followup, fixes and specs

moonboots:
to: @majapw @lencioni 

Bug in my previous `isSameDay` change. Moment's `.day()` unfortunately returns day of week, not day of month like I thought. I didn't realize the PR was going to be merged (was adding specs for the modifiers changes, starting to add specs for `isSameDay` when I realized the error).

### DIFF
--- a/src/utils/isSameDay.js
+++ b/src/utils/isSameDay.js
@@ -2,8 +2,5 @@ import moment from 'moment';
 
 export default function isSameDay(a, b) {
   if (!moment.isMoment(a) || !moment.isMoment(b)) return false;
-  // Speeding up by comparing most likely to differ unit first
-  return a.day() === b.day() &&
-    a.month() === b.month() &&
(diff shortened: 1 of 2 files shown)


src/utils/isSameDay.js
```diff
@@ -2,8 +2,5 @@ import moment from 'moment';
 
 export default function isSameDay(a, b) {
   if (!moment.isMoment(a) || !moment.isMoment(b)) return false;
-  // Speeding up by comparing most likely to differ unit first
-  return a.day() === b.day() &&
```
moonboots: Moment's day of month function is [`.date()`](http://momentjs.com/docs/#/get-set/date/) 😓

majapw: wait why are we changing this back completely

moonboots:
Thanks, will do

PR #418, opened 2017-04-02
[Tests] increase test coverage

ljharb:
Please note that `OutsideClickHandler` was uselessly forbidding extra defaultProps, instead of extra actual props. It'll be semver-major to bump that, so I added a TODO.


ljharb:
@majapw, how about a :stamp: for me? :-D

src/components/OutsideClickHandler.jsx
```diff
@@ -1,16 +1,16 @@
 import React, { PropTypes } from 'react';
-import { forbidExtraProps } from 'airbnb-prop-types';
+// import { forbidExtraProps } from 'airbnb-prop-types'; // TODO: add to propTypes; semver-major
```
ljharb: That's a tough call; since someone could always have imported it directly by a deep file path, i'd say yes. Some say no, since it's not a documented entry point.

PR #419, opened 2017-04-03
update DayPicker story to show single month

dburles:
Looks like at one point the default was changed to two months while the story was representing a default of one

### DIFF
--- a/stories/DayPicker.js
+++ b/stories/DayPicker.js
@@ -55,8 +55,12 @@ storiesOf('DayPicker', module)
   .addWithInfo('single month', () => (
     <DayPicker numberOfMonths={1} />
   ))
+  .addWithInfo('3 months', () => (
+    <DayPicker numberOfMonths={3} />
+  ))
   .addWithInfo('vertical', () => (
     <DayPicker
+      numberOfMonths={2}
       orientation={VERTICAL_ORIENTATION}
     />
   ))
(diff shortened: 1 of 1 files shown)


stories/DayPicker.js
```diff
@@ -52,12 +52,11 @@ storiesOf('DayPicker', module)
   .addWithInfo('with custom day size', () => (
     <DayPicker daySize={50} />
   ))
-  .addWithInfo('more than one month', () => (
```
ljharb: We should have all of explicit 1, > 1, and "default". That way we don't have to update this in the future if the default changes.

stories/DayPicker.js
```diff
@@ -52,12 +52,11 @@ storiesOf('DayPicker', module)
...
-    <DayPicker numberOfMonths={2} />
+  .addWithInfo('single month', () => (
+    <DayPicker numberOfMonths={1} />
   ))
   .addWithInfo('vertical', () => (
     <DayPicker
-      numberOfMonths={2}
```
ljharb: Similarly, having these explicit ensures that a change of the default won't suddenly make these stories useless.

stories/DayPicker.js
dburles: Just noticed there's  a pattern used in other stories that's *single month* and *3 months*, I'll go with that

ljharb: Sounds great!

majapw:
Thanks!

Issue #340, opened 2017-02-21
Firefox SingleDatePicker withPortal not usable


codepunkt:
On Firefox the `SingleDatePicker` with the `withPortal` option triggers `onFocusChange` twice, once with `{ focused: true }` and right afterwards `{ focused: false }`. This makes the portal open and close in a matter of half a second, rendering the datepicker unusable.

```javascript
import { SingleDatePicker } from 'react-dates';

export default ({ date, focused, onFocusChange, onDateChange }) => (
    <SingleDatePicker
      id="date_input"
      withPortal
      date={date}
      focused={focused}
      onDateChange={onDateChange}
      onFocusChange={onFocusChange}
      isOutsideRange={(day) => day.isAfter(moment())}
      displayFormat="DD.MM.YYYY"
    />
);
```

Reproduced on Firefox Developer Edition 53.0a2 and Firefox 51.0.1.

SiqiTian-minted:
I'm having a similar issue, but with Chrome when force device to mobile. Besides the double firing of `onFocus`, I also see this error on init:

```
Uncaught (in promise) TypeError: Cannot read property 'getBoundingClientRect' of undefined
    at SingleDatePicker.responsivizePickerPosition (SingleDatePicker.js:443)
    at SingleDatePicker.componentDidMount (SingleDatePicker.js:258)
    at ReactCompositeComponent.js:264
    at measureLifeCyclePerf (ReactCompositeComponent.js:74)
    at ReactCompositeComponent.js:263
    at CallbackQueue.notifyAll (CallbackQueue.js:67)
    at ReactReconcileTransaction.close (ReactReconcileTransaction.js:81)
    at ReactReconcileTransaction.closeAll (Transaction.js:204)
    at ReactReconcileTransaction.perform (Transaction.js:151)
    at ReactUpdatesFlushTransaction.perform (Transaction.js:138)
```

But none of the problem happens on desktop view.

jonathanong:
weird issue... i found that the overlay is shown on firefox if i hold down the left click button or just right click, but then i get the context menu

rafacianci:
I having the same problem.. I set `withPortal: false` and `withFullScreenPortal: true` then I changed the closed button to have it working.. 😓

sam-cook:
Same here, a fix would be awesome

lukefrake:
Yeah, the same issue in Firefox here 😢

phamcharles:
[...]

PR #421, opened 2017-04-03
Uses React's onClick handler instead of `OutsideClickHandler` with `withPortal` implementations

majapw:
Fix for https://github.com/airbnb/react-dates/issues/340

This fixes an issue with Firefox specifically where the order of operations of the events caused the `withPortal` DayPicker to immediately close upon opening.

to: @lencioni 
fyi: @phamcharles

### DIFF
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -292,10 +292,11 @@ export default class DateRangePicker extends React.Component {
     const closeIcon = customCloseIcon || (<CloseButton />);
 
     return (
-      <div
+      <div // eslint-disable-line jsx-a11y/no-static-element-interactions
         ref={(ref) => { this.dayPickerContainer = ref; }}
         className={this.getDayPickerContainerClasses()}
         style={dayPickerContainerStyles}
+        onClick={onOutsideClick}
       >
         <DayPickerRangeController
           ref={(ref) => { this.dayPicker = ref; }}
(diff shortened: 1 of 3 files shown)


ljharb:
Do we want to do this unconditionally, or only when `withPortal` is false?

majapw:
@ljharb If you look at where `onOutsideClick` the variable is defined, we do it conditionally:
```
const onOutsideClick = (!withFullScreenPortal && withPortal) ? this.onClearFocus : undefined;
```
which is what we want.

ljharb:
Is this an issue with OutsideClickHandler itself; or with our event handler not checking the event target properly? In other words, is this something we could solve while keeping OutsideClickHandler (even if this is a good hotfix in the interim)?

majapw:
@ljharb So I don't think it is specifically an issue with the `OutsideClickHandler`. It def does check event target properly, but the issue is with the way the native event system plays with the react event system/render cycle.

Basically, here is what happens on Chrome:
1) input takes focus event when you click on the input
2) focus event moves up the tree and triggers `onFocusChange`
3) `onFocusChange` updates `state.focusedInput`
4) click event happens somewhere between 2 and here
5) render cycle begins, rendering the withPortal calendar to the page

Conversely, here is what happens in Firefox
1) input takes focus event when you click on the input
2) focus event moves up the tree and triggers `onFocusChange`
3) `onFocusChange` updates `state.focusedInput`
4) render cycle begins, rendering the withPortal calendar to the page
5) click event gets triggered 
6) because the click is now outside the root node (the visible calendar area) of the now rendered`OutsideClickHandler`, `onOutsideClick` gets triggered
7) the calendar closes

Because the initial event is a (1) a focus event and (2) a react event, and the outside click event is (1) a click event and (2) a native JS event, I think there is some conflict between making them play nice together in this one weird edge case. I discussed this with @lencioni and I think this is *the* solution and not just a hotfix.

majapw:
Do you want to :stamp: me @ljharb? :D

PR #429, opened 2017-04-07
Fix a couple of nits as a result of the a11y PR

majapw:
- blur the CalendarDay is month navigation was done with the mouse
- prevent home/end/pagedown/pageup from scrolling the page
- fixed some weirdness with the aria-label text on Calendar Day components when not using the DRP explicitly

to: @airbnb/webinfra @moonboots

### DIFF
--- a/src/components/CalendarDay.jsx
+++ b/src/components/CalendarDay.jsx
@@ -102,8 +102,7 @@ export default class CalendarDay extends React.Component {
     const formattedDate = `${day.format('dddd')}, ${day.format('LL')}`;
 
     let ariaLabel = getPhrase(chooseAvailableDate, {
-      checkin_date: formattedDate,
-      checkout_date: formattedDate,
+      date: formattedDate,
     });
 
     if (BLOCKED_MODIFIER in modifiers && modifiers[BLOCKED_MODIFIER](day)) {
--- a/src/components/DayPicker.jsx
+++ b/src/components/DayPicker.jsx
@@ -21,6 +21,7 @@ import DayPickerKeyboardShortcuts, {
 import getTransformStyles from '../utils/getTransformStyles';
 import getCalendarMonthWidth from '../utils/getCalendarMonthWidth';
 import isTouchDevice from '../utils/isTouchDevice';
+import getActiveElement from '../utils/getActiveElement';
 
 import ScrollableOrientationShape from '../shapes/ScrollableOrientationShape';
 
@@ -185,6 +186,7 @@ export default class DayPicker extends React.Component {
       showKeyboardShortcuts: props.showKeyboardShortcuts,
       onKeyboardShortcutsPanelClose() {},
       isTouchDevice: isTouchDevice(),
+      withMouseInteractions: true,
     };
 
     this.onKeyDown = this.onKeyDown.bind(this);
@@ -244,6 +246,7 @@ export default class DayPicker extends React.Component {
           showKeyboardShortcuts,
           onKeyboardShortcutsPanelClose,
           focusedDate,
+          withMouseInteractions: false,
         });
       } else {
         this.setState({ focusedDate: null });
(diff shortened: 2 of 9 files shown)


src/components/DayPicker.jsx
```diff
@@ -509,6 +517,12 @@ export default class DayPicker extends React.Component {
       translationValue: 0,
       nextFocusedDate: null,
       focusedDate: newFocusedDate,
+    }, () => {
+      // we don't want to focus on the relevant calendar day after a month transition
+      // if the user is navigating around using a mouse
+      if (this.state.withMouseInteractions && document.activeElement) {
```
lencioni: Should you check that document.activeElement is not the body?

majapw: Would it cause a problem if it was?

src/components/DayPicker.jsx
```diff
@@ -301,10 +305,12 @@ export default class DayPicker extends React.Component {
         didTransitionMonth = this.maybeTransitionPrevMonth(newFocusedDate);
         break;
       case 'Home':
+        if (e) e.preventDefault();
```
ljharb: I'm curious when there will not be an `e` object in an event handler?

(it's probably best to keep this as-is, ofc, for consistency with the rest of the branches; and removing the `if` in the other branches would be semver-major, but i'd like to see us do that in the future)

src/components/DayPicker.jsx
ljharb: Should this also check for `document`?

src/components/DayPicker.jsx
majapw: Mostly for testing purposes, which like, could be done better. I can remove these in another PR.

majapw:
I also fixed some weirdness that I had not fully thought through with the phrases when it came to non-drp usage! PTAL @lencioni @ljharb

src/components/DayPicker.jsx
lencioni: `document.activeElement` is the body by default. So when you blur, it gets set to the body. So this will end up blurring on the body every time. Probably not a big deal, but it is possible that this could kick off event listeners or other side-effects, depending on what has been registered on the page. Something like this might work? (not sure about browser support of HTMLBodyElement off the top of my head though)

```js
!(document.activeElement instanceof HTMLBodyElement)
```

lencioni: or I guess more simply

```js
document.activeElement !== document.body
```

majapw: I'll add the check, although this is somewhat of an extreme edge case given the way the datepicker is managing focus. 

Basically, the `setState` that this is a callback of triggers a series of events that always ends in a `CalendarDay` being focused. So while it is possible that the user could blur between the `componentDidUpdate` call that focuses the CalendarDay in that leaf node and this setState callback 4 levels up in the tree, it's highly unlikely.

src/components/DayPicker.jsx
ljharb: Awesome, let's do that - we can pass `{ preventDefault() {} }` in tests if needed.

src/components/CalendarDay.jsx
```diff
@@ -102,8 +102,7 @@ export default class CalendarDay extends React.Component {
     const formattedDate = `${day.format('dddd')}, ${day.format('LL')}`;
 
     let ariaLabel = getPhrase(chooseAvailableDate, {
-      checkin_date: formattedDate,
-      checkout_date: formattedDate,
+      date: formattedDate,
```
ljharb: is this a bugfix, or semver-major?

src/utils/getPhrase.jsx
```diff
@@ -1,6 +1,8 @@
 import React from 'react';
 
 export default function getPhrase(phrase, args) {
+  if (!phrase) return '';
```
ljharb: tests for this?

src/components/CalendarDay.jsx
majapw: bugfix i'll say.

src/utils/getPhrase.jsx
majapw: I should
probably write some tests
sigh

PR #446, opened 2017-04-13
Update conditions when adjustDayPickerHeight is called

nboliver:


majapw:
Can you explain the motivation for why you want to do this? Because reaching into the child component to grab a method is generally an antipattern.

nboliver:
Updated this PR in response to your comments. Let me know if you think that covers the possible situations when it might need to update.

nboliver:
@majapw Any further thoughts on this? I was wondering if maybe `adjustDayPickerHeight` should always be called when the component receives props. It's a tough balance between performance and making sure it is always the correct dimensions.

```
...
  if (this.isHorizontal()) {
    this.adjustDayPickerHeight();
  }
...
}
```

nboliver:
@ljharb rebased - anything else required? what do you think of my comment above about calling `adjustDayPickerHeight` every time the component receives props?

src/components/DayPicker.jsx
```diff
@@ -260,13 +260,20 @@ export default class DayPicker extends React.Component {
...
   componentDidUpdate(prevProps, prevState) {
+    const { orientation, daySize } = this.props;
     const { monthTransition, currentMonth, focusedDate } = this.state;
     if (monthTransition || !currentMonth.isSame(prevState.currentMonth)) {
       if (this.isHorizontal()) {
         this.adjustDayPickerHeight();
       }
     }
 
+    if (orientation !== prevProps.orentation || daySize !== prevProps.daySize) {
```
majapw: Not super sure why this isn't just integrated into the above conditional? Let's make that change and then merge this in.

ljharb: I think it still needs nonzero test coverage before going in tho, if possible.

brosewarne:
I've attempted to update / finish this in #640

majapw:
Closing in favor of #640

PR #448, opened 2017-04-14
Only update phrases object when the focusedInput actually changes.

majapw:
Realized this when I was prototyping some perf changes... but basically the modifiers object is not the only one that is changing (unnecessarily) on every render.

to: @moonboots @ljharb @airbnb/webinfra

### DIFF
--- a/src/components/DayPickerRangeController.jsx
+++ b/src/components/DayPickerRangeController.jsx
@@ -116,6 +116,7 @@ export default class DayPickerRangeController extends React.Component {
     super(props);
     this.state = {
       hoverDate: null,
+      phrases: props.phrases,
     };
 
     this.isTouchDevice = isTouchDevice();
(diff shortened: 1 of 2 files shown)


majapw:
@ljharb I'm not entirely sure how to test this, especially given that it is in `componentWillReceiveProps`

ljharb:
@majapw `wrapper.instance().componentWillReceiveProps`, unless `.setProps()` works

majapw:
Added some tests! 🎉

test/components/DayPickerRangeController_spec.jsx
```diff
@@ -28,6 +28,125 @@ describe('DayPickerRangeController', () => {
...
 
+  describe('#componentWillReceiveProps', () => {
+    const phrases = {
+      chooseAvailableDate: 'test1',
+      chooseAvailableStartDate: 'test2',
+      chooseAvailableEndDate: 'test3',
+    };
+
+    describe('neither props.focusedInput nor props.phrases have changed', () => {
+      it('state.phrases does not change', () => {
```
lencioni: Did you mean to actually write a test here?

majapw: I did. I just forgot. :P

src/components/DayPickerRangeController.jsx
```diff
@@ -127,6 +128,27 @@ export default class DayPickerRangeController extends React.Component {
     this.getFirstFocusableDay = this.getFirstFocusableDay.bind(this);
   }
 
+  componentWillReceiveProps(nextProps) {
+    const { focusedInput, phrases } = nextProps;
+
+    if (focusedInput !== this.props.focusedInput || phrases !== this.props.phrases) {
+      // set the appropriate CalendarDay phrase based on focusedInput
+      let chooseAvailableDate = phrases.chooseAvailableDate;
```
lencioni: You could use destructuring here if you like.

lencioni:
Just curious, did you measure the performance impact of this change?

majapw:
@lencioni this change on its own will not have any performance impact due to the fact that each `CalendarDay` will still be updated on every hover or mouse interaction because `modifiers` gets rebuilt in every render at the top-level. However, I have been working on rearchitecturing modifiers pretty much entirely in a separate change and the combination of the two will lead to a huge speed-up.

Issue #447, opened 2017-04-14
Change calendar system to Jalali canendar


cyclops24:
I want to user `react-dates` as a Persian datepicker. Persians use another calendar system that called `jalali calendar`. We have a popular library called [moment-jalaali](https://momentjs.com/docs/#/plugins/jalaali/) that works perfectly as a `moment` plug-in.
Now my question is that: How we can use `jalali calendar` system with `react-dates`?
I try below code, but it's just change locale and strings not entire calendar system.

```
import React, { Component } from 'react';
import './App.css';
import moment from 'moment-jalaali';
import { DateRangePicker, SingleDatePicker, DayPickerRangeController } from 'react-dates';
import 'react-dates/lib/css/_datepicker.css';

class App extends Component {

  render() {
    moment.loadPersian(); // >>>>>>>>>> This change locale to Persian successfully but how to change calendar system?
    return (
      <div className="App">

        <SingleDatePicker
          date={moment()}
          onDateChange={date => console.log(date)} // PropTypes.func.isRequired
        focused={true}
          onFocusChange={({ focused }) => console.log(focused)} // PropTypes.func.isRequired
        />

      </div>
    );
  }
}

export default App;
```

This is output screen shot for above code you can see that calendar system is wrong:
![jalali_1](https://cloud.githubusercontent.com/assets/8266479/25031858/f7fb2940-20e4-11e7-8074-82a1a3bbc653.png)

cyclops24:
I think it's related to #259

cyclops24:
I added "renderMonth" and a simple example as a start point ( #449 ) for adding Persian calendar. Some small problem still exist in completely support Persian calendar (like days order, ...). But I think we can do this by add some other PR.

majapw:
Fixed in v10.2.0!

mpourismaiel:
`renderMonth` and `renderDay` fixes the locale problem but the issue with the calendar itself still stands. 

![calendar](https://image.ibb.co/drxmav/Clipboard_image_2017_06_07_17_03_22.png)

First of the month in the calendar is marked 11, which is 11th of the respective month in Jalali while it should obviously be 1. Am I missing something?

ImanMh:
@mpourismaiel Have you found any solutions?

mpourismaiel:
@ImanMh They thought the issue was that there was no way to change the locale for month names which actually was another bug.

ljharb:
[...]

PR #449, opened 2017-04-14
[Feature] add "renderMonth" support + add Persian locale story

cyclops24:
This PR add `renderMonth` to react-dates and help to add custom calendar like Jalali (Persian) calendar in future.
It's needed for #447 

This is a simple screenshot from story page:
jalali_5

### DIFF
--- a/src/components/CalendarMonth.jsx
+++ b/src/components/CalendarMonth.jsx
@@ -35,6 +35,7 @@ const propTypes = forbidExtraProps({
   onDayClick: PropTypes.func,
   onDayMouseEnter: PropTypes.func,
   onDayMouseLeave: PropTypes.func,
+  renderMonth: PropTypes.func,
   renderDay: PropTypes.func,
 
   focusedDate: momentPropTypes.momentObj, // indicates focusable day
@@ -55,6 +56,7 @@ const defaultProps = {
   onDayClick() {},
   onDayMouseEnter() {},
   onDayMouseLeave() {},
+  renderMonth: null,
   renderDay: null,
 
   focusedDate: null,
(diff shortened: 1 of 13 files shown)


src/components/CalendarMonth.jsx
```diff
@@ -104,7 +107,7 @@ export default class CalendarMonth extends React.Component {
     } = this.props;
 
     const { weeks } = this.state;
-    const monthTitle = month.format(monthFormat);
+    const monthTitle = renderMonth ? renderMonth(month) : month.format(monthFormat);
```
ljharb: If we're calling a user function here, we'd need to somehow assert that `monthTitle` is a string, to match `month.format()` - and throw if not.

stories/SingleDatePicker.js
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import moment from 'moment';
+import momentJalaali from 'moment-jalaali';
```
ljharb: Does this need to be added as a dev dep?

cyclops24: Yes I add it to devDependencies.

src/components/CalendarMonth.jsx
cyclops24: So what is your suggestion?
I do this part as a code that now exist in `CalendarDay` component:
https://github.com/airbnb/react-dates/blob/master/src/components/CalendarDay.jsx#L131

ljharb: hm, ok, fair point. this can stay as-is, and we can address them all holistically in the future.

cyclops24: OK, Any other change needs for this PR?

cyclops24:
@majapw do you have any idea about this?

**CC: ** @ljharb

Issue #272, opened 2017-01-30
Too slow for mobile


wub:
The testing team at work have just (sadly!) rejected my branch because of how slow this is on mobile. Even after memoization, hacks, etc, I can't make this fast enough for anything but the latest devices, even if I drop in the vanilla picker.

I did notice that Airbnb themselves use a different component for mobile devices.

Anything I can do about this?

Should there be a note somewhere in the README, making sure people test on older devices? It's important to know what a potentially large percentage of your userbase will experience.

Once the project is done, I'll loop back and see if I can contribute some performance enhancements.

Cheers for all the hard work guys!

sontek:
Are you testing with `5.2.0`?   There were some performance improvements in `5.1.0` that might help you.  Its still super slow but it is reasonable to use after the dialog is opened.  It takes a couple seconds for the datepicker to open still

wub:
Thanks, @sontek. Yep, I'm using the latest version.

We've moved to [another one](http://react-day-picker.js.org/) that suits our needs; it's much faster, and much lighter.

majapw:
Hi @wub, thanks for the feedback. We're continuously improving the perf and working to get this into a state we're happy with on mobile. Would love to take any contributions you have in the future.

Can you provide any information on the range of devices your testing team looked at? 

As for this statement:
> As testament to this, even Airbnb themselves use a completely different component for mobile devices.
This is the legacy datepicker that we are working to replace. This is a testament to nothing beyond the fact that we haven't had the bandwidth to implement and land on a mobile design we are happy with. This is likely to change soon.

lencioni:
Depending on where you are seeing the slowness, #286 might help with this some.

isaachinman:
@wub Can you explain (or show examples) of how you memoized functions? I assume you're referring to the methods that are getting called thousands of times, like `isDayBlocked` and `isOutsideRange`?

While mobile is less of a concern for me, I believe memoization and user-provided performance enhancement is unfortunately necessary to make this component usable.

wub:
@majapw, @lencioni, @isaachinman: thanks for the feedback! I've also edited my post to sound less snarky, sorry about that...

@majapw - We experienced these slowdowns on the iPhone 5S, and the iPad 2. I understand these are old devices, but sadly they represent a sizeable portion of our userbase.

@isaachinman - Thanks for the heads up on moize, I'll look into it and report back. I can see it being useful for all kinds of functional stateless components, actually.

[...]

PR #450, opened 2017-04-14
Perf rearchitecting

majapw:
This PR is a dramatic rearchitecture of the way that we handle modifiers in `react-dates` on the whole.

Instead of passing down an object of modifier string => modifiers functions to each CalendarDay component (one that gets rebuilt on every render call) and expecting the CalendarDay component to take care of updating itself, we now put the burden on the top of the tree instead of on the leaves. The previous model meant that whenever an interaction happened on an individual day, even a hover interaction, every single CalendarDay would have to recalculate its modifiers and rerender as a result. The idea was that an interaction on one day might have effects elsewhere in the calendar (like the `hovered-span`). The effect unfortunately was that react-dates was [really god damn slow](https://github.com/airbnb/react-dates/issues/272).

In this new model, the DayPickerRangeController maintains a map with the following structure:
```
{
  MONTH_ISO_1: {
    DAY_ISO_1: Set(['modifier_1', 'modifier_2', ...]),
    DAY_ISO_2: Set(['modifier_1', 'modifier_2', ...]),
    ...
  },
  ...
}
```
It passes this down the tree such that each `CalendarMonth` and each `CalendarDay` only gets the information that pertains to it. This means that the updating of these modifiers is also handled at the top-level and is done in the `componentWillReceiveProps`, `onDayMouseEnter`, and `onDayMouseLeave` methods of the `DayPickerRangeController`. Fortunately, this allows us to more finely tune which days get updated and speeds up the rerendering/updating process dramatically.

I did some preliminary testing and saw the following improvements:
*Before:*
screen shot 2017-04-13 at 10 35 39 pm

*After:*
screen shot 2017-04-13 at 10 55 41 pm

I will update those graphs when everything is in place.

I still need to finish writing some tests, clean up some other tests, and also get this working with `enableOutsideDays` and the `SingleDatePicker`.  The bulk of the work is in `DayPickerRangeController` right now, but I've been musing on ways to pull that in an HOC or something. SUGGESTIONS ARE WELCOME.

In any case, I would appreciate a first review pass while I finish up the above points! <3
@lencioni @ljharb @moonboots @backwardok 

UPDATE: Here are the final perf measurements!
**Before:**
screen shot 2017-05-01 at 3 19 09 pm

**After:**
screen shot 2017-05-01 at 4 38 44 pm

Here are the findings:
 - The overhead on initial mount remains about the same (~400-500ms)
 - Any calendar interaction (hover, click, etc.) consistently caused a 65ms update in the old version and now only takes about 10ms - 15ms.
 - Transitions are now a bit slower (likely due to a bug in the code and therefore an opportunity) at about 120ms as opposed to 95ms previously. There does seem to be an unnecessary update occurring that I'll investigate in a follow up PR.

### DIFF
--- a/src/components/CalendarDay.jsx
+++ b/src/components/CalendarDay.jsx
@@ -16,7 +16,7 @@ const propTypes = forbidExtraProps({
   day: momentPropTypes.momentObj,
   daySize: nonNegativeInteger,
   isOutsideDay: PropTypes.bool,
-  modifiers: PropTypes.object,
+  modifiers: PropTypes.instanceOf(Set),
   isFocused: PropTypes.bool,
   tabIndex: PropTypes.oneOf([0, -1]),
   onDayClick: PropTypes.func,
@@ -32,7 +32,7 @@ const defaultProps = {
   day: moment(),
   daySize: DAY_SIZE,
   isOutsideDay: false,
-  modifiers: {},
+  modifiers: new Set(),
   isFocused: false,
   tabIndex: -1,
   onDayClick() {},
@@ -43,10 +43,6 @@ const defaultProps = {
   // internationalization
   phrases: CalendarDayPhrases,
 };
-
-export function getModifiersForDay(modifiers, day) {
-  return day ? Object.keys(modifiers).filter(key => modifiers[key](day)) : [];
-}
 
 export default class CalendarDay extends React.Component {
   shouldComponentUpdate(nextProps, nextState) {
@@ -93,12 +89,9 @@ export default class CalendarDay extends React.Component {
 
     if (!day) return <td />;
 
-    const modifiersForDay = getModifiersForDay(modifiers, day);
-
     const className = cx('CalendarDay', {
       'CalendarDay--outside': isOutsideDay,
-    }, modifiersForDay.map(mod => `CalendarDay--${mod}`));
-
+    }, Array.from(modifiers, mod => `CalendarDay--${mod}`));
 
     const formattedDate = `${day.format('dddd')}, ${day.format('LL')}`;
 
--- a/src/components/DayPickerRangeController.jsx
+++ b/src/components/DayPickerRangeController.jsx
@@ -12,6 +12,14 @@ import isTouchDevice from '../utils/isTouchDevice';
 import isInclusivelyAfterDay from '../utils/isInclusivelyAfterDay';
 import isNextDay from '../utils/isNextDay';
 import isSameDay from '../utils/isSameDay';
+import isAfterDay from '../utils/isAfterDay';
+import isBeforeDay from '../utils/isBeforeDay';
+
+import getVisibleDays from '../utils/getVisibleDays';
+import isDayVisible from '../utils/isDayVisible';
+
+import toISODateString from '../utils/toISODateString';
+import toISOMonthString from '../utils/toISOMonthString';
 
 import FocusedInputShape from '../shapes/FocusedInputShape';
 import ScrollableOrientationShape from '../shapes/ScrollableOrientationShape';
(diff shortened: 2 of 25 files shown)


src/components/CalendarDay.jsx
```diff
@@ -16,7 +16,7 @@ const propTypes = forbidExtraProps({
   day: momentPropTypes.momentObj,
   daySize: nonNegativeInteger,
   isOutsideDay: PropTypes.bool,
-  modifiers: PropTypes.object,
+  modifiers: PropTypes.instanceOf(Set),
```
ljharb: it's worth noting that this may make the PR semver-major, as this adds a requirement for es6-shimmed or above, and breaks IE <= 8 (if it wasn't already broken)

src/components/CalendarDay.jsx
```diff
@@ -93,12 +89,9 @@ export default class CalendarDay extends React.Component {
 
     if (!day) return <td />;
 
-    const modifiersForDay = getModifiersForDay(modifiers, day);
-
     const className = cx('CalendarDay', {
       'CalendarDay--outside': isOutsideDay,
-    }, modifiersForDay.map(mod => `CalendarDay--${mod}`));
-
+    }, [...modifiers].map(mod => `CalendarDay--${mod}`));
```
ljharb: `Array.from(modifiers, mod => `CalendarDay--${mod}`)` is much much more efficient because it avoids the intervening array.

src/components/CalendarDay.jsx
majapw: This PR is already semver-major due to the fact that the `DayPicker` now functions totally differently. Do you think this is a concern? I kind of think this project was not working with IE8 already.

src/components/CalendarDay.jsx
majapw: okeedoke. Didn't know that syntax!

src/components/CalendarDay.jsx
ljharb: As long as the PR is already semver-major, and as long as we ensure that these requirements (`Set` and `Array.from` being globally polyfilled) are documented in the readme, I'm a solid 👍 on it.

src/utils/isAfterDay.js
```diff
@@ -0,0 +1,12 @@
+import moment from 'moment';
+
+export default function isAfterDay(a, b) {
+  if (!moment.isMoment(a) || !moment.isMoment(b)) return false;
+
+  const isSameYear = a.year() === b.year();
```
lencioni: You might be able to squeeze out a bit more perf by storing the results of x.year/month/date, depending on how expensive those calls are and how often this function is called.

src/utils/isDayVisible.js
```diff
@@ -0,0 +1,7 @@
+// TODO(maja): get this working for enableOutsideDays
+export default function isDayVisible(day, month, numberOfMonths) {
+  const firstDayOfFirstMonth = month.clone().startOf('month');
+  const lastDayOfLastMonth = month.clone().add(numberOfMonths - 1, 'months').endOf('month');
+
+  return !day.isBefore(firstDayOfFirstMonth) && !day.isAfter(lastDayOfLastMonth);
```
lencioni: Should this use your isBeforeDay/isAfterDay utility methods?

src/utils/isDayVisible.js
```diff
@@ -0,0 +1,7 @@
+// TODO(maja): get this working for enableOutsideDays
+export default function isDayVisible(day, month, numberOfMonths) {
+  const firstDayOfFirstMonth = month.clone().startOf('month');
+  const lastDayOfLastMonth = month.clone().add(numberOfMonths - 1, 'months').endOf('month');
```
lencioni: You can avoid this work some of the time by doing an early return and putting this after it.

src/components/SingleDatePicker.jsx
```diff
@@ -84,23 +90,46 @@ const defaultProps = {
...
+      // TODO: update these if they change
+      'blocked-calendar': day => props.isDayBlocked(day),
+      'blocked-out-of-range': day => props.isOutsideRange(day),
+      'highlighted-calendar': day => props.isDayHighlighted(day),
+      valid: day => !this.isBlocked(day),
+      hovered: day => this.isHovered(day),
+      selected: day => this.isSelected(day),
+    };
+
+    const initialVisibleMonthThunk = props.initialVisibleMonth || (() => (props.date || moment()));
```
ljharb: this could be more efficient with `props.initialVisibleMonth || (props.date ? () => props.date : () => moment())` (and even more so if the moment thunk was cached at module level instead of recreated every render)

src/components/SingleDatePicker.jsx
```diff
@@ -84,23 +90,46 @@ const defaultProps = {
...
+    const evaluatedMonth = initialVisibleMonthThunk();
+    const currentMonth = moment.isMoment(evaluatedMonth) ? evaluatedMonth : moment();
+    const visibleDays = getVisibleDays(currentMonth, props.numberOfMonths, props.enableOutsideDays);
```
ljharb: also, all this prop-dependent logic needs to be in both the constructor and `componentWillReceiveProps`. Rather than duplicating it, is there some chunk of it that could be in a shared function?

src/components/SingleDatePicker.jsx
```diff
@@ -188,14 +273,35 @@ export default class SingleDatePicker extends React.Component {
...
+      },
     });
   }
 
   onDayMouseLeave() {
+    const { hoverDate, visibleDays } = this.state;
+    if (this.isTouchDevice || !hoverDate) return;
+
+    let modifiers = {};
+    modifiers = this.deleteModifier(modifiers, hoverDate, 'hovered');
```
ljharb: `const modifiers = this.deleteModifier({}, hoverDate, 'hovered');`?

altho this looks kind of weird, why the empty object?

src/utils/isAfterDay.js
```diff
@@ -0,0 +1,18 @@
+import moment from 'moment';
+
+export default function isAfterDay(a, b) {
```
ljharb: it seems unfortunate to have both isAfterDay and isBeforeDay - would it be worth it to make one of the implementations (say this one) into `return !isBeforeDay(a, b) && !isSameDay(a, b);`?

src/utils/isDayVisible.js
```diff
@@ -0,0 +1,13 @@
+import isBeforeDay from './isBeforeDay';
+import isAfterDay from './isAfterDay';
+
+// TODO(maja): get this working for enableOutsideDays
+export default function isDayVisible(day, month, numberOfMonths) {
+  const firstDayOfFirstMonth = month.clone().startOf('month');
+  if (isBeforeDay(day, firstDayOfFirstMonth)) return false;
+
+  const lastDayOfLastMonth = month.clone().add(numberOfMonths - 1, 'months').endOf('month');
+  if (isAfterDay(day, lastDayOfLastMonth)) return false;
```
ljharb: `return !isAfterDay(day, lastDayOfLastMonth);`?

src/utils/isInclusivelyBeforeDay.js
```diff
@@ -1,8 +1,9 @@
 import moment from 'moment';
 
 import isSameDay from './isSameDay';
+import isBeforeDay from './isBeforeDay';
 
 export default function isInclusivelyBeforeDay(a, b) {
   if (!moment.isMoment(a) || !moment.isMoment(b)) return false;
-  return a.isBefore(b) || isSameDay(a, b);
+  return isBeforeDay(a, b) || isSameDay(a, b);
```
ljharb: isn't this `return !isAfterDay(a, b);`?

src/utils/isInclusivelyAfterDay.js
```diff
@@ -1,8 +1,9 @@
 import moment from 'moment';
 
 import isSameDay from './isSameDay';
+import isAfterDay from './isAfterDay';
 
 export default function isInclusivelyAfterDay(a, b) {
   if (!moment.isMoment(a) || !moment.isMoment(b)) return false;
-  return a.isAfter(b) || isSameDay(a, b);
+  return isAfterDay(a, b) || isSameDay(a, b);
```
ljharb: isn't this `return !isBeforeDay(a, b);`?

src/components/SingleDatePicker.jsx
majapw: I think there is a problem with the `this.props`/`props` paradigm. Also do you mean that this needs to be reset when the month changes?

src/components/SingleDatePicker.jsx
majapw: The empty object is so that when we call `addModifier`/`deleteModifier` the effects stack rather than overwriting each other. So for instance if something happens that causes two different modifiers to be added, we spread everything onto this object and then set the state at the end.

src/utils/isAfterDay.js
majapw: we could also probs update usage to only have one. I just thought it was a bit clear to use one over the other in some cases.

src/components/SingleDatePicker.jsx
ljharb: What I mean is, any prop-dependent logic in any constructor needs to be re-executed in `componentWillReceiveProps` - so in this case, yes, when `numberOfMonths` or `enableOutsideDays` changes, `visibleDays` and thus `this.state.visibleDays` would need to be recomputed.

majapw:
@ljharb I think I addressed your concerns! Can you take a look?

src/utils/isAfterDay.js
ljharb: I don't mean, let's get rid of both - I mean, let's implement one in terms of the other.

src/components/SingleDatePicker.jsx
ljharb: Thanks, that makes sense.

src/components/DayPickerRangeController.jsx
```diff
@@ -259,12 +639,13 @@ export default class DayPickerRangeController extends React.Component {
   }
 
   isHovered(day) {
-    return isSameDay(day, this.state.hoverDate);
+    const { hoverDate } = this.state || {};
```
ljharb: when would the state not be initialized to an empty object?

src/components/DayPickerRangeController.jsx
```diff
@@ -259,12 +639,13 @@ export default class DayPickerRangeController extends React.Component {
...
+    return isSameDay(day, hoverDate);
   }
 
   isInHoveredSpan(day) {
     const { startDate, endDate } = this.props;
-    const { hoverDate } = this.state;
+    const { hoverDate } = this.state || {};
```
ljharb: when would the state not be initialized to an empty object?

src/components/DayPickerRangeController.jsx
```diff
@@ -249,7 +629,7 @@ export default class DayPickerRangeController extends React.Component {
 
   isDayAfterHoveredStartDate(day) {
     const { startDate, endDate, minimumNights } = this.props;
-    const { hoverDate } = this.state;
+    const { hoverDate } = this.state || {};
```
ljharb: when would the state not be initialized to an empty object?

src/components/SingleDatePicker.jsx
```diff
@@ -284,18 +460,118 @@ export default class SingleDatePicker extends React.Component {
...
+        month = updatedDays[monthIso] || visibleDays[monthIso];
+      }
+    }
+
+    const modifiers = new Set(month[iso]);
+    modifiers.delete(modifier);
+    return {
+      ...updatedDays,
+      ...{
+        [monthIso]: {
```
ljharb: this spread isn't necessary - you can just do:
```js
return {
   ...updatedDays,
  [monthISO]: {
    ...month,
    [iso]: modifiers,
  },
};
```

src/components/SingleDatePicker.jsx
```diff
@@ -344,7 +620,8 @@ export default class SingleDatePicker extends React.Component {
   }
 
   isHovered(day) {
-    return isSameDay(day, this.state.hoverDate);
+    const { hoverDate } = this.state || {};
```
ljharb: when would the state not be initialized to an empty object?

majapw: When this function is called from the constructor, `this.state` has not yet been set.

ljharb: :-/ hmm - in that case it kind of seems like the meat of it should be extracted out to a pure function in a closure, and have both the constructor and `isHovered` call into it with the data it needs.

src/components/DayPickerRangeController.jsx
majapw: When this function is called from the constructor, this.state has not yet been set. 

... this same response applies on the other areas too. 

It's a weird edge case, but this seemed like the cleanest solution?

ljharb: gotcha (sorry for the multi-comment).

https://github.com/airbnb/react-dates/pull/450#discussion_r114226407 is what i'd expect to be the solution

src/components/SingleDatePicker.jsx
majapw: can you clarify that a little bit?

do you mean something like `isHovered(day, hoverDate)`? I feel like that might be odd given that it'd suddenly be different from every other modifiers method

ljharb: Hmm - yeah, i can see how that'd be unnecessarily weird. I guess this is fine as-is.

dburles:
Hey guys, I have the following on a `DayPicker` component in order to highlight a range of dates that contain information, along with highlighting the currently selected day. However it now no longer works after this refactor. Is there another approach I can take to make it work? thanks.

```
            modifiers={{
              selected: day =>
                moment(this.props.selectedDay).isSame(day, 'day'),
              'highlighted-calendar': day1 =>
                this.props.highlightedDays.some(day2 =>
                  moment(day2).isSame(day1, 'day'),
                ),
            }}
```

majapw:
Hey @dburles, can you open an issue for this so that we can have a convo on a separate thread?

dburles:
@majapw sure thing! #562

PR #455, opened 2017-04-18
Fix moment peer dependency to allow 2.18.x patch releases

ConAntonakos:
Fix the `peerDependencies` for the `moment` package to prevent warnings from being emitted by package managers.


ConAntonakos:
Ah, I apologize. Thanks for the update, @ljharb!

PR #473, opened 2017-04-26
Add the props and the mandatory calls, resolve #94

### LINKED ISSUE #94: SingleDatePicker Doesnt accept onDayMouseEnter param
The SingleDatePicker component doesnt accept onDayMouseEnter param to pass along to the DayPicker component it creates.

```
onDayMouseEnter(day, modifiers, e) {
    this.setState({
      hoverDate: day,
    });
    this.props.onDayMouseEnter(day, modifiers, e);
}
```

instead of the current:

```
onDayMouseEnter(day) {
    this.setState({
      hoverDate: day,
    });
}
```

kane-thornwyrd:


majapw:
I don't believe this is a functionality that we want to introduce to the `DateRangePicker`. It very intentionally does not expose the day interaction methods. We can discuss more on #94, but I'm going to close this for now.

Issue #427, opened 2017-04-07
Keyboard Accessibility help modal broken when only 1 month is visible


emmenko:
Version: `react-dates@10.0.0`

![image](https://cloud.githubusercontent.com/assets/1110551/24788983/f2c18d94-1b70-11e7-9424-e1b94f46c2d8.png)

emmenko:
Also, is there a way to **not show** the `?` button?

ConAntonakos:
I noticed this as well.

![screen shot 2017-04-10 at 3 34 58 pm](https://cloud.githubusercontent.com/assets/1504302/24879121/c5155e9e-1e03-11e7-8c74-912655288dd0.png)

Should we add `overflow: auto` to the `.DayPickerKeyboardShortcuts__panel` class?

ConAntonakos:
Also, there seems to be some clipping for `.DayPickerKeyboardShortcuts__show--bottom-right`?

![screen shot 2017-04-10 at 4 10 55 pm](https://cloud.githubusercontent.com/assets/1504302/24880535/c8c26c44-1e08-11e7-9a01-2b4a605214a5.png)

Ostensibly from this rule: `border-top: 26px solid #fff`. Maybe make it `transparent` instead?

majapw:
Huzzah! Totally not surprised that these things broke (it was a very big PR) and very happy to accept PRs to fix them. 

Quick stop-gap to hide the `?` would be just to hide it in CSS (I think it's `.DayPickerKeyboardShortcuts__show`?), but yeah we should probs fix the panel. :)

atkinchris:
We hastily patched this with `.DayPickerKeyboardShortcuts__panel {  position: fixed; }`. Not perfect, but functional.

dburles:
Can see the behaviour here: http://airbnb.io/react-dates/?selectedKind=DayPickerRangeController&selectedStory=single%20month&full=0&down=1&left=1&panelRight=0&downPanel=kadirahq%2Fstorybook-addon-actions%2Factions-panel

majapw:
Ping! Sorry for the delay, but here's a PR - https://github.com/airbnb/react-dates/pull/477

emmenko:
Cool, thanks for addressing this 👍

PR #477, opened 2017-04-27
Address some issues introduced by the a11y PR

majapw:
Sorry this took so long! This should address issues raised in https://github.com/airbnb/react-dates/issues/427.

Namely, a handful of things happen in this PR:
1. The keyboard shortcuts panel button is now transparent instead of white. It also has a higher z-index so it doesn't flash in transitions.
2. The panel has an overflow of auto so it doesn't totally crap out on single month datepickers
3. The alignment of text in the panel has been adjusted a bit to support the single month case.
4. `enableOutsideDays` was having some weirdness with using the keyboard to transition months. This has been fixed.
5. There is now a prop to hide the panel entirely. The recommendation is still to either style the panel to best suit your use-case or to communicate the keyboard shortcuts elsewhere, but yaknow.

to: @airbnb/webinfra

### DIFF
--- a/css/DayPickerKeyboardShortcuts.scss
+++ b/css/DayPickerKeyboardShortcuts.scss
@@ -17,10 +17,11 @@
 .DayPickerKeyboardShortcuts__show {
   width: 22px;
   position: absolute;
+  z-index: $react-dates-z-index + 2;
 }
 
 .DayPickerKeyboardShortcuts__show--bottom-right {
-  border-top: 26px solid $react-dates-color-white;
+  border-top: 26px solid transparent;
   border-right: 33px solid $react-dates-color-primary;
   bottom: 0;
   right: 0;
@@ -36,7 +37,7 @@
 }
 
 .DayPickerKeyboardShortcuts__show--top-right {
-  border-bottom: 26px solid $react-dates-color-white;
+  border-bottom: 26px solid transparent;
   border-right: 33px solid $react-dates-color-primary;
   top: 0;
   right: 0;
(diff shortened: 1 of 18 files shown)


css/DayPickerKeyboardShortcuts.scss
```diff
@@ -17,10 +17,11 @@
 .DayPickerKeyboardShortcuts__show {
   width: 22px;
   position: absolute;
+  z-index: 2;
```
lencioni: Why 2 here? Can this be based off of a variable to make the dependency more explicit?

majapw: good idea!

sag1v:
@majapw Great work, thank you.

> There is now a prop to hide the panel entirely 

What is the name of the `prop`? (not that i will use it but yaknow lol)

majapw:
`hideKeyboardShortcutsPanel` on both the DRP and the SDP

majapw:
@ljharb while I believe we should provide a reasonable default option for explaining the keyboard shortcuts (which we do, with the panel), I think that because we are providing a tool for developers to integrate into their own products, we should allow for them to customize how they decide to convey this information to their users, whether it is embedded on the page directly, in a modal separate from the datepicker completely, or something I haven't yet considered. We should not force developers to use our solution, and that is why I added the `hideKeyboardShortcutsPanel` option. 

My first recommendation would still be to use css overrides to style the panel to fit your use case, but I want to provide an option to do your own thing as well. This is why I think this prop makes a lot of sense.

ljharb:
I'm still not a fan of hiding the shortcut panel, but after some offline discussion with @majapw I'm definitely not a blocker.

kbariotis:
Aw! I came too late! :( Shouldn't README be updated too with the newly added `hideKeyboardShortcutsPanel` flag?

Issue #479, opened 2017-04-28
DayPickerRangeController still displaying hover when blurred


chris-verclytte:
Currently, the behaviour of the DayPickerRangeController is to hover days even when it has been blurred.
I would have expected it to be hoverable only when it has focus i.e. on [isHovered modifier](url) (https://github.com/airbnb/react-dates/blob/5b54d63c00fd18ddb4bfcef6a56fb297c1fd47f9/src/components/DayPickerRangeController.jsx#L261):
```javascript 
 isHovered(day) {
    const { focusedInput } = this.props;
    return (focusedInput === START_DATE || focusedInput === END_DATE) && isSameDay(day, this.state.hoverDate);
  }
```
Does this make sense ?

majapw:
That seems reasonable to me! We primarily use the full `DateRangePicker` component at airbnb so we often don't catch little edge cases like this. :) Would be happy to accept a PR for this, although maybe it could just be:
```
isHovered(day) {
    const { focusedInput } = this.props;
    return focusedInput && isSameDay(day, this.state.hoverDate);
  }
```
since if `focusedInput` is truthy, then the `DayPickerRangeController` should be interactive.

PR #483, opened 2017-05-03
Disable hover when DatePicker does not have focus

chris-verclytte:
Adresses #479.

Usefull for DatePicker "resetting" focusedInput when clicking outside of the component.
It avoids DatePicker from showing hover style on day when hovered if the DayPicker does not have
focus.

### DIFF
--- a/src/components/DayPickerRangeController.jsx
+++ b/src/components/DayPickerRangeController.jsx
@@ -382,52 +382,54 @@ export default class DayPickerRangeController extends React.Component {
     const { startDate, endDate, focusedInput, minimumNights } = this.props;
     const { hoverDate, visibleDays } = this.state;
 
-    let modifiers = {};
-    modifiers = this.deleteModifier(modifiers, hoverDate, 'hovered');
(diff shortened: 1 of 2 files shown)


src/components/DayPickerRangeController.jsx
```diff
@@ -642,7 +642,8 @@ export default class DayPickerRangeController extends React.Component {
 
   isHovered(day) {
     const { hoverDate } = this.state || {};
-    return isSameDay(day, hoverDate);
+    const { focusedInput } = this.props;
+    return focusedInput && isSameDay(day, hoverDate);
```
majapw: I think with the new perf rearchitecture in v11.0.0, we will also have to wrap https://github.com/chris-verclytte/react-dates/blob/c5e7fae7986518dec6cb4e2438d5f0a402d15605/src/components/DayPickerRangeController.jsx#L400-L425 in a `if (focusedInput)` conditional in this PR.

chris-verclytte:
Thanks for this one, I did not found the time to do it this week !

Issue #497, opened 2017-05-08
Moment version


jgwconsulting:
I was just wondering what the issue is with `moment@2.18.1` which doesn't play well with react-dates? Thanks!

majapw:
I think it broke the tests, and we just haven't gotten around to updating things yet! We will probably do this shortly, but other work has been higher priority.

jgwconsulting:
Thanks Maja, and thanks for all your hard work on this component - it's excellent.

PR #505, opened 2017-05-11
Bump moment and remove locale tests

majapw:
Fix for #497 

We were pegging moment to a lower version because a number of our date conversion tests were testing for format explicitly and moment frequently updates locale formats in minor and patch releases. This updates moment and removes the offending tests.

The tests are a bit of a relic of when this repository was internal to Airbnb. At that time, the formatted string values were being passed around and had to be in a specific format or face breaking pages. Now that there's a bit of healthy separation (the input elements themselves use ISO format and the callbacks from the components all pass around moment objects instead of strings) the locale tests don't do as much for us within the repo itself. :) As such, just knowing the `toLocalizedDateString`, `toMomentObject`, and `toISODateString` all handle the `L` format appropriately is enough. 

to: @ljharb @jasonkb @airbnb/webinfra


package.json
```diff
@@ -113,7 +113,7 @@
     "react-portal": "^3.0.0"
   },
   "peerDependencies": {
-    "moment": "2.10 - 2.14 || ~2.15.1 || ~2.16 || ~2.17",
+    "moment": "^2.18.1",
```
ljharb: to be a patch, this needs to be `2.10 - 2.14 || ~2.15.1 || ~2.16 || ~2.17 || ^2.18.1`

Issue #504, opened 2017-05-10
Add example for `readOnly` prop


ljharb:
See https://github.com/airbnb/react-dates/issues/496#issuecomment-300615526

b0gok:
@ljharb I think you can close this issue (:

ljharb:
Closed in #512.

PR #512, opened 2017-05-15
Add examples with disabled prop and readOnly to storybook

b0gok:
According to issue https://github.com/airbnb/react-dates/issues/504


majapw:
Awesome! :D Thank you!

b0gok:
@ljharb @majapw thank you for this awesome package (:

PR #516, opened 2017-05-17
[Fix] isTouchDevice: improve detection logic

### LINKED ISSUE #515: Hover on a day on Firefox doesn't add classes to the day
When I hover on the calendar on firefox nothing happens. After debugging a little bit, I think there is a bug in the `isTouchDevice` function. 

`ontouchstart` is defined in firefox, but it has a `null` value.

screen shot 2017-05-16 at 5 35 09 pm

Since the `onDayMouseEnter` and `onDayMouseLeave` will return if `this.isTouchDevice`, the hover logic doesn't get applied on Firefox.

tagoro9:
`window.ontouchstart` is a defiend property in Firefox with a value of null. This PR fixes the `isTouchDevice` function to check that `ontouchstart != null`.

This PR fixes #515.

### DIFF
--- a/src/utils/isTouchDevice.js
+++ b/src/utils/isTouchDevice.js
@@ -1,4 +1,11 @@
 export default function isTouchDevice() {
-  return !!(typeof window !== 'undefined' && 'ontouchstart' in window) ||
-    !!(typeof navigator !== 'undefined' && navigator.maxTouchPoints);
+  return (
+    !!(typeof window !== 'undefined' &&
+      ('ontouchstart' in window ||
+        (window.DocumentTouch &&
+          typeof document !== 'undefined' &&
+          document instanceof window.DocumentTouch))) ||
+    !!(typeof navigator !== 'undefined' &&
+      (navigator.maxTouchPoints || navigator.msMaxTouchPoints))
+  );
 }
--- a/test/utils/isTouchDevice_spec.js
+++ b/test/utils/isTouchDevice_spec.js
@@ -4,6 +4,7 @@ import wrap from 'mocha-wrap';
 import isTouchDevice from '../../src/utils/isTouchDevice';
 
 const describeIfNoWindow = typeof window === 'undefined' ? describe : describe.skip;
+function DocumentTouch() { }
 
 describeIfNoWindow('isTouchDevice', () => {
   describe('with neither `window` nor `navigator`', () => {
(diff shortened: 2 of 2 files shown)


src/utils/isTouchDevice.js
```diff
@@ -1,4 +1,4 @@
 export default function isTouchDevice() {
-  return !!(typeof window !== 'undefined' && 'ontouchstart' in window) ||
+  return !!(typeof window !== 'undefined' && window.ontouchstart != null) ||
```
ljharb: What value will `window.ontouchstart` have in a browser that supports touch?

However, https://hacks.mozilla.org/2013/04/detecting-touch-its-the-why-not-the-how/ seems to imply that we should be checking a few more things.

tagoro9: By looking at what Modernizr [does](https://github.com/Modernizr/Modernizr/blob/5eea7e2a213edc9e83a47b6414d0250468d83471/feature-detects/touchevents.js), it looks like they do some more checks too. Even though it is not still a bug I could still update this to have a better check.

majapw: I think that would be super helpful!

ljharb: Agreed, that'd be appreciated :-)

tagoro9:
I updated the `isTouchDevice` function to also check for mxMaxTouchPoints, which seems to be used on some old IE versions and to also use the `DocumentTouch` check, which is the one used by `Modernizr`.

They also use some media queries, but that involves adding a lot of vendor prefixes:

https://drafts.csswg.org/mediaqueries/#pointer
https://www.quirksmode.org/css/mediaqueries/touch.html

These two articles seemed relevant:

http://www.stucox.com/blog/you-cant-detect-a-touchscreen/
https://medium.com/@david.gilbertson/the-only-way-to-detect-touch-with-javascript-7791a3346685 

Let me know what you think of this approach.

test/utils/isTouchDevice_spec.js
```diff
@@ -26,6 +27,20 @@ describeIfNoWindow('isTouchDevice', () => {
...
+    .withOverride(() => window, 'DocumentTouch', () => DocumentTouch)
+    .withGlobal('document', () => new (function Document() {})())
+    .it('returns false when document is not an instance of DocumentTouch', () => {
+      expect(isTouchDevice()).to.equal(false);
+    });
+
+    wrap()
+    .withOverride(() => window, 'DocumentTouch', () => DocumentTouch)
+    .withGlobal('document', () => new DocumentTouch())
+    .it('returns true when document is an instance of DocumentTouch', () => {
```
ljharb: might be nice to assert that document is an instance of DocumentTouch, as well

test/utils/isTouchDevice_spec.js
```diff
@@ -26,6 +27,20 @@ describeIfNoWindow('isTouchDevice', () => {
     .it('returns true with `window.ontouchstart', () => {
       expect(isTouchDevice()).to.equal(true);
     });
+
+    wrap()
+    .withOverride(() => window, 'DocumentTouch', () => DocumentTouch)
+    .withGlobal('document', () => new (function Document() {})())
+    .it('returns false when document is not an instance of DocumentTouch', () => {
```
tagoro9: Added.

test/utils/isTouchDevice_spec.js
tagoro9: Added.

src/utils/isTouchDevice.js
```diff
@@ -1,4 +1,11 @@
 export default function isTouchDevice() {
-  return !!(typeof window !== 'undefined' && 'ontouchstart' in window) ||
-    !!(typeof navigator !== 'undefined' && navigator.maxTouchPoints);
+  return (
+    !!(typeof window !== 'undefined' &&
```
ljharb: This logic could probably be split up into if/elses at this point, but nbd

PR #519, opened 2017-05-17
Bump react-portal to v3.1.0

### LINKED ISSUE #518: PropTypes deprecation warning
I was getting the warning  `Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead.` when consuming `react-timelines`.

After some digging around on the stack trace I've narrowed down the error to be caused by the `react-portal` dependency.
screen shot 2017-05-17 at 15 20 55

It seems the `react-portal` maintainers have recently bumped their version to `3.1.0` in turn adding support for React `15.5` while `react-dates` is still has a dependency on `3.0.0` I can issue a PR for this if that is of any help?

rbrtsmith:
* Fixes #518

### DIFF
--- a/package.json
+++ b/package.json
@@ -110,7 +110,7 @@
     "object.values": "^1.0.4",
     "prop-types": "^15.5.6",
     "react-moment-proptypes": "^1.3.0",
-    "react-portal": "^3.0.0"
+    "react-portal": "^3.1.0"
   },
   "peerDependencies": {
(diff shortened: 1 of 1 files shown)


majapw:
Also what is the issue with v3.1.0?

rbrtsmith:
@majapw That was a mistake in my part.  I meant to bump to `v3.1.0`  Thanks.

Issue #469, opened 2017-04-24
`numberOfMonths` propType ignored in `DateRangePicker` and `SingleDatePicker` when orientation is set to `vertical`


byrondevonwall:
When trying to set the `numberOfMonths` propType on either the single or range picker, the pickers will display 2 months no matter how many months are specified in the prop when the orientation is set to `vertical`.

The prop works as expected when orientation is set to `horizontal`.

ljharb:
[This line](https://github.com/airbnb/react-dates/blob/master/src/components/DayPicker.jsx#L684) indeed explicitly only considers `numberOfMonths` when the orientation is not vertical.

Perhaps @majapw can weigh in on whether this is intentional; if so, then the propType should probably error when vertical and provided; if not, let's fix it.

byrondevonwall:
@ljharb Yeah, it's not that big of a deal, I was just kind of banging my head against the wall trying to get it to work for some number of minutes before switching the orientation.  If it's intentional it would be nice if it were in the docs somewhere is all.  Thanks for the response!

majapw:
In general, the `vertical` orientation is slightly less well-developed than the horizontal orientation so... this is something that should be fixed.

PR #521, opened 2017-05-17
verticalHeight depends on the numberOfMonths prop in DayPicker

b0gok:
Related to https://github.com/airbnb/react-dates/issues/469
It's not the best solution for this case, but I think it's better then nothing. Check this, please.

### DIFF
--- a/src/components/DayPicker.jsx
+++ b/src/components/DayPicker.jsx
@@ -721,9 +721,8 @@ export default class DayPicker extends React.Component {
       phrases,
     } = this.props;
 
-    const numOfWeekHeaders = this.isVertical() ? 1 : numberOfMonths;
     const weekHeaders = [];
-    for (let i = 0; i < numOfWeekHeaders; i += 1) {
+    for (let i = 0; i < numberOfMonths; i += 1) {
       weekHeaders.push(this.renderWeekHeader(i));
     }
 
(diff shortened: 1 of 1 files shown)


src/components/DayPicker.jsx
```diff
@@ -721,9 +721,8 @@ export default class DayPicker extends React.Component {
       phrases,
     } = this.props;
 
-    const numOfWeekHeaders = this.isVertical() ? 1 : numberOfMonths;
     const weekHeaders = [];
-    for (let i = 0; i < numOfWeekHeaders; i += 1) {
+    for (let i = 0; i < numberOfMonths; i += 1) {
```
majapw: Wait there should def still only be one week header.

src/components/DayPicker.jsx
```diff
@@ -751,9 +750,7 @@ export default class DayPicker extends React.Component {
 
     const horizontalWidth = (calendarMonthWidth * numberOfMonths) + (2 * DAY_PICKER_PADDING);
 
-    // this is a kind of made-up value that generally looks good. we'll
-    // probably want to let the user set this explicitly.
-    const verticalHeight = 1.75 * calendarMonthWidth;
+    const verticalHeight = (calendarMonthWidth * numberOfMonths) + (2 * MONTH_PADDING);
```
majapw: We should still probably leave a comment as to the effect of this number, but yes, I agree this is better.

ljharb:
@b0gok are you not able to reply to the review and update the PR?

b0gok:
@ljharb I decided that this is an outdated problem, but I did not check. If it is relevant, I can update the PR.

ljharb:
Since the issue is still open, it still seems relevant to me.

Issue #517, opened 2017-05-17
visibleDays  should be recalculated after focusedInput change


Masalovych:
Hi,
I am using DateRangePicker component. And I need to setup isOutsideRange function for each input separately. For example for startDate only available days should be monday and for endDate only sunday. To do this I am using focusedInput variable in my isOutsideRange function.

The problem that available days don't changed after focusedInput changed. 
I think condition in the file react-dates/src/components/DayPickerRangeController.jsx:201 should be changed that recalculate "newMonthState" when focusedInput changed

When can we expect this issue be fixed?

Best Regards,
Dmytro
dmasalovych@gmail.com

timhwang21:
@Masalovych fixed in #522!

PR #522, opened 2017-05-17
Recomputes more modifiers when focusedInput changes

timhwang21:
Previously, only `props.isDayBlocked()` and `props.isDayHighlighted()` are computed. However, in my use case, where there are different validations on the start and end date inputs, `'blocked'` and `'blocked-out-of-range'` persisted when the focused input changes. This PR adds recomputation of modifiers with `this.isBlocked()` and `props.isOutOfRange()`.

I've also added tests; however, it seems like `this.isBlocked()` is getting called for all visible dates instead of only the changed dates. I commented out the failing tests for now (returns 183 calls or something) -- more clarification on that would be appreciated!

Also, if there's any reason I missed that these checks shouldn't be done, please let me know. I tested my fork in my use case and the problem was fixed.

Thanks!

### DIFF
--- a/test/components/DayPickerRangeController_spec.jsx
+++ b/test/components/DayPickerRangeController_spec.jsx
@@ -796,6 +796,140 @@ describe('DayPickerRangeController', () => {
         });
       });
 
+      describe('blocked', () => {
+        describe.skip('focusedInput did not change', () => {
+          it('does not call isBlocked', () => {
+            const isBlockedStub =
+              sinon.stub(DayPickerRangeController.prototype, 'isBlocked');
+            const wrapper = shallow(<DayPickerRangeController {...props} />);
+            wrapper.instance().componentWillReceiveProps({
+              ...props,
+            });
+            expect(isBlockedStub.callCount).to.equal(0);
+          });
+        });
+
+        describe('focusedInput changed', () => {
+          const visibleDays = {
+            [toISOMonthString(today)]: {
+              [toISODateString(today)]: [],
+              [toISODateString(moment().add(1, 'day'))]: [],
+              [toISODateString(moment().add(2, 'day'))]: [],
+            },
+          };
+          const numVisibleDays = 3;
+
+          it.skip('calls isBlocked for every visible day', () => {
+            const isBlockedStub =
+              sinon.stub(DayPickerRangeController.prototype, 'isBlocked');
+            const wrapper = shallow(<DayPickerRangeController {...props} />);
+            wrapper.setState({ visibleDays });
+            wrapper.instance().componentWillReceiveProps({
+              ...props,
+              focusedInput: END_DATE,
+            });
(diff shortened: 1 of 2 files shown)


timhwang21:
This PR is meant to fix #517.

majapw:
Woo! Thanks for adding this! I will look over this tonight or tomorrow!

test/components/DayPickerRangeController_spec.jsx
```diff
@@ -796,6 +796,140 @@ describe('DayPickerRangeController', () => {
         });
       });
 
+      describe('blocked', () => {
+        describe.skip('focusedInput did not change', () => {
```
timhwang21: This is one of the two tests I referenced in my commit message. I wasn't sure if it was appropriate to leave them out, or if it getting called 183 times actually isn't expected.

majapw: So you probably have to reset the stub after the initial shallow render. The first mount is going to call this method a bunch of times! You want to isolate it to make sure that you're only checking how many times `componentWillReceiveProps` is calling it!

test/components/DayPickerRangeController_spec.jsx
```diff
@@ -796,6 +796,142 @@ describe('DayPickerRangeController', () => {
...
+              sinon.stub(DayPickerRangeController.prototype, 'isBlocked').returns(true);
+            const wrapper = shallow(<DayPickerRangeController {...props} />);
+            wrapper.setState({ visibleDays });
+            wrapper.instance().componentWillReceiveProps({
+              ...props,
+              focusedInput: START_DATE,
+            });
+            const blockedCalendarCalls = getCallsByModifier(addModifierSpy, 'blocked');
+            expect(blockedCalendarCalls.length).to.equal(numVisibleDays);
+            isBlockedStub.restore();
```
timhwang21: @majapw thanks for the tip re: resetting stubs! However, in this test and the following one, I need to make a stub that returns `true` or `false`. However, I don't ever need to access the stub, as `blockedCalendarCalls` is what's used to verify the test.

I realized I was mistakenly passing the stub in with the props in my previous commit (doing nothing) but after removing it I started getting a `no-unused-vars` lint error. So I shoehorned `isBlockedStub.restore()` in. Would this be an acceptable place for `// eslint-ignore-line no-unused-vars`?

Thanks!

ljharb: You don't need the restore - just don't assign the stub to a variable. In other words, `sinon.stub(foo, bar);` is fine by itself.

timhwang21: Thank you, fixed

timhwang21:
Awesome, thanks!

dvmorris:
Any idea when this will be released? I ran into this issue today.

PR #550, opened 2017-06-08
SingleDatePicker now re-evaluates isOutsideRange on componentWillReceiveProps

### LINKED ISSUE #549: isOutsideRange not re-evaluated in SingleDatePicker on render
Due to various factors, we are needing to use two SingleDatePickers instead of a DateRangePicker in our project (e.g., allowing open-ended selections). The problem we've run into is that when changing the start date in one picker, the second date picker won't re-evaluate `isOutsideRange` when gaining focus again, meaning that if the user moved the start date earlier they still wouldn't be able to pick an earlier end date until shifting between months and forcing a re-render.

Preparing a PR to address this.

jlyman:
Fixes #549 by adding a new call to `isOutsideRange` on `componentWillReceiveProps`, just like what is being done for `isDayBlocked` and `isDayHighlighted`.

### DIFF
--- a/src/components/SingleDatePicker.jsx
+++ b/src/components/SingleDatePicker.jsx
@@ -218,6 +218,12 @@ export default class SingleDatePicker extends React.Component {
       values(visibleDays).forEach((days) => {
         Object.keys(days).forEach((day) => {
           const momentObj = moment(day);
+          if (this.isBlocked(momentObj)) {
+            modifiers = this.addModifier(modifiers, momentObj, 'blocked');
+          } else {
+            modifiers = this.deleteModifier(modifiers, momentObj, 'blocked');
+          }
+
           if (isOutsideRange(momentObj)) {
             modifiers = this.addModifier(modifiers, momentObj, 'blocked-out-of-range');
           } else {
--- a/src/components/SingleDatePicker.jsx
+++ b/src/components/SingleDatePicker.jsx
@@ -218,6 +218,12 @@ export default class SingleDatePicker extends React.Component {
       values(visibleDays).forEach((days) => {
         Object.keys(days).forEach((day) => {
           const momentObj = moment(day);
+          if (isOutsideRange(momentObj)) {
+            modifiers = this.addModifier(modifiers, momentObj, 'blocked-out-of-range');
+          } else {
+            modifiers = this.deleteModifier(modifiers, momentObj, 'blocked-out-of-range');
+          }
+
           if (isDayBlocked(momentObj)) {
             modifiers = this.addModifier(modifiers, momentObj, 'blocked-calendar');
           } else {
--- a/test/components/SingleDatePicker_spec.jsx
+++ b/test/components/SingleDatePicker_spec.jsx
@@ -366,6 +366,70 @@ describe('SingleDatePicker', () => {
         });
       });
 
+      describe('blocked', () => {
+        describe('props.focused did not change', () => {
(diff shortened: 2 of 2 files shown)


src/components/SingleDatePicker.jsx
```diff
@@ -218,6 +218,12 @@ export default class SingleDatePicker extends React.Component {
       values(visibleDays).forEach((days) => {
         Object.keys(days).forEach((day) => {
           const momentObj = moment(day);
+          if (isOutsideRange(momentObj)) {
```
timhwang21: This might need to account for `this.isBlocked()` as well, maybe?

jlyman: Could you elaborate for me @timhwang21? There are a couple other checks (`isDayBlocked()` and `isDayHighlighted()` that happen right afterwards, and I believe `this.isBlocked()` checks both `isDayBlocked` and `isOutsideRange` (on [line 628](https://github.com/jlyman/react-dates/blob/master/src/components/SingleDatePicker.jsx#L628)).

timhwang21: i have to admit my memory is a bit fuzzy on this now but I believe isBlocked adds its own modifier on top of the others. So you can successfully remove the modifiers for isDayBlocked and isOutsideRange but still have the calendar day blocked. This is what i remember from my own pr, hopefully one of the maintainers can clarify.

majapw: What @timhwang21 is saying is we should update this component to look more like what's in the `DayPickerRangeController`, ie:
```
if (didFocusChange) {
      values(visibleDays).forEach((days) => {
        Object.keys(days).forEach((day) => {
          const momentObj = moment(day);

          if (this.isBlocked(momentObj)) {
            modifiers = this.addModifier(modifiers, momentObj, 'blocked');
          } else {
            modifiers = this.deleteModifier(modifiers, momentObj, 'blocked');
          }

          if (isOutsideRange(momentObj)) {
            modifiers = this.addModifier(modifiers, momentObj, 'blocked-out-of-range');
          } else {
            modifiers = this.deleteModifier(modifiers, momentObj, 'blocked-out-of-range');
          }

          if (isDayBlocked(momentObj)) {
            modifiers = this.addModifier(modifiers, momentObj, 'blocked-calendar');
          } else {
            modifiers = this.deleteModifier(modifiers, momentObj, 'blocked-calendar');
          }

          if (isDayHighlighted(momentObj)) {
            modifiers = this.addModifier(modifiers, momentObj, 'highlighted-calendar');
          } else {
            modifiers = this.deleteModifier(modifiers, momentObj, 'highlighted-calendar');
          }
        });
      });
    }
```

majapw:
This looks good! Just the one comment on adding the `this.isBlocked` => `--blocked` modifier as well.

src/components/SingleDatePicker.jsx
jlyman: Gotchya... Thanks @majapw and @timhwang21, that makes sense. Let me go ahead and make that improvement.

vebjorni:
Is this feature in the v12.1.2?

jlyman:
@vebjorni I don't believe so, as 12.1.2 was cut seven days ago, and this was merged 4 days ago.

vebjorni:
@jlyman Thanks, that makes sense.

@majapw Any idea when the next version will be released?

Issue #529, opened 2017-05-23
Cannot call method 'querySelector' of null


jherencia:
We are using react-dates in a website and have installed RavenJS to track the errors, it appears this error quite often:

<img width="1008" alt="captura de pantalla 2017-05-23 a las 19 18 02" src="https://cloud.githubusercontent.com/assets/459373/26366786/9d37200a-3fec-11e7-877c-632be84b9158.png">

ljharb:
In what browser do you see that?

jherencia:
I attach the browser statistics:

[PEPECAR-1S3-browser.csv.txt](https://github.com/airbnb/react-dates/files/1023590/PEPECAR-1S3-browser.csv.txt)

ljharb:
Hmm, I'm not sure why `findDOMNode` would return null in those browsers.

johnryan:
@jherencia did you find a solution to this? i'm seeing the same issue.

majapw:
So we can add a check for null in these locations, but it seems weird to me that it's not set in `componentDidMount` or that `React.findDOMNode` doesn't return anything.

johnryan:
@majapw so not sure if this helps but what i'm seeing that's a bit weird is that when using create-react-app it works fine in development. The issue only shows up when it's compiled for production.

ljharb:
@jherencia is `this.calendarMonthGrid` a ref that's being set in, perhaps, an inline arrow callback?

Refs can sometimes be null unless you use a constructor-bound instance method (such that the ref callback's reference is `===` every time). You'll either need to check for null, and/or, not use an inline function.

majapw:
Hullo! I made a PR. Please checkout out #552

johnryan:
@majapw Thanks for the PR! is there any recommended way to test this to make sure it fixes the issue? I tried installing with `"react-dates": "git@github.com:airbnb/react-dates.git#maja-add-null-check-for-calendarmonthgrid"` but it seems that the npm version is different than what's on github for some reason.

ljharb:
Anything that's got a build process won't really work when installed from github.

johnryan:
@ljharb Yea that's what I was afraid of, was just hoping to verify there wasn't some other similar null ref issue that will be revealed by this fix.

majapw:
Yeah unfortunately, we'll have to do a release first I think. I'll get on that ASAP.

majapw:
Check out v12.1.2! This is hopefully addressed there.

[...]

PR #552, opened 2017-06-08
Adds null check for this.calendarMonthGrid ref

majapw:
Fix for https://github.com/airbnb/react-dates/issues/529

This should *probably* address #529. Having the ref be set in a constructor-bound method instead of in an inline arrow function should prevent the ref from being set to null *unless* it is unmounted. The extra null check should, in edge cases we are not thinking of, prevent the error itself from being thrown.

to: @ljharb @jherencia @johnryan

### DIFF
--- a/src/components/DayPicker.jsx
+++ b/src/components/DayPicker.jsx
@@ -209,6 +209,7 @@ export default class DayPicker extends React.Component {
     this.onKeyDown = this.onKeyDown.bind(this);
     this.onPrevMonthClick = this.onPrevMonthClick.bind(this);
     this.onNextMonthClick = this.onNextMonthClick.bind(this);
+    this.setCalendarMonthGridRef = this.setCalendarMonthGridRef.bind(this);
     this.multiplyScrollableMonths = this.multiplyScrollableMonths.bind(this);
     this.updateStateAfterMonthTransition = this.updateStateAfterMonthTransition.bind(this);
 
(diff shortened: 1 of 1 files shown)


src/components/DayPicker.jsx
```diff
@@ -498,12 +503,15 @@ export default class DayPicker extends React.Component {
...
   initializeDayPickerWidth() {
-    this.dayPickerWidth = calculateDimension(
-      // eslint-disable-next-line react/no-find-dom-node
-      ReactDOM.findDOMNode(this.calendarMonthGrid).querySelector('.CalendarMonth'),
-      'width',
-      true,
-    );
+    // eslint-disable-next-line react/no-find-dom-node
+    const calendarMonthGridDOMNode = ReactDOM.findDOMNode(this.calendarMonthGrid);
+    if (calendarMonthGridDOMNode) {
```
ljharb: i'd actually wrap the findDOMNode in a truthy check for the ref; if it's nullary, there's no point in doing the findDOMNode

src/components/DayPicker.jsx
```diff
@@ -538,12 +546,15 @@ export default class DayPicker extends React.Component {
...
 
-    // clear the previous transforms
-    applyTransformStyles(
-      // eslint-disable-next-line react/no-find-dom-node
-      ReactDOM.findDOMNode(this.calendarMonthGrid).querySelector('.CalendarMonth'),
-      'none',
-    );
+    // eslint-disable-next-line react/no-find-dom-node
+    const calendarMonthGridDOMNode = ReactDOM.findDOMNode(this.calendarMonthGrid);
+    if (calendarMonthGridDOMNode) {
```

Issue #525, opened 2017-05-19
Clear Dates Button doesn't clear partially entered or erroneous date


AntiFish03:
If you type in the same date or a partial date the clear dates button will not clear the data from the field. 

<img width="650" alt="screen shot 2017-05-19 at 10 50 22 am" src="https://cloud.githubusercontent.com/assets/381196/26256021/2e3bc4bc-3c81-11e7-9d78-fb4dd590ff9a.png">
<img width="651" alt="screen shot 2017-05-19 at 10 50 32 am" src="https://cloud.githubusercontent.com/assets/381196/26256030/35b0c454-3c81-11e7-81f6-bf0fdcdfdce5.png">

Also it doesn't matter if it is start date or end date.
<img width="359" alt="screen shot 2017-05-19 at 10 53 13 am" src="https://cloud.githubusercontent.com/assets/381196/26256078/61e3eb32-3c81-11e7-8d10-88e6d0770d27.png">

timhwang21:
Seems like the solution would be to lift `this.state.dateString` out of `<DateInput/>` and into `<DateRangePickerInputController/>` and `<SingleDatePicker/>` and instead pass it down as a prop, and clear out that value when `clearDates()` or `clearDate()` is called. This issue is that there's no way for `<DateInput/>` to know that a "clear values" action just occurred. @majapw WDYT?

tienlx93:
Any update on this issue?

PR #554, opened 2017-06-10
Allow clear dates to clear partially entered dates

timhwang21:
Addresses #525.

Move user input state out of `<DateInput/>` and into either `<DateRangePickerInputController/>` or `<SingleDatePicker/>`, passing down the value and a setter as props.


timhwang21:
Agreed, I couldn't think of anything better that differentiated from the existing variable inputValue.

timhwang21:
@majapw rebased this, #555, and #580, please let me know if there's anything else I can do for these PRs!

jrparish:
We are running into #525 and would love to have this change merged. Is there any reason we're holding on this PR?

PR #555, opened 2017-06-12
Allow same day input when `minimumNights={0}`

### LINKED ISSUE #353: DateRangePicker manual input breaks for same day
I'm using the DateRangePicker, and if I only interact with my mouse, I can choose the same date assuming I set the prop `minimumNights={ 0 }` (per the docs). However, if I try to manually type in the same date, the picker won't accept the same date for my start and end dates.

timhwang21:
Fixes #353.

### DIFF
--- a/src/components/DateRangePickerInputController.jsx
+++ b/src/components/DateRangePickerInputController.jsx
@@ -16,6 +16,7 @@ import toISODateString from '../utils/toISODateString';
 
 import isInclusivelyAfterDay from '../utils/isInclusivelyAfterDay';
 import isInclusivelyBeforeDay from '../utils/isInclusivelyBeforeDay';
+import isSameDay from '../utils/isSameDay';
 
 import { START_DATE, END_DATE } from '../../constants';
 
@@ -41,6 +42,7 @@ const propTypes = forbidExtraProps({
   keepOpenOnDateSelect: PropTypes.bool,
   reopenPickerOnClearDates: PropTypes.bool,
   withFullScreenPortal: PropTypes.bool,
+  minimumNights: PropTypes.number,
   isOutsideRange: PropTypes.func,
   displayFormat: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
 
@@ -85,6 +87,7 @@ const defaultProps = {
   keepOpenOnDateSelect: false,
   reopenPickerOnClearDates: false,
   withFullScreenPortal: false,
+  minimumNights: 1,
   isOutsideRange: day => !isInclusivelyAfterDay(day, moment()),
   displayFormat: () => moment.localeData().longDateFormat('L'),
 
@@ -130,6 +133,7 @@ export default class DateRangePickerInputController extends React.Component {
     const {
       startDate,
       isOutsideRange,
+      minimumNights,
       keepOpenOnDateSelect,
       onDatesChange,
     } = this.props;
@@ -137,7 +141,8 @@ export default class DateRangePickerInputController extends React.Component {
     const endDate = toMomentObject(endDateString, this.getDisplayFormat());
 
     const isEndDateValid = endDate && !isOutsideRange(endDate) &&
-      !isInclusivelyBeforeDay(endDate, startDate);
+      (!isInclusivelyBeforeDay(endDate, startDate) ||
+      (minimumNights === 0 && isSameDay(endDate, startDate)));
     if (isEndDateValid) {
       onDatesChange({ startDate, endDate });
       if (!keepOpenOnDateSelect) this.onClearFocus();
@@ -166,10 +171,11 @@ export default class DateRangePickerInputController extends React.Component {
     const startDate = toMomentObject(startDateString, this.getDisplayFormat());
 
     let { endDate } = this.props;
-    const { isOutsideRange, onDatesChange, onFocusChange } = this.props;
+    const { isOutsideRange, minimumNights, onDatesChange, onFocusChange } = this.props;
     const isStartDateValid = startDate && !isOutsideRange(startDate);
     if (isStartDateValid) {
-      if (isInclusivelyBeforeDay(endDate, startDate)) {
+      if (isInclusivelyBeforeDay(endDate, startDate) &&
+        !(minimumNights === 0 && isSameDay(endDate, startDate))) {
         endDate = null;
       }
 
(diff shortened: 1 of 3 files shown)


src/components/DateRangePickerInputController.jsx
```diff
@@ -41,6 +42,7 @@ const propTypes = forbidExtraProps({
   keepOpenOnDateSelect: PropTypes.bool,
   reopenPickerOnClearDates: PropTypes.bool,
   withFullScreenPortal: PropTypes.bool,
+  minimumNights: PropTypes.number,
```
ljharb: this should probably be a non-negative integer (`nonNegativeInteger` from `airbnb-prop-types`), not just a number

test/components/DateRangePickerInputController_spec.jsx
```diff
@@ -113,33 +113,168 @@ describe('DateRangePickerInputController', () => {
...
-        const onDatesChangeStub = sinon.stub();
-        const wrapper =
-          shallow(<DateRangePickerInputController onDatesChange={onDatesChangeStub} />);
-        wrapper.instance().onEndDateChange(validFutureDateString);
-        expect(onDatesChangeStub.callCount).to.equal(1);
+      describe('when props.startDate is not provided', () => {
+        it('calls props.onDatesChange with provided end date', () => {
+          const onDatesChangeStub = sinon.stub();
+          const wrapper =
+            shallow(<DateRangePickerInputController onDatesChange={onDatesChangeStub} />);
```
ljharb: please don't add line breaks after assignments (same throughout, even if that was already in the file); if this is too long, let's indent like so:
```jsx
const wrapper = shallow((
  <DateRangePickerInputController onDatesChange={onDatesChangeStub} />
));
```

test/components/DateRangePickerInputController_spec.jsx
```diff
@@ -113,33 +113,168 @@ describe('DateRangePickerInputController', () => {
...
+          wrapper.instance().onEndDateChange(validFutureDateString);
+          expect(onDatesChangeStub.callCount).to.equal(1);
+
+          const onDatesChangeArgs = onDatesChangeStub.getCall(0).args[0];
```
ljharb: `const [{ startDate, endDate }] = onDatesChangeStub.getCall(0).args;`

test/components/DateRangePickerInputController_spec.jsx
```diff
@@ -401,6 +536,74 @@ describe('DateRangePickerInputController', () => {
...
+
+        describe('props.minimumNights is greater than 0', () => {
+          it('calls props.onDatesChange with provided start date and null end date', () => {
+            const onDatesChangeStub = sinon.stub();
+            const wrapper = shallow(
+              <DateRangePickerInputController
+                onDatesChange={onDatesChangeStub}
+                endDate={endDate}
+                minimumNights={1}
+              />);
```
ljharb: similarly, since the opening `(` ends a line, the closing `)` should start a line.

test/components/DateRangePickerInputController_spec.jsx
timhwang21: +1, on a related note, wondering if there are plans to add anything like Prettier to the code base?

test/components/DateRangePickerInputController_spec.jsx
timhwang21: +1, on a related note, wondering if there are plans to add anything like Prettier to the code base?

ljharb: Definitely not; if there's anything eslint doesn't do, let's make eslint better. I'm not yet aware of anything prettier is theoretically capable of that eslint theoretically isn't :-)

test/components/DateRangePickerInputController_spec.jsx
```diff
@@ -58,8 +60,9 @@ describe('DateRangePickerInputController', () => {
       describe('props.onFocusChange', () => {
         it('is not called', () => {
           const onFocusChangeStub = sinon.stub();
-          const wrapper =
-            shallow(<DateRangePickerInputController onFocusChange={onFocusChangeStub} />);
+          const wrapper = shallow(
+            <DateRangePickerInputController onFocusChange={onFocusChangeStub} />,
```
timhwang21: @ljharb putting a trailing comma seems like the more consistent pattern in the codebase vs. double parens, does this look right to you? Thanks!

ljharb: yes, either one's totally fine :-)

unfortunately comma-dangle for functions doesn't have an exception for "1-arg jsx", which is the one case it looks weird.

src/components/DateRangePickerInputController.jsx
```diff
@@ -130,14 +133,16 @@ export default class DateRangePickerInputController extends React.Component {
...
       keepOpenOnDateSelect,
       onDatesChange,
     } = this.props;
 
     const endDate = toMomentObject(endDateString, this.getDisplayFormat());
 
     const isEndDateValid = endDate && !isOutsideRange(endDate) &&
-      !isInclusivelyBeforeDay(endDate, startDate);
+      (!isInclusivelyBeforeDay(endDate, startDate) ||
+      (minimumNights === 0 && isSameDay(endDate, startDate)));
```
majapw: This is a bit of a nit, but could we change this to:
```
(
  !isInclusivelyBeforeDay(endDate, startDate) ||
  (minimumNights === 0 && isSameDay(endDate, startDate))
);
```

Also, I feel like it's a bit weird that we don't check for minimum nights requirements in other scenarios. Maybe this should just be:
```
 isInclusivelyAfterDay(endDate, startDate + minimumNights)
```
Would that work?

timhwang21: +1 to the first part, and ah yeah, I see how the `Inclusively` functions work now, should have taken a closer look -- `!isInclusivelyBeforeDay()` is equivalent to `isAfterDay()` so yeah, makes sense to me. Though we'd have to do `startDate.clone().add(minimumNights, 'days')`. Do you think it's worth the clone?

Thanks!

timhwang21: Hey! Wondering if there's anything else I can do WRT this PR? Thanks!

majapw: I think this is worth the clone in this scenario. Let's change it to be `isInclusivelyAfterDay(endDate, startDate + minimumNights)`.

timhwang21:
Sounds good, made the change (though I used `isBeforeDay` to make the boolean logic less ugly)

Issue #25, opened 2016-08-19
Allow Year/month navigation


rodryquintero:
This allows for a quick way to jump to another year or month without having to use the nav keys  and go one month at a time.

Here is an example. Clicking on the year or month displays another dialog showing the years/months.

[React DatePicker](http://zippyui.com/docs/react-date-picker/examples)

ljharb:
While I love the idea of direct yearly navigation, having one modally thing pop up over another seems horrifying to me.

rodryquintero:
How about a dropdown.

supnate:
+1 for year navigation. Birthday input is a case.

themre:
year/month navigation seems to me very important. I wouldn't have additional popup, I would replace the existing popup content with years/months.

madisvain:
I think that this is a very important feature also but the UX of this should not be a new pop up but rather open inside the same widget.

Something like this:
http://eonasdan.github.io/bootstrap-datetimepicker/

agupta93:
I had one doubt if we go by @madisvain suggestion which seems doable, we would need to create another component `monthPicker` which would render the month and year view right? @ljharb

rodryquintero:
Seems like it. Btw, sorry for not working on my PR. Have not found the time
to do it.

agupta93:
@rodryquintero but we still need to finalize what kind of behaviour is desired right?

rodryquintero:
I vote for the menu option. Easier to implement and more practical.

agupta93:
@rodryquintero the example suggested by @madisvain seems doable and practical to me lets see what @ljharb and @majapw have to say about this

infacq:
hi... is there already in implementation?

englishtom:
+1 here. As @supnate says- it's a blocking issue for DoB use case.

watsonink:
+1 here also. We would like to navigate through the the months and years.

lokat:
+1

alsma:
+1

ljharb:
Please do not post "+1" comments - everywhere on Github, these add noise and do not add value. Please indicate your support with emojis on the original post.

I will delete all such comments now, and any further in the future. Thanks!

infacq:
I think sometimes it's better to stick with html5 date component... just for this purpose

ljharb:
Sadly the HTML5 date component does not style consistently across browsers and the experience is subpar.

khmelevskii:
+1

[...]

PR #558, opened 2017-06-12
[Issue 25] Month and Year Dropdowns

jvanderz22:
This adds the feature requested in https://github.com/airbnb/react-dates/issues/25 with simple `<select>` boxes for month and year, which can be turned on with the `enableDropdowns` options. When enabled, this looks like:

screen shot 2017-06-12 at 5 59 40 pm

This is my first PR to this package and the first thing I've ever really built in React, so I'm open to and appreciate all feedback

### DIFF
--- a/src/components/DayPicker.jsx
+++ b/src/components/DayPicker.jsx
@@ -38,10 +38,13 @@ const MONTH_PADDING = 23;
 const DAY_PICKER_PADDING = 9;
 const PREV_TRANSITION = 'prev';
 const NEXT_TRANSITION = 'next';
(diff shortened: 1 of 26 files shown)


src/components/DayPicker.jsx
```diff
@@ -409,6 +419,30 @@ export default class DayPicker extends React.Component {
     });
   }
 
+  onMonthChange(currentMonth) {
+    // Translation value is a hack to force an invisible transition that
```
jvanderz22: I couldn't figure out a better way to do this, but that might have been due to my inexperience with React more than anything

PR #565, opened 2017-06-14
remove shallow compare in favor of PureComponent

ingoclaro:
React 15.3 introduces PureComponent making shallowCompare a legacy add-on if favor of React.PureComponent.
https://facebook.github.io/react/docs/shallow-compare.html

This removes that dependency and uses PureComponent where shallow compare was used.

### DIFF
--- a/package.json
+++ b/package.json
@@ -88,7 +88,6 @@
     "nyc": "^10.3.2",
     "raw-loader": "^0.5.1",
     "react": "^15.5.4",
-    "react-addons-shallow-compare": "^15.5.2",
     "react-addons-test-utils": "^15.5.1",
     "react-dom": "^15.5.4",
     "react-svg-loader": "^1.1.1",
@@ -114,9 +113,8 @@
   },
   "peerDependencies": {
     "moment": "^2.18.1",
-    "react": ">=0.14",
-    "react-dom": ">=0.14",
-    "react-addons-shallow-compare": ">=0.14"
+    "react": ">=15.3.0",
+    "react-dom": ">=15.3.0"
   },
   "greenkeeper": {
     "ignore": [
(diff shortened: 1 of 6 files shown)


package.json
```diff
@@ -114,9 +113,8 @@
   },
   "peerDependencies": {
     "moment": "^2.18.1",
-    "react": ">=0.14",
-    "react-dom": ">=0.14",
-    "react-addons-shallow-compare": ">=0.14"
+    "react": ">=15.3.0",
+    "react-dom": ">=15.3.0"
```
majapw: Hey @ljharb, I feel like it's a reasonable thing for us to expect `15.3.0` or higher at this point in time, but I'm unsure... do you think this is fine?

ljharb: I'm not personally comfortable with dropping react 14 support yet (that would be a semver major).

 Unless we want to implement a detection mechanism, such that we only inherit from PureComponent in 15.3+, I don't think this is a change we should make.

majapw: What do you think is the estimated time to live on that? React 15 was currently released over a year ago (April 2016).

ingoclaro: Something to consider is that doing the upgrade facilitates the use of preact instead of react :)

ljharb: @ingoclaro if preact doesn't support shouldComponentUpdate, that sounds like a bug in preact :-)

@majapw enzyme supports 0.13 still. I think that a) a general guideline should be "latest, + 2 versions back" (we never supported 0.13 here, ofc, but we probably should have) and b) we shouldn't drop support for something without a good reason.

PureComponent in React iirc is just a shouldComponentUpdate that's written for you - imo that's not a good enough reason by itself.

detj:
React 16 release is nearby. Any plans to drop `react-addons-shallow-compare` in favor of `PureComponent` now ?

ljharb:
Once v16 is released, then certainly that makes 0.13 droppable; but to support PureComponent we'd need to drop React <= 15.2, which can't happen until v18 comes out (when all of v15 could be dropped).

Then separately, there'd have to be a better reason to drop a React version than "it implements shouldComponentUpdate for us".

detj:
Sure. Makes sense.

PR #575, opened 2017-06-16
Fix issue where range doesn't clear on invisible months

majapw:
Fix for https://github.com/airbnb/react-dates/issues/567

I think this is a more bullet-proof solution than https://github.com/airbnb/react-dates/pull/574

It took me quite a while to figure out how to get this working for the `enableOutsideDays` prop, but I think this solution is holistic and does not have outside side effects.

PTAL @airbnb/webinfra @acr13


src/components/DayPickerRangeController.jsx
```diff
@@ -600,34 +600,42 @@ export default class DayPickerRangeController extends React.Component {
     if (!day || !isDayVisible(day, currentMonth, numberOfMonths, enableOutsideDays)) {
       return updatedDays;
     }
-    let monthIso = toISOMonthString(day);
-    let month = updatedDays[monthIso] || visibleDays[monthIso];
     const iso = toISODateString(day);
 
+    let updatedDaysAfterAddition = { ...updatedDays };
     if (enableOutsideDays) {
```
majapw: @ljharb This first change is in the `addModifier` method. The basic premise is that if outside days are enabled that means that days at the beginning and end of the month will exist as a part of two separate `CalendarMonth` objects. As a result if you click on a day, you may want to add modifiers to more than one month instance in the state. This is what this branch of the conditional accomplishes.

majapw: @ljharb This first change is in the `addModifier` method. The basic premise is that if outside days are enabled that means that days at the beginning and end of the month will exist as a part of two separate `CalendarMonth` objects. As a result if you click on a day, you may want to add modifiers to more than one month instance in the state. This is what this branch of the conditional accomplishes.

src/components/DayPickerRangeController.jsx
```diff
@@ -643,38 +651,55 @@ export default class DayPickerRangeController extends React.Component {
...
     if (!day || !isDayVisible(day, currentMonth, numberOfMonths, enableOutsideDays)) {
       return updatedDays;
     }
 
-    let monthIso = toISOMonthString(day);
-    let month = updatedDays[monthIso] || visibleDays[monthIso];
     const iso = toISODateString(day);
+
+    let updatedDaysAfterDeletion = { ...updatedDays };
     if (enableOutsideDays) {
```
majapw: This addition to the `deleteModifier` method is very similar to what happens in `addModifier`. Namely, for outside days, sometimes a selected day might be in two `CalendarMonth`s and we need to remove the modifiers in both cases.

src/components/DayPickerRangeController.jsx
```diff
@@ -643,38 +651,55 @@ export default class DayPickerRangeController extends React.Component {
...
 
   deleteModifier(updatedDays, day, modifier) {
-    const { numberOfMonths, enableOutsideDays } = this.props;
-    const { currentMonth, visibleDays } = this.state;
+    const { numberOfMonths: numberOfVisibleMonths, enableOutsideDays, orientation } = this.props;
+    const { currentMonth: firstVisibleMonth, visibleDays } = this.state;
+
+    let currentMonth = firstVisibleMonth;
+    let numberOfMonths = numberOfVisibleMonths;
+    if (orientation !== VERTICAL_SCROLLABLE) {
```
majapw: `currentMonth` and `numberOfMonths` only apply to the visible months and not to those hidden for animation. However, we want to make sure modifiers are updates even in the invisible months so we expand the range here. ... I didn't think we needed this when adding modifiers but on second thought I think we do (there's like this edge case where you select a start date, scroll it off screen, select and end date and scroll back).

majapw:
@ljharb I added comments here as well to help guide the review!

src/components/DayPickerRangeController.jsx
ljharb: This `if` logic (ie, getting `currentMonth` and `numberOfMonths`) is repeated in DayPickerRangeController, and in SingleDatePicker. Could we move it to a shared function instead?

src/components/DayPickerRangeController.jsx
```diff
@@ -595,39 +595,55 @@ export default class DayPickerRangeController extends React.Component {
...
-        monthIso = Object.keys(visibleDays).filter(monthKey => (
-          monthKey !== monthIso && Object.keys(visibleDays[monthKey]).indexOf(iso) > -1
-        ))[0];
-        month = updatedDays[monthIso] || visibleDays[monthIso];
-      }
+      const monthsToUpdate = Object.keys(visibleDays).filter(monthKey => (
+        Object.keys(visibleDays[monthKey]).indexOf(iso) > -1
+      ));
+
+      updatedDaysAfterAddition = monthsToUpdate.reduce((days, monthIso) => {
```
ljharb: also one or both branches of this if/else seem repeated in both components; could they also be shared?

majapw: I think it is complicated to share these methods because a lot of these variables depend on state and/or props. I think the dream scenario is eventually to have a `withModifiers` HOC of some sort that could do a lot of this stuff on its own.

PR #576, opened 2017-06-18
[minor] deprecate `utils/isTouchDevice`; use `is-touch-device` instead

ljharb:
The only part of this that's semver-minor is the deprecation warning; otherwise it's a patch.


LillaEli:
Hi,
I think you've missed one usage of the deprecated util, i.e. line 11 in DayPickerSingleDateController.jsx
Could you add a patch for this, for some reason the warning is messing with my Karma tests :P

ljharb:
@LillaEli thanks, fixed in 77e2135d2009994fbf2c62e3ff68ce82e5786194

Issue #570, opened 2017-06-16
Modifiers not recalculated when focusedInput NOT changed, but dates changed


timhwang21:
Kind of a tricky case that's a followup of #517.

In #517, the issue was that the act of switching from `startDate` being focused to `endDate` being focused didn't lead to recomputation. That was fixed.

However, I've run into an interesting situation that seems related. Basically, the default of the behavior, when a start date is selected, is to switch focus over to `endDate`. However, if a date PRECEDING `startDate` is selected while `endDate` is focused, `startDate` value is reset, but modifiers are not recalculated.

This affects cases where valid values for end date depend on the currently selected start date (e.g. must be the same day of week as start date).

I can look into this further, and I'll upload a GIF showing this interaction.

timhwang21:
![jun-15-2017 20-10-24](https://user-images.githubusercontent.com/5831434/27206661-c3eb349e-5206-11e7-8bab-c69aadde77c0.gif)

Gif 1: Showcases expected behavior. I need to prevent users from selecting ranges that include blocked dates (issue #23), for now I'm achieving that with validation functions. When a date before any blocked date is selected, all dates after the first blocked date are blocked.

![jun-15-2017 20-10-40](https://user-images.githubusercontent.com/5831434/27206662-c3ec0e32-5206-11e7-9b4a-4154a51d9fff.gif)

Gif 2: Showcases bugged case. When selecting a start date, and then selecting a previous start date while `endDate` is focused, modifiers are not recomputed. However, as seen from the gif, dates seem to still be properly blocked, they just lack hover effects.

PR #580, opened 2017-06-18
[WIP] Recompute modifiers on date change, not just focus change

timhwang21:
Addresses #570.

Currently, block / range related modifiers are only recomputed when the currently focused input changes. However, this causes some issues when more complex block / range functions are used, e.g. those that make selectable dates conditional on current state.

This commit gives one potential fix, which is to recompute modifiers not just on focus change, but also on start / end date change. This DOES result in a slight perf cost because modifiers are computed more frequently; however from my experience perf bottlenecks have mainly arisen on month changes (where 30 new days are added and recomputed) and on hover changes (fixed in v11.x), not on selection change.

The added story gives an example of a case where range modifier computation does not work as expected. The added wrapper component has the following validation: start dates must be no earlier than 1 week after selected end date, and end dates must be no later than 1 week after selected start date. Without the change in `DayPickerRangeController`, the following actions lead to a bug:

1. Pick a start date. Note the modifiers correctly recompute after the end date is focused.
2. Click on a date BEFORE the start date. `react-dates` behavior is to change the currently selected start date. However, note that the modifiers DO NOT recompute.

Additionally, there is a similar bug when `keepOpenOnDateSelect` is enabled, which is more noticeable with the `SingleDatePicker`. Because modifiers recompute when the focused input toggles between `null` and the input, modifiers are never recomputed until the calendar is closed and reopened. The story in `SingleDatePicker` showcases this -- when the change in `SingleDatePicker.jsx` is undone, the modifier never recomputes.

If this change seems ok I'll delete the changes to unrelated files, and add test cases. Thanks!

### DIFF
--- a/src/components/DayPickerSingleDateController.jsx
+++ b/src/components/DayPickerSingleDateController.jsx
@@ -547,9 +547,10 @@ export default class DayPickerSingleDateController extends React.Component {
       renderCalendarInfo,
       isFocused,
       isRTL,
+      phrases,
     } = this.props;
 
-    const { phrases, currentMonth, visibleDays } = this.state;
+    const { currentMonth, visibleDays } = this.state;
 
     return (
       <DayPicker


examples/TimTest.jsx
```diff
@@ -0,0 +1,160 @@
...
+  onDatesChange({ startDate, endDate }) {
+    this.setState({ startDate, endDate });
+  }
+
+  onFocusChange(focusedInput) {
+    this.setState({ focusedInput });
+  }
+
+  isOutsideRange(day) {
+    const { focusedInput, startDate, endDate } = this.state;
```
timhwang21: Example of a problematic range function. In my use cases, validations like this are quite frequently used.

mikkelwf:
Is there any update on this..?
I'm trying to implement react-dates in my app right now, and i'm struggling with this particular problem.
Would really appreciate a fix/merge soon

timhwang21:
Hey @majapw ! anything I can do to move this along? Thanks!

timhwang21:
I screwed something up when manually rebasing on the CLI, re-filed at #652 😓

ljharb:
@timhwang21 for future reference, as long as you haven't deleted the remote branch, it can always be repaired, and that's highly preferable to opening up a duplicate PR. Happy to help in the future if you have git issues.

timhwang21:
Thanks, and sorry! I ended up doing an interactive rebase and a forced push to the existing remote branch to get the commits in the right order, which ended up closing the existing PR showing it with 0 files changed, and giving a prompt to file a PR on my fork.

ljharb:
ah, gotcha :-/

PR #587, opened 2017-06-20
OZ-327: es6-set polyfill added to run tests with phantom-js

pyotruk:
ES6 Set polyfill added to be able to run tests with PhantomJS.


majapw:
@ljharb is this our preferred Set polyfill? Do we want the polyfill in react-dates directly?

Also this change appears to introduce a peer dep issue...

ljharb:
No, we should be running airbnb-browser-shims, which includes es6-shim.

PR #601, opened 2017-06-27
Bump storybook version to v3 and webpack to v2

majapw:
Hooray!

There is one small issue with @storybook/info-addon where we get a proptype warning in the console (it is on the way to being fixed - https://github.com/storybooks/storybook/pull/1374). Other than that, the storybook, build script, and tests all seem to work as intended.

@ljharb can you take another look?


.storybook/.eslintrc
```diff
@@ -0,0 +1,10 @@
+{
+  "rules": {
+    "import/no-extraneous-dependencies": [2, {
+      "devDependencies": true
+    }],
+    "react/jsx-filename-extension": 2,
```
ljharb: why can't our jsx files be .jsx for storybook?

.storybook/.eslintrc
```diff
@@ -0,0 +1,10 @@
...
+    "global-require": 2,
+    "comma-dangle": 2
```
ljharb: also, por que?

.storybook/addons.js
```diff
@@ -1,2 +1,4 @@
-import '@kadira/storybook/addons';
-import '@kadira/storybook-addon-options/register';
+/* eslint-disable import/no-extraneous-dependencies, import/no-unresolved, import/extensions */
```
ljharb: why is this override needed?

.storybook/.eslintrc
majapw: i think they might have to be js due to storybook's internal config?

.storybook/.eslintrc
majapw: mostly to get my webpack config to stop yelling at me in sublime but also i guess maybe not necessary.

also not working. :/

.storybook/addons.js
majapw: This is the default storybook config as per the slow start guide. The reason is that these modules are included as part of the `@storybook/react` package and not explicitly defined in our package.json

stories/DateRangePicker_calendar.js
```diff
@@ -136,12 +136,6 @@ storiesOf('DRP - Calendar Props', module)
       autoFocus
     />
   ))
-  .addWithInfo('with keyboard shorcuts panel hidden', () => (
```
majapw: This was a duplicate story which used to just be ignored but now errors out.

.storybook/.eslintrc
ljharb: k

.storybook/addons.js
ljharb: sure, but `no-unresolved` still should be validating that it's importable, no?

.storybook/webpack.config.js
```diff
@@ -2,19 +2,35 @@ const path = require('path');
...
-    loaders: [
+    rules: [
       {
-        test:   /\.scss$/,
-        loaders: ['style', 'raw', 'sass'],
-        include: path.resolve(__dirname, '../css/')
+        test: /\.scss$/,
+        use: ['style-loader', 'raw-loader', 'sass-loader'],
+        include: [
+          path.resolve(__dirname, './'),
```
lencioni: Is this intentional? If you have this line, do you need the next line?

majapw: Yes, so this is looking in the `/.storybook` config folder, which contains the scss for the storybook whereas the next line looks for the `react-dates` scss files, which are located in the `/css` folder

Issue #560, opened 2017-06-14
Respecting readOnly prop


aschiopu:
Howdy,

[**SingleDatePicker.jsx**](https://github.com/airbnb/react-dates/blob/master/src/components/SingleDatePicker.jsx#L782)
I am able to pass in the `readOnly` prop but it only works if the prop itself is true. If I specify **false**, it's not respected on touch devices because of the isTouch logic [here](https://github.com/airbnb/react-dates/blob/master/src/components/DateInput.jsx#L176).

There are two ways of solving this:

- react-dates stops enforcing behavior on touch devices (ripping out the `isTouch` logic in the `DateInput` component)
- we remove the default value of `isTouch` to be false in the `DateInput` component. 

Let me know which you prefer and I'll make a pr. Thank you!

ljharb:
The default value of `isTouch` is already false - it's only `true` when ran on a touch device.

To handle this, we'd have to remove the defaultProp of `readOnly`, and make it be three-state - true, false, or "not provided" - and then in the `true` or the "not provided" case, set `readOnly` to true.

In other words, I don't think either of your suggestions is a good way to solve this.

majapw:
Hi @aschiopu, what is the effect you are going for? Do you want a typeable input on touch devices? The problem with this approach is that it generally pops up the keyboard which then interferes with the datepicker itself.

aschiopu:
> In other words, I don't think either of your suggestions is a good way to solve this.
@ljharb You haven't mentioned why `DateInput`dictates the behavior of the component on different devices? It's something that the consuming parent component can dictate easily making this component more flexible and re-usable. 

@majapw I understand the reasoning behind having it read only for most touch devices, except the Surface and most windows laptops. More and more windows laptops come out with touch screens, yet they still have a keyboard and mouse.

Overall I think think this behavior/logic should probably be dictated by the parent consuming component as it relies on the devices that use of this component. The windows machines might not be the case at AirBnB, but it could be true at other companies.

ljharb:
It's a really fair point that `DateInput` should probably be dumb, and the wrappers should be determining touch support and `readOnly` - unless we can detect the problematic touch cases inside DateInput somehow.

majapw:
Yeah, we probably want to actually make the readOnly prop totally configurable and then push up that decision into our implementation... what do you think @ljharb?

ljharb:
That sounds like the right decision.

majapw:
Will be in the next release! ✌️

PR #602, opened 2017-06-27
issue 560: updates DateInput component's readOnly prop to be configurable by consumer implementation

ViktoriyaSavkina:
This PR solves issue 560 as per discussion there
https://github.com/airbnb/react-dates/issues/560

### DIFF
--- a/src/components/DateInput.jsx
+++ b/src/components/DateInput.jsx
@@ -37,7 +37,7 @@ const defaultProps = {
   focused: false,
   disabled: false,
   required: false,
-  readOnly: false,
+  readOnly: null,
   showCaret: false,
 
   onChange() {},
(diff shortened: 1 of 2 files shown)


majapw:
Please rebase and fix the corresponding test. O/w looks good to me! :D

ViktoriyaSavkina:
@majapw done :)

src/components/DateInput.jsx
```diff
@@ -131,7 +131,6 @@ export default class DateInput extends React.Component {
   render() {
     const {
       dateString,
-      isTouchDevice: isTouch,
```
ljharb: do we still need the `isTouchDevice` state?

src/components/DateInput.jsx
```diff
@@ -172,7 +171,7 @@ export default class DateInput extends React.Component {
           placeholder={placeholder}
           autoComplete="off"
           disabled={disabled}
-          readOnly={readOnly || isTouch}
+          readOnly={readOnly}
```
ljharb: i wonder if we could make this semver-patch by changing this to `typeof readOnly === 'boolean' ? ‼readOnly : isTouch`, and making the `defaultProp` be `null`?

ViktoriyaSavkina:
@ljharb updated per your comments

ljharb:
we shouldn't fear breaking changes, but we should strive to avoid them :-)

as for coveralls, it might be having issues on all repos; i've seen coverage badges broken for awhile on a bunch of them. i'm not sure what the "unknown coverage" is for tho.

Issue #593, opened 2017-06-23
DateRangePicker disable endDate?


amitava82:
I could not find anyway to disable endDate. Is that possible? Need to disable for one way flight form.

timhwang21:
We handled this by using two SingleDatePickers, one disabled. This might be a fairly easy fix.

ljharb:
Having separate renders for two way and one way (i.e., range vs two singles) seems like the right way to handle this, as opposed to a range option to disable half of it.

timhwang21:
@ljharb could you clarify on that? At least in our case, we need to create start and end dates, with either the start or end date being conditionally fixed. The two SingleDatePicker approach works, but doesn't strike me as ideal (in part because the DateRangePicker is unmounting and two SingleDatePickers mounting when the condition to fix one of the dates is met). Thanks!

ljharb:
I'm envisioning the common UI on airline websites, where you pick "one way" or "round trip", and based on your selection, a different form/datepicking interface appears. Can you help me understand your use case, if it differs from this?

amitava82:
The way I see it, if user selects "one way" then endDate field simply disabled instead of rendering another set of SDP which isn't very clean IMHO. For now I'm rendering 2 sets of components as suggested.

timhwang21:
Sure -- basically, our business objects have a start date and an end date, and can be created or edited. In one case, when editing an item whose start date is in the past, the start date will be fixed, but the end date can be edited. We don't want the start date to be editable, but still displayed on the form as it as it's relevant information for the user. Currently what we do when a date is disabled is replace the range picker with a disabled input placed next to a single date picker, so the form field layout remains the same between conditions. (Our range input is also styled to look like two discrete inputs, vs. the default style of one input and different focus areas).

ljharb:
I guess it does seem reasonable to have a prop on `DateRangePicker` that allows disabling of either "start", "end", or "none" (default). Thanks for the explanation!

dart-wakar:
@ljharb The solution I have provided might not be sufficient.Please,have a look and suggest me if I need to make any changes

colemerrick:
[...]

PR #606, opened 2017-06-29
Fix for the issue of selectively disabling start or end date

dart-wakar:
#593 
The above mentioned issue is fixed by introducing a new prop into the DateRangePicker component: selectivelyDisable. selectivelyDisable is a string proptype,which is none by default,but if defined as 'startDate' in DateRangePickerWrapper it disables the start-date and so is the case with end-date.

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,7 +117,7 @@ The following is a list of other *OPTIONAL* props you may provide to the `DateRa
 // input related props
 startDatePlaceholderText: PropTypes.string,
 endDatePlaceholderText: PropTypes.string,
-disabled: PropTypes.bool,
+disabled: PropTypes.oneOfType([PropTypes.bool, PropTypes.oneOf([START_DATE, END_DATE])]),
 required: PropTypes.bool,
 readOnly: PropTypes.bool,
 screenReaderInputMessage: PropTypes.string,
--- a/src/components/DateRangePickerInput.jsx
+++ b/src/components/DateRangePickerInput.jsx
@@ -9,6 +9,7 @@ import openDirectionShape from '../shapes/OpenDirectionShape';
 
 import DateInput from './DateInput';
 import IconPositionShape from '../shapes/IconPositionShape';
+import DisabledShape from '../shapes/DisabledShape';
 
 import RightArrow from './RightArrow';
 import LeftArrow from './LeftArrow';
@@ -48,7 +49,7 @@ const propTypes = forbidExtraProps({
   isStartDateFocused: PropTypes.bool,
   isEndDateFocused: PropTypes.bool,
   showClearDates: PropTypes.bool,
-  disabled: PropTypes.oneOfType([PropTypes.bool, PropTypes.oneOf([START_DATE, END_DATE])]),
+  disabled: DisabledShape,
   required: PropTypes.bool,
   readOnly: PropTypes.bool,
   openDirection: openDirectionShape,
--- a/src/components/DateRangePickerInput.jsx
+++ b/src/components/DateRangePickerInput.jsx
@@ -3,8 +3,6 @@ import PropTypes from 'prop-types';
 import { forbidExtraProps, nonNegativeInteger } from 'airbnb-prop-types';
 import { css, withStyles, withStylesPropTypes } from 'react-with-styles';
 
-import selectivelyDisabled from '../shapes/DateRangePickerShape';
-
 import { DateRangePickerInputPhrases } from '../defaultPhrases';
 import getPhrasePropTypes from '../utils/getPhrasePropTypes';
 import openDirectionShape from '../shapes/OpenDirectionShape';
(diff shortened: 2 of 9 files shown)


src/components/DateRangePickerInput.jsx
```diff
@@ -41,6 +41,7 @@ const propTypes = forbidExtraProps({
   isEndDateFocused: PropTypes.bool,
   showClearDates: PropTypes.bool,
   disabled: PropTypes.bool,
+  selectivelyDisabled: PropTypes.string,
```
ljharb: instead of a string, let's use `PropTypes.oneOf(['none', 'startDate', 'endDate'])`, and let's have that be imported from the same shared module.

timhwang21: @ljharb what do you think about instead providing two boolean props, `disableStart` and `disableEnd`?

It might make the logic a little easier for users. Rather than
```
let selectivelyDisabled;

if (condition1) {
  selectivelyDisabled = 'startDate';
}
if (condition2) {
  selectivelyDisabled = 'endDate';
}
<DatePicker selectivelyDisabled={selectivelyDisabled}/>
```

it could be

```
<DatePicker disableStart={condition1} disableEnd={condition2}/>
```

Would also simplify the other code a bit as well.

I guess a potential gotcha would be that behavior of `<DatePicker disableStart disableEnd/>` would probably need to be identical to `<DatePicker disabled/>`.

ljharb: Generally we've been moving in the direction, for > 2-state props, of having a single enum prop instead of multiple boolean props. (With your suggestion, we'd want to specify a `mutuallyExclusiveTrueProp` propType to prevent both from being true anyways)

dart-wakar: I have made the changes as suggested.Have a look

ajwild:
Is there anything I can do to help get this merged? It would be very useful for a project I'm currently working on.

majapw:
If you'd like to take over this PR @dart-wakar, that would be great. 

I think that I'd personally prefer to roll this functionality up into the disabled prop and have it be able to be equal to `true`, `false`, `'startDate'` or `'endDate'`. I'd also like to see the prop type definition use the constants instead of hardcoding the strings.

what do you think @ljharb? i know you had some comments on the original PR.

ljharb:
I agree on constants; but I’m hesitant to roll up more meaning into an existing prop - but maybe it makes sense here for “disabled”.

If @ajwild or anyone else wants to update this PR, please comment here with a link to your branch and we can pull in the commits and rebase for you (rather than making a new PR).

ajwild:
@majapw @ljharb I've added a commit to the master branch of my fork: https://github.com/ajwild/react-dates

I had to pass through the disabled property from `DateRangePicker` to `DayPickerRangeController` to avoid updating or focusing on a disabled date when clicking on a calendar date.

It's also necessary to provide an `isOutsideRange` function to get the `blocked-out-of-range` CSS classes on the calendar, but the functionality still works without this. I think it would be messy trying to do this by updating the default `isOutsideRange` function.

Let me know if there's anything else required once you've had a chance to review/test it.

majapw:
Awesome @ajwild I'll pull it into this branch and take a look. Thank you for the work!

Can you clarify a bit about what you mean here? I'm not sure I understand.
> It's also necessary to provide an isOutsideRange function to get the blocked-out-of-range CSS classes on the calendar, but the functionality still works without this. I think it would be messy trying to do this by updating the default isOutsideRange function.

ajwild:
Sorry if I wasn't clear, I was referring to the "greyed-out" effect on the calendar dates.

Here's the code from the stories I added:
```
  .addWithInfo('disabled start date', () => (
    <DateRangePickerWrapper
      initialStartDate={moment().add(3, 'months')}
      initialEndDate={moment().add(3, 'months').add(10, 'days')}
      disabled="startDate"
      isOutsideRange={day => !isInclusivelyAfterDay(day, moment().add(3, 'months'))}
    />
  ))
  .addWithInfo('disabled end date', () => (
    <DateRangePickerWrapper
      initialStartDate={moment().add(3, 'months')}
      initialEndDate={moment().add(3, 'months').add(10, 'days')}
      disabled="endDate"
      isOutsideRange={day => !isInclusivelyAfterDay(day, moment()) ||
        !isInclusivelyBeforeDay(day, moment().add(3, 'months').add(10, 'days'))}
    />
  ))
```
A screenshot of the "disabled end date" example without `isOutsideRange`:
screenshot from 2018-03-26 13 55 53
A screenshot of the "disabled end date" example with `isOutsideRange`:
screenshot from 2018-03-26 13 56 22

README.md
```diff
@@ -117,7 +117,7 @@ The following is a list of other *OPTIONAL* props you may provide to the `DateRa
 // input related props
 startDatePlaceholderText: PropTypes.string,
 endDatePlaceholderText: PropTypes.string,
-disabled: PropTypes.bool,
+disabled: PropTypes.oneOfType([PropTypes.bool, PropTypes.oneOf([START_DATE, END_DATE])]),
```
majapw: We could probably pull this out into its own variable

majapw:
Ah @ajwild you are right. I think we want to leave it up to the developer to block out days when disabling if that's the UI they want. I don't think we should make any attempt to mess with it automagically.

README.md
ajwild: Should I reference it from "DateRangePickerShape.js" in the three other locations it is used, and leave it as it is (or simplify it) here in the readme?

majapw: I cleaned it up! :P But left it in the README for clarity.

Issue #613, opened 2017-07-03
is `initialVisibleMonth` now required for DayPickerRangeController?


terebentina:
I am calling `DayPickerRangeController` like this:
```
              <DayPickerRangeController
                startDate={this.state.customIntervalStart}
                endDate={this.state.customIntervalEnd}
                onDatesChange={this.onCustomDatesChange}
                focusedInput={this.state.focusedInput}
                onFocusChange={this.onDatePickerFocusChange}
                minimumNights={0}
                keepOpenOnDateSelect={false}
                renderCalendarInfo={() => (
                  <CalendarInfo onClick={this.onClearDatesClick} className={styles.calendarInfo} />
                )}
              />
```
It used to be fine until react-dates 12.2.0. With every 12.2 so far I am now getting
```
Uncaught TypeError: props.initialVisibleMonth is not a function
    at new DayPicker (DayPicker.js:555)
```

Am I supposed to specify `initialVisibleMonth` in `DayPickerRangeController` now? The docs still say it's optional.

Si1kIfY:
Same issue here but slightly different error message:

```
Uncaught (in promise) TypeError: props.initialVisibleMonth is not a function
    at new DayPicker (DayPicker.js:555)
    at ReactCompositeComponent.js:294
    at measureLifeCyclePerf (ReactCompositeComponent.js:75)
    at ReactCompositeComponentWrapper._constructComponentWithoutOwner (ReactCompositeComponent.js:293)
    at ReactCompositeComponentWrapper._constructComponent (ReactCompositeComponent.js:279)
    at ReactCompositeComponentWrapper.mountComponent (ReactCompositeComponent.js:187)
    at Object.mountComponent (ReactReconciler.js:45)
    at ReactCompositeComponentWrapper.performInitialMount (ReactCompositeComponent.js:370)
    at ReactCompositeComponentWrapper.mountComponent (ReactCompositeComponent.js:257)
    at Object.mountComponent (ReactReconciler.js:45)
```

aleib:
Same issue. As mentioned by @terebentina specifying the initialVisibleMonth removes the error.
So something like:
```initialVisibleMonth={() => moment().subtract(1, 'month')}```

majapw:
Thanks for the catch! I commented on the PR. :)

majapw:
Fixed in v12.2.4!

PR #617, opened 2017-07-05
Fix for initialVisibleMonth in DayPickerRangeController (#613)

aleib:
Original issue:
[is initialVisibleMonth now required for DayPickerRangeController](https://github.com/airbnb/react-dates/issues/613)

Looking at the commit logs I think this [commit](https://github.com/airbnb/react-dates/commit/48dfbd1b12bfd4a3658bf639630ea74ff073fe25#diff-a5cfcc83315d3bea09a9b06437d7ee80) broke the component when initialVisibleMonth is not set on DayPickerRangeController

By changing the default prop from:
```initialVisibleMonth: DayPickerDefaultProps.initialVisibleMonth```
to
```initialVisibleMonth: null,```
as it will pass ```null``` down to the DayPicker component on render

So made the prop a function as in [DateRangePicker](https://github.com/airbnb/react-dates/blob/master/src/components/DateRangePicker.jsx#L320)

### DIFF
--- a/src/components/DayPickerRangeController.jsx
+++ b/src/components/DayPickerRangeController.jsx
@@ -806,8 +806,6 @@ export default class DayPickerRangeController extends React.Component {
       onOutsideClick,
       withPortal,
       enableOutsideDays,
-      startDate,
-      initialVisibleMonth,
       hideKeyboardShortcutsPanel,
       daySize,
       focusedInput,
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/components/DayPickerRangeController.jsx [hunk 1]
```diff
@@ -819,6 +820,8 @@ export default class DayPickerRangeController extends React.Component {
     } = this.props;
 
     const { phrases, visibleDays } = this.state;
+    const initialVisibleMonthThunk =
+      initialVisibleMonth || (startDate ? () => startDate : () => moment());
```
majapw: This looks great! This is definitely a bug (I originally didn't catch is because the wrapper I'm using has an `initialVisibleMonth` default prop).

PR #620, opened 2017-07-06
Added missing keepOpenOnDateSelect prop

erin-doyle:
The DayPickerSingleDateController is using keepOpenOnDateSelect but that's not currently being passed to it from SingleDatePicker so it is instead always the default value of false.


erin-doyle:
Oh wow, not sure how I missed all of those! 😳  Fixed!

PR #630, opened 2017-07-19
Set tab index as -1 for default and custom date icons in SingleDatePicker

SainathB:
Currently date icons are coming in the tab navigation, which is not of any use. So, setting its tab index to -1.


sathvikc:
@majapw Have you added tab index to the button?

ljharb (approved):

majapw (approved):

Issue #132, opened 2016-10-18
[feature request] Prop for day of week format


daltones:
Currently month headers have the format `dd` hardcoded for week days (that resolves to “Su Mo Tu We Th Fr Sa” in English).

In my case, the default translation for pt-BR shipped in Moment.js resolves to “Dom 2ª 3ª 4ª 5ª 6ª Sáb”, very unusual here. A better format would be `ddd`.

I guess even in English or other languages, one could like to set `ddd` (or even `dddd` for full names). So a prop like `weekDayFormat` for components would be nice.

majapw:
This should be addressed in v12.6.0

PR #650, opened 2017-07-28
Add custom week day format

### LINKED ISSUE #453: custom weekday format on week header
weekday format seems to be hardcoded

`<small>{moment().weekday(i).format('dd')}</small>`

it would be good to be able to set this up in the config

schaable:
Add a prop to customize the day format in the header.

Resolves #453  and #132

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,7 @@ isDayHighlighted: PropTypes.func,
 // internationalization props
 displayFormat: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
 monthFormat: PropTypes.string,
+weekDayFormat: PropTypes.string,
 phrases: PropTypes.shape(getPhrasePropTypes(DateRangePickerPhrases)),
 ```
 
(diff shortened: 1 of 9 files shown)


ljharb:
Would you mind rebasing this on the command line so there's no merge commit?

schaable:
Done!

jseminck:
I would like the same functionality, but I think it's more consistent to provide a 1render()1 function rather than just specifying the format. The same idea applies for `renderDay` and `renderMonth` which already exist right now.

I have a branch in my fork that implements a `renderWeekHeaderDay()`  prop. Please have a look and let me know what you think: https://github.com/jseminck/react-dates/commit/494117bd86e0c0ad3812eea1d3a09afa4ae0d001

README.md
```diff
@@ -130,6 +130,7 @@ isDayHighlighted: PropTypes.func,
 // internationalization props
 displayFormat: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
 monthFormat: PropTypes.string,
+weekDayFormat: PropTypes.string,
```
ljharb: Is there a moment proptype we could use for "valid format"?

majapw: Technically I think any string is a valid format (you can include arbitrary strings by surrounding them with brackets)

PR #652, opened 2017-07-29
Recompute modifiers on date change, not just focus change

timhwang21:
(Re-filing because I seem to have majorly screwed something up in Git CLI 😓)

Addresses #570.

Currently, block / range related modifiers are only recomputed when the currently focused input changes. However, this causes some issues when more complex block / range functions are used, e.g. those that make selectable dates conditional on current state.

This commit gives one potential fix, which is to recompute modifiers not just on focus change, but also on start / end date change. This DOES result in a slight perf cost because modifiers are computed more frequently; however from my experience perf bottlenecks have mainly arisen on month changes (where 30 new days are added and recomputed) and on hover changes (fixed in v11.x), not on selection change.

The added story gives an example of a case where range modifier computation does not work as expected. The added wrapper component has the following validation: start dates must be no earlier than 1 week after selected end date, and end dates must be no later than 1 week after selected start date. Without the change in `DayPickerRangeController`, the following actions lead to a bug:

1. Pick a start date. Note the modifiers correctly recompute after the end date is focused.
2. Click on a date BEFORE the start date. `react-dates` behavior is to change the currently selected start date. However, note that the modifiers DO NOT recompute.

Additionally, there is a similar bug when `keepOpenOnDateSelect` is enabled, which is more noticeable with the `SingleDatePicker`. Because modifiers recompute when the focused input toggles between `null` and the input, modifiers are never recomputed until the calendar is closed and reopened. The story in `SingleDatePicker` showcases this -- when the change in `SingleDatePicker.jsx` is undone, the modifier never recomputes.

If this change seems ok I'll delete the changes to unrelated files, and add test cases. Thanks!

### DIFF
--- a/test/components/DayPickerRangeController_spec.jsx
+++ b/test/components/DayPickerRangeController_spec.jsx
@@ -882,6 +882,7 @@ describe('DayPickerRangeController', () => {
                 expect(visibleDays[monthString][dateString]).to.include('blocked-minimum-nights');
                 day.add(1, 'day');
               }
+          });
             });
 
             it('updates state to include `blocked` on the appropriate days', () => {
@@ -912,7 +913,7 @@ describe('DayPickerRangeController', () => {
       });
 
       describe('blocked-out-of-range', () => {
-        describe('focusedInput did not change', () => {
+        describe('focusedInput, startDate, endDate did not change', () => {
           it('does not call isOutsideRange if unchanged', () => {
             const isOutsideRangeStub = sinon.stub();
             const wrapper = shallow(<DayPickerRangeController
@@ -938,7 +939,7 @@ describe('DayPickerRangeController', () => {
           });
         });
 
-        describe('focusedInput changed', () => {
+        describe('focusedInput, startDate, endDate changed', () => {
           const numVisibleDays = 3;
           let visibleDays;
           beforeEach(() => {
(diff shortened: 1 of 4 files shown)


test/components/DayPickerRangeController_spec.jsx
```diff
@@ -838,6 +838,32 @@ describe('DayPickerRangeController', () => {
             expect(isBlockedStub.callCount).to.equal(numVisibleDays);
           });
 
+          it('calls isBlocked for every visible day when startDate changes', () => {
```
timhwang21: Note: I only added test cases to verify the call to the modifier function, because the code path followed when `startDate` / `endDate` change is otherwise identical to when `focusedInput` changes. I felt duplicating all the tests, with only the prop changing, would be low reward.

timhwang21:
Hm, maybe if needed this change and that in #668 can depend on a boolean prop that defaults to true? So if a user is certain that their use case does not require these checks, they can explicitly disable this behavior. Just thinking out loud...

ljharb:
@timhwang21 boolean props should generally always default to false, so if you want the default behavior to be "on", then it'd be better to pick a prop name where "false" means "on" :-)

mikkelwf:
Any update to if this is going to be merged in..?

ljharb:
@timhwang21 sorry for the long delay :-) i've rebased this; could you take another look and lmk if you think it's still landable?

test/components/DayPickerRangeController_spec.jsx
```diff
@@ -882,6 +882,7 @@ describe('DayPickerRangeController', () => {
                 expect(visibleDays[monthString][dateString]).to.include('blocked-minimum-nights');
                 day.add(1, 'day');
               }
+          });
```
timhwang21, line 885: Note to self: remove

timhwang21:
Wow what a throwback. It turns out I still make extensive use of this library and have retained _some_ idea of what's going on 😆 I fixed what appeared to be a minor rebase issue; everything else LGTM. But it would probably be good to get a sanity check from someone who's worked in this repo in the past 3 years!

ljharb:
@timhwang21 unfortunately that's basically nobody :-) i'm trying to pick it back up.

timhwang21:
I rebased on the main branch and force pushed, which seems to have fixed the issue with the automatic rebase pipeline.

PR #659, opened 2017-08-03
Support changing withPortal/withFullScreenPortal

mikelambert:
I am attempting to set `withFullScreenPortal` in response to the window size (so it only applies on narrow screens). Unfortunately, if `responsivizePickerPosition` is ever called in widescreen (non portal) mode, and the user shrinks the window width....it will now refrain from ever updating/clearing the `dayPickerContainerStyles` (since its a portal!). This bug causes the portal to be displayed with an offset, partially offscreen.


mikelambert:
Good point, updated SingleDatePicker.jsx too.

PR #664, opened 2017-08-05
updating propType usage to nonNegativeInteger()

palashkaria:
– fixes proptype warnings

see [this comment](https://github.com/airbnb/react-dates/pull/406#discussion_r131527804) 

example:
```sh
Warning: CalendarDay: type specification of prop `daySize` is invalid; the type checker function must return `null` or an `Error` but returned a function. You may have forgotten to pass an argument to the type checker creator (arrayOf, instanceOf, objectOf, oneOf, oneOfType, and shape all require an argument).
```

### DIFF
--- a/src/components/CalendarDay.jsx
+++ b/src/components/CalendarDay.jsx
@@ -14,7 +14,7 @@ import { BLOCKED_MODIFIER, DAY_SIZE } from '../../constants';
 
 const propTypes = forbidExtraProps({
   day: momentPropTypes.momentObj,
-  daySize: nonNegativeInteger,
+  daySize: nonNegativeInteger(),
   isOutsideDay: PropTypes.bool,
   modifiers: PropTypes.instanceOf(Set),
   isFocused: PropTypes.bool,
--- a/src/components/CalendarMonth.jsx
+++ b/src/components/CalendarMonth.jsx
@@ -33,7 +33,7 @@ const propTypes = forbidExtraProps({
   enableOutsideDays: PropTypes.bool,
   modifiers: PropTypes.object,
   orientation: ScrollableOrientationShape,
-  daySize: nonNegativeInteger,
+  daySize: nonNegativeInteger(),
   onDayClick: PropTypes.func,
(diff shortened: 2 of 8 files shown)


ljharb:
Hmm, the tests are failing tho. Are you sure this is the right fix?

palashkaria:
@ljharb In the tranpirde code, if i change 

```daySize: _airbnbPropTypes.nonNegativeInteger```

to

```daySize: _airbnbPropTypes.nonNegativeInteger()```

I stop getting the warnings

But this currently converts to 
```daySize: (0, _airbnbPropTypes.nonNegativeInteger)()```

which doesn't fix the issue. So yes, this PR is wrong, but the issue exists. Any ideas on how to solve this? I'm using this in a project and it's logging 100s of warnings.

I'm on React 15.4.1. Thanks :)

**Comment by ljharb** [comment]
What's the value of the `daySize` prop there? If it's a function, what's the toString of that function?

**Comment by palashkaria** [comment]
daySize is an integer, value is '36'

**Comment by ljharb** [comment]
What's the output of `propTypes.daySize + ''` tho?

**Comment by palashkaria** [comment]
Oh, sorry for that, here's the value of propTypes.daySize:

```
function noopThunk() {
  return noop;
}
```

ljharb:
Are you running this in production? airbnb-prop-types returns mocks in production (based on NODE_ENV). However, React doesn't check PropTypes in production either.

I do see that perhaps there's a bug in airbnb-prop-types in that the mock for nonNegativeInteger might need to be fixed - I'll take a look at that today.

palashkaria:
@ljharb thanks, looks like that was the issue. NODE_ENV was being set to production in my config, fixing that works. But yes, something wrong with the mocks here. React still runs the proptypes check though, I'm guessing that because it uses some other variable?

Anyway, this can be closed here

ljharb:
@palashkaria it means you're not using the minified version of React along with NODE_ENV as production.

I'll close this when I fix airbnb-prop-types; thanks!

PR #666, opened 2017-08-07
Added OutsideClickHandler component with onOutsideClick callback

mattRoyten:
Corrected onOutsideClick functionality missing from the DayPickerSingleDateController component. This prop was within the defaultProps and now is properly applied.

### DIFF
--- a/src/components/DayPickerSingleDateController.jsx
+++ b/src/components/DayPickerSingleDateController.jsx
@@ -28,6 +28,7 @@ import {
 } from '../../constants';
 
 import DayPicker from './DayPicker';
+import OutsideClickHandler from './OutsideClickHandler';
 
 const propTypes = forbidExtraProps({
   date: momentPropTypes.momentObj,
@@ -105,7 +106,7 @@ const defaultProps = {
 
   onPrevMonthClick() {},
   onNextMonthClick() {},
-  onOutsideClick() {},
+  onOutsideClick: null,
 
   renderDay: null,
   renderCalendarInfo: null,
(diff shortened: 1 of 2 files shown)


src/components/DayPickerSingleDateController.jsx
```diff
@@ -548,39 +549,44 @@ export default class DayPickerSingleDateController extends React.Component {
...
-        renderCalendarInfo={renderCalendarInfo}
-        isFocused={isFocused}
-        getFirstFocusableDay={this.getFirstFocusableDay}
-        onBlur={this.onDayPickerBlur}
-        phrases={phrases}
-        daySize={daySize}
-        isRTL={isRTL}
-      />
+      <OutsideClickHandler
+        onOutsideClick={onOutsideClick}
```
ljharb: if `onOutsideClick` is omitted, should the DayPicker still be wrapped?

mattRoyten: I do not see any negative impact if `onOutsideClick` is omitted. Are you thinking to conditionally wrap the DayPicker component depending on the `onOutsideClick` value?

ljharb: Yes, that way if people don't want outside click behavior, they wouldn't incur any overhead.

mattRoyten:
@ljharb - Updated PR with suggestion.

src/components/DayPickerSingleDateController.jsx
```diff
@@ -548,39 +549,47 @@ export default class DayPickerSingleDateController extends React.Component {
       isFocused,
       isRTL,
       phrases,
+      onOutsideClick,
     } = this.props;
 
     const { currentMonth, visibleDays } = this.state;
 
+    const DayPickerWrapper = (onOutsideClick) ? OutsideClickHandler : 'div';
```
ljharb: Let's remove these unnecessary parents on the ternary conditionals.

src/components/DayPickerSingleDateController.jsx
```diff
@@ -548,39 +549,47 @@ export default class DayPickerSingleDateController extends React.Component {
...
+    const wrapperProps = (onOutsideClick) ? { onOutsideClick } : {};
```
ljharb: This shouldn't be needed; `onOutsideClick={null}` on a DOM element is a no-op.

majapw:
I'm interested in why the refactor failed. I too think I would prefer what @ljharb suggested and don't know why that would cause tests to fail.

mattRoyten:
@majapw - all of the DayPickerSingleDateController tests were getting the following error:
_Invariant Violation: Unexpected node: [object Object]_

ljharb:
@mattRoyten You had an error in your code, in the unwrapped case, you were returning an object (ie, `{ dayPickerComponent }` instead of directly returning `dayPickerComponent`.

Could you rebase on the command line so as to remove all merge commits, and also remove the last revert commit, and fix that error?

mattRoyten:
@ljharb - got it. Updated.

src/components/DayPickerSingleDateController.jsx
```diff
@@ -565,40 +566,51 @@ export default class DayPickerSingleDateController extends React.Component {
...
-        renderCalendarInfo={renderCalendarInfo}
-        isFocused={isFocused}
-        getFirstFocusableDay={this.getFirstFocusableDay}
-        onBlur={this.onDayPickerBlur}
-        phrases={phrases}
-        daySize={daySize}
-        isRTL={isRTL}
-      />
-    );
+    const dayPickerComponent = (<DayPicker
```
ljharb: one last nit: the `(` should end the line, and the `);` should begin one - iow:
```jsx
const foo = (
  <div>
  </div>
);
```

src/components/DayPickerSingleDateController.jsx
```diff
@@ -599,6 +601,18 @@ export default class DayPickerSingleDateController extends React.Component {
         isRTL={isRTL}
       />
     );
+
+    if (onOutsideClick) {
```
majapw: Oh this looks *so* much better

mattRoyten:
@majapw - tests added!

test/components/DayPickerSingleDateController_spec.jsx
```diff
@@ -1401,4 +1402,22 @@ describe('DayPickerSingleDateController', () => {
...
   });
+
+  describe('onOutsideClick', () => {
+    it('should render OutsideClickHandler as it has onOutsideClick prop', () => {
+      const wrapper = shallow(
+        <DayPickerSingleDateController
+          onOutsideClick={() => null}
+        />,
+      );
+      expect(wrapper.find(OutsideClickHandler).length).to.equal(1);
```
ljharb: `expect(wrapper.find(OutsideClickHandler).to.have.lengthOf(1)` has a nicer failure message; nbd tho :-)

mattRoyten: sure, want me to change this as well?

ljharb: Sure, that'd be great :-D

mattRoyten: done! 💥

PR #667, opened 2017-08-08
Pass currentMonth into onPrevMonthClick and onNextMonthClick

MathSquared:
Sometimes it's helpful to know what month is currently showing on the calendar (for instance, to fetch data on blocked or highlighted dates for upcoming months). This change adds the current month as an argument to the `onPrevMonthClick` and `onNextMonthClick` callbacks, allowing parent components to know the currently-displayed month.

This change also adds Mocha specs, which pass (I hope).

Alex Meed

### DIFF
--- a/src/components/DayPickerRangeController.jsx
+++ b/src/components/DayPickerRangeController.jsx
@@ -500,7 +500,7 @@ export default class DayPickerRangeController extends React.Component {
       },
     });
 
-    onPrevMonthClick(newCurrentMonth);
+    onPrevMonthClick(newCurrentMonth.clone());
   }
 
   onNextMonthClick() {
@@ -524,7 +524,7 @@ export default class DayPickerRangeController extends React.Component {
       },
     });
 
-    onNextMonthClick(newCurrentMonth);
+    onNextMonthClick(newCurrentMonth.clone());
   }
 
   onMultiplyScrollableMonths() {
--- a/src/components/DayPickerRangeController.jsx
+++ b/src/components/DayPickerRangeController.jsx
@@ -499,7 +499,7 @@ export default class DayPickerRangeController extends React.Component {
       },
     });
 
-    onPrevMonthClick();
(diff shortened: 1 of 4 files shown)


src/components/DayPickerRangeController.jsx
```diff
@@ -499,7 +499,7 @@ export default class DayPickerRangeController extends React.Component {
       },
     });
 
-    onPrevMonthClick();
+    onPrevMonthClick(this.state.currentMonth.clone());
```
majapw: So `currentMonth` is already available to you as a destructured aspect of state. What you probably want to pass down is the new month anyways, no?

if you expect `this.state.currentMonth` to equal `currentMonth.clone().subtract(1, 'month')` at this point, you're probably going to have a bad time (setting state is not synchronous in the way you are expecting).  Better to set that to a separate variable and then update and also pass back.

MathSquared: I was intending to pass the new month.

✅ &nbsp;Fixed, fd064d4.

src/components/DayPickerRangeController.jsx
```diff
@@ -491,15 +491,16 @@ export default class DayPickerRangeController extends React.Component {
     const prevMonth = currentMonth.clone().subtract(2, 'months');
     const prevMonthVisibleDays = getVisibleDays(prevMonth, 1, enableOutsideDays, true);
 
+    const newCurrentMonth = currentMonth.clone().subtract(1, 'month');
     this.setState({
-      currentMonth: currentMonth.clone().subtract(1, 'month'),
+      currentMonth: newCurrentMonth,
```
majapw: So it miiiight be a good idea to return a different clone to the callback as the one in the state... just in case.

MathSquared: Good catch.&nbsp; ✅ &nbsp;Fixed, 1c4bc4e.

PR #668, opened 2017-08-08
Recalculate blocked days when day-blocking function prop changes

MathSquared:
Sometimes it's useful to force the picker to recompute which days should be blocked or highlighted (for instance, if we need to force the set of blocked days to change based on data fetched asynchronously). This change forces the picker to refetch the blocked/highlighted modifiers for all visible dates when the `isOutsideRange`, `isDayBlocked`, or `isDayHighlighted` props change.

Alex Meed

### DIFF
--- a/src/components/DayPickerRangeController.jsx
+++ b/src/components/DayPickerRangeController.jsx
@@ -309,7 +309,7 @@ export default class DayPickerRangeController extends React.Component {
             modifiers = this.deleteModifier(modifiers, momentObj, 'blocked');
           }
 
-          if (recomputeOutsideRange) {
+          if (didFocusChange || recomputeOutsideRange) {
             if (isOutsideRange(momentObj)) {
               modifiers = this.addModifier(modifiers, momentObj, 'blocked-out-of-range');
             } else {
@@ -317,7 +317,7 @@ export default class DayPickerRangeController extends React.Component {
             }
           }
 
-          if (recomputeDayBlocked) {
+          if (didFocusChange || recomputeDayBlocked) {
             if (isDayBlocked(momentObj)) {
               modifiers = this.addModifier(modifiers, momentObj, 'blocked-calendar');
             } else {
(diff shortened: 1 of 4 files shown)


timhwang21:
This is sweet, would love to see this. Falls in the same vein as #652

src/components/DayPickerRangeController.jsx
```diff
@@ -187,16 +187,21 @@ export default class DayPickerRangeController extends React.Component {
     } = nextProps;
     let { visibleDays } = this.state;
 
+    let recomputeBlockedDays = false;
```
majapw: I'd rather have this be named something like `recomputePropModifiers` since it is recomputing more than just the blocked days.

majapw: Or maybe we even want to separate out this logic such that we're only recalculating the functions that we need to? (e.g. only recompute outside days, blocked days, etc.)

MathSquared:
Sounds reasonable—made such a change in e17b8a1.

src/components/DayPickerRangeController.jsx
```diff
@@ -299,22 +309,28 @@ export default class DayPickerRangeController extends React.Component {
             modifiers = this.deleteModifier(modifiers, momentObj, 'blocked');
           }
 
-          if (isOutsideRange(momentObj)) {
-            modifiers = this.addModifier(modifiers, momentObj, 'blocked-out-of-range');
-          } else {
-            modifiers = this.deleteModifier(modifiers, momentObj, 'blocked-out-of-range');
+          if (recomputeOutsideRange) {
```
majapw: One final nit - to maintain the behavior from before, we want this to be `if (recomputeOutsideRange || didFocusChange) {`. Same for all the ones below.

MathSquared: ✅ &nbsp;Fixed (hoping the superfluous merge commit cece6c0 doesn't screw anything up).

majapw:
if you rebase/squash it could clean up the commits, if you like.

Issue #622, opened 2017-07-12
Clear hovered-span modifier when re-rendered with no start and end date


jseminck:
First of all thanks for the amazing repo. This is one of the most advanced dae-pickers that we have worked with and we were happy to recently migrate to using `react-dates`.

We currently ran into an issue (because we use the date-picker in a non-standard way) where we need to reset the state of the date-picker if the user performs some action.

Unfortunately, currently when re-rendering the date-picker with `startDate` and `endDate` as undefined, the previously selected days remain with the css class `CalendarDay--hovered-span`, even though they are currently not being hovered because nothing should be selected.

I played around with the code and managed to fix it by introducing the following code-block in `DayPickerRangeController` at line 223 inside the `componentWillReceiveProps` block.

```js
    let modifiers = {};

    if (!startDate && !endDate) {
      modifiers = this.deleteModifierFromRange(
        modifiers,
        initialVisibleMonth(),
        initialVisibleMonth().add(4, 'months').endOf('month'),
        'hovered-span',
      );
    }
```

This basically resets the `hovered-span` classes whenever the date-picker is re-rendered with empty start and end date.

I'm wondering if this is a valid approach (and if you would accept a PR), or if there is another/better way to achieve this.

Thanks

majapw:
These seems legit to me! There's no reason for `hovered-span` to be applied if there is no startDate  or endDate. 

It does seem a bit expensive to clear all the months every time this state is reached. Can you give a little bit more context to what leads you to clear the dates while the user is presumably hovering over the calendar?

jseminck:
Thanks for the quick reply.

As I mentioned in the original post, we use the datepicker in a bit of a non-standard way: we let the user select 2 date ranges within the same datepicker. Once the first range is selected, we store the range in our state and use the `renderDay()` hook to colour those days differently.

Then  we allow the user to select the second range. Once the second range is selected, the user gets redirected to another page with info regarding those two ranges. This is the happy path. 😄 

However, there is a possibility for the user to "reset" his first selected range (in case it was incorrect). In this case, he can clear his first range. However, because of the `--hovered-span` not disappearing, those styles will still be applied. 

I suppose they are also still applied to the first range while user is selecting the second range, but because we use `renderDay()` to override styles for values in the first range, it is not affecting us.

I should probably try and create a simple example to illustrate this use-case, as I find it very hard to explain it with words 😕

jseminck:
[...]

PR #680, opened 2017-08-23
Clear hovered span when there is no start and end date selected

jseminck:
My issue and my use case is described in: https://github.com/airbnb/react-dates/issues/622

Basically, in short, I would like to be able to clear all the hovered--span classes from the datepicker days when the `startDate` and `endDate` are not defined. 

Performance wise, I think this shouldn't affect anything in the majority of cases. The only way to get `startDate` and `endDate` to be not defined is:

* When the datepicker first renders. But there are no classes applied yet at this point, so removing the modifiers should be a noop (or close to a noop).
* When a "clear" button is implemented (like in our case), in which case I currently don't see a more performant way to do this.

This PR is not ready for merging yet as it is missing tests, but probably adding tests for this won't be so easy 😄 I will of course add tests if the proposed fix is approved.

### DIFF
--- a/src/components/DayPickerRangeController.jsx
+++ b/src/components/DayPickerRangeController.jsx
@@ -234,15 +234,6 @@ export default class DayPickerRangeController extends React.Component {
 
     let modifiers = {};
 
-    if (modifiers.length && !startDate && !endDate) {
-      modifiers = this.deleteModifierFromRange(
-        modifiers,
-        initialVisibleMonth(),
-        initialVisibleMonth().add(numberOfMonths, 'months').endOf('month'),
-        'hovered-span',
-      );
-    }
-
     if (didStartDateChange) {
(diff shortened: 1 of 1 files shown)


angusfretwell:
Would be nice to get this merged. I'm having the same problem when clearing dates in `onDatesChange`

src/components/DayPickerRangeController.jsx
```diff
@@ -254,6 +254,15 @@ export default class DayPickerRangeController extends React.Component {
         );
       }
 
+      if (!startDate && !endDate) {
+        modifiers = this.deleteModifierFromRange(
+          modifiers,
+          this.state.currentMonth,
```
cyancey: I had to change this to:
```
this.state.currentMonth.clone().startOf('month')
``` 

`this.currentMonth` is a specific date and when I ran this locally, it wouldn't clear days in subsequent months properly if this wasn't pegged to the beginning of the `currentMonth`.

PR #681, opened 2017-08-23
Fix: use constructor-bound methods for ref callbacks

### LINKED ISSUE #437: TypeError: Cannot read property 'contains' of null
What error is this? How to solve?

```
OutsideClickHandler.js:1 Uncaught TypeError: Cannot read property 'contains' of null
    at OutsideClickHandler.onOutsideClick (webpack:///./~/react-dates/lib/components/OutsideClickHandler.js?:1:3926)
    at eval (webpack:///./~/consolidated-events/lib/TargetEventHandlers.js?:1:1684)
    at Array.forEach (native)
    at TargetEventHandlers.handleEvent (webpack:///./~/consolidated-events/lib/TargetEventHandlers.js?:1:1620)
```

ljharb:
Use constructor-bound methods for ref callbacks instead of arrow functions.

Because this avoids creating a new function every render, it's both more performant, and also ensures React never sees two `!==` ref callbacks, which thus causes it to call the old one with `null` and the new one with the ref, which can sometimes cause issues like #437.

Separately, DayPickerKeyboardShortcuts has a ref but was an SFC, which makes no sense - so this also fixes it to be a class-based component.

Fixes #437.

### DIFF
--- a/src/components/CalendarDay.jsx
+++ b/src/components/CalendarDay.jsx
@@ -45,6 +45,12 @@ const defaultProps = {
 };
 
 export default class CalendarDay extends React.Component {
+  constructor(...args) {
+    super(...args);
+
+    this.setButtonRef = this.setButtonRef.bind(this);
+  }
+
   shouldComponentUpdate(nextProps, nextState) {
     return shallowCompare(this, nextProps, nextState);
   }
@@ -71,6 +77,10 @@ export default class CalendarDay extends React.Component {
   onDayMouseLeave(day, e) {
     const { onDayMouseLeave } = this.props;
     onDayMouseLeave(day, e);
+  }
+
+  setButtonRef(ref) {
+    this.buttonRef = ref;
   }
 
   render() {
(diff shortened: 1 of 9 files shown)


src/components/DateRangePicker.jsx
```diff
@@ -116,6 +116,9 @@ export default class DateRangePicker extends React.Component {
     this.showKeyboardShortcutsPanel = this.showKeyboardShortcutsPanel.bind(this);
 
     this.responsivizePickerPosition = this.responsivizePickerPosition.bind(this);
+
+    this.setDayPickerContainerRef = this.setDayPickerContainerRef.bind(this);
+    this.setDayPickerRef = this.setDayPickerRef(this);
```
calinoracation: Looks to be missing a `.bind`.

ljharb: whoops, thanks!

src/components/CalendarDay.jsx
```diff
@@ -112,7 +122,7 @@ export default class CalendarDay extends React.Component {
       <td className={className} style={daySizeStyles}>
         <button
           type="button"
-          ref={(ref) => { this.buttonRef = ref; }}
+          ref={this.setButtonRef}
```
majapw: This is a silly effect of react

ljharb: agreed

Issue #514, opened 2017-05-16
Synthetic event warning onKeyDown


JemarJones:
With the latest version of the daterangepicker i get the following warnings logged if i for example click into the start date input, type something and hit `cmd` + `A`.

```
Warning: This synthetic event is reused for performance reasons. If you're seeing this, you're accessing the method `stopPropagation` on a released/nullified synthetic event. This is a no-op function. If you must keep the original synthetic event around, use event.persist(). See https://fb.me/react-event-pooling for more information.

printWarning	@	warning.js?8a56:36
warning	@	warning.js?8a56:60
warn	@	SyntheticEvent.js?2535:265
get	@	SyntheticEvent.js?2535:259
onKeyDown	@	DateInput.js?2f69:265
invokeFunc	@	index.js?f580:160
trailingEdge	@	index.js?f580:207
timerExpired	@	index.js?f580:195
```

```
Warning: This synthetic event is reused for performance reasons. If you're seeing this, you're accessing the method `key` on a released/nullified synthetic event. This is a no-op function. If you must keep the original synthetic event around, use event.persist(). See https://fb.me/react-event-pooling for more information.

printWarning	@	warning.js?8a56:36
warning	@	warning.js?8a56:60
warn	@	SyntheticEvent.js?2535:265
get	@	SyntheticEvent.js?2535:259
onKeyDown	@	DateInput.js?2f69:272
invokeFunc	@	index.js?f580:160
trailingEdge	@	index.js?f580:207
timerExpired	@	index.js?f580:195
```

I notice that you're throttling the `onKeyDown` event handler, so it may be that the event is nullified by the time the handler is actually called, as described here http://stackoverflow.com/questions/35435074/using-debouncer-with-react-event. This makes sense to me as the warning is being thrown on the trailing edge of the throttle.

JemarJones:
I can confirm that modifying the source and setting the `trailing` to `false` does fix the issue. I believe this would result in onKeyDown only being called on the leading edge. Another possible solution would be wrapping the call to the throttled handler and persisting the event.

basarat:
Verified incident report in our code base :rose:: 

![image](https://cloud.githubusercontent.com/assets/874898/26538532/c057d172-448a-11e7-84c7-daf63e45bb59.png)

mikelambert:
I just stumbled on this warning message in my codebase, and tracked it down to react-dates.

[...]

PR #682, opened 2017-08-24
Fix React warning when Events are referenced later

mikelambert:
Fixes https://github.com/airbnb/react-dates/issues/514 , which happens due to Events being stored to be delivered on the trailing edge of a keyDown. I believe we don't need the trailing edge, and it is better than a persisted Event.

### DIFF
--- a/src/components/DateInput.jsx
+++ b/src/components/DateInput.jsx
@@ -20,6 +20,8 @@ const FANG_STROKE_TOP = `M0,${FANG_HEIGHT_PX} ${FANG_WIDTH_PX / 2},0 ${FANG_WIDT
 const FANG_PATH_BOTTOM = `M0,0 ${FANG_WIDTH_PX},0 ${FANG_WIDTH_PX / 2},${FANG_HEIGHT_PX}z`;
 const FANG_STROKE_BOTTOM = `M0,0 ${FANG_WIDTH_PX / 2},${FANG_HEIGHT_PX} ${FANG_WIDTH_PX},0`;
 
+const MODIFIER_NAMES = ['Shift', 'Control', 'Alt', 'Meta'];
+
 const propTypes = forbidExtraProps({
   ...withStylesPropTypes,
   id: PropTypes.string.isRequired,
@@ -127,7 +129,7 @@ class DateInput extends React.Component {
 
   onKeyDown(e) {
     e.stopPropagation();
-    if (!['Shift', 'Control', 'Alt', 'Meta'].includes(e.key)) {
+    if (!MODIFIER_NAMES.includes(e.key)) {
       this.throttledKeyDown(e);
     }
   }
--- a/src/components/DateInput.jsx
+++ b/src/components/DateInput.jsx
@@ -209,7 +209,7 @@ class DateInput extends React.Component {
           ref={this.setInputRef}
           value={value}
           onChange={this.onChange}
-          onKeyDown={throttle(this.onKeyDown, 300)}
+          onKeyDown={throttle(this.onKeyDown, 300, {trailing: false})}
           onFocus={onFocus}
           placeholder={placeholder}
           autoComplete="off"
--- a/src/components/DateInput.jsx
+++ b/src/components/DateInput.jsx
@@ -84,6 +84,7 @@ class DateInput extends React.Component {
     this.onChange = this.onChange.bind(this);
     this.onKeyDown = this.onKeyDown.bind(this);
     this.setInputRef = this.setInputRef.bind(this);
+    this.throttledKeyDown = throttle(this.onFinalKeyDown, 300, { trailing: false });
   }
 
   componentDidMount() {
@@ -126,15 +127,20 @@ class DateInput extends React.Component {
 
   onKeyDown(e) {
     e.stopPropagation();
+    if (!['Shift', 'Control', 'Alt', 'Meta'].includes(e.key)) {
+      this.throttledKeyDown(e);
+    }
+  }
 
+  onFinalKeyDown(e) {
     const {
       onKeyDownShiftTab,
       onKeyDownTab,
       onKeyDownArrowDown,
       onKeyDownQuestionMark,
     } = this.props;
-
     const { key } = e;
+
     if (key === 'Tab') {
       if (e.shiftKey) {
         onKeyDownShiftTab(e);
@@ -209,7 +215,7 @@ class DateInput extends React.Component {
           ref={this.setInputRef}
           value={value}
           onChange={this.onChange}
-          onKeyDown={throttle(this.onKeyDown, 300, {trailing: false})}
+          onKeyDown={this.onKeyDown}
           onFocus={onFocus}
           placeholder={placeholder}
           autoComplete="off"


mikelambert:
Please don't approve/submit yet. My fix may not be correct, but the existing code is also broken.

When a throttled function, you have the leading edge (the first time it's called), the inprogress events (that only get called-through if more than 300ms have elapsed since the last one), and then the trailing edge (which happens 300ms after the last call, if there have been no intermediate calls, and represents the last call that was received and buffered while waiting for 300ms to elapse.) There are probably lodash throttle/debounce documentation/stackoverflows that explain this better than I can. :)

Regardless, the trailing edge is delivered and called some number of milliseconds *after* the event was fired. This means that React has already cleared the event and it back into the pool for later reuse. So at this point, it is too late to call `e.persist()`, as the data has been lost.

In my pull request assumed that a series of keyDown events would all be identical, and so we could ignore the trailing edge. I was wrong. Consider this sequence of events:
a) hit Shift
b) now while Shift is pressed, hit Tab

If (b) happens within 300ms of (a), it will not be delivered. With the original code, (b) will be cleared, and be a dummy event by the time it is delivered. With my code, (b) will never be delivered. In both cases, the use of throttle causes (b) to not be delivered at all.

If you access the storybook and quickly hit shift-tab out of the Start field, the Picker will incorrectly remain open. If you wait a bit between shift-and-tab, it will correctly close.

I believe the correct fix here is to do one of the following:
1) turn off throttle entirely (I am curious if you can explain why it was necessary in the first place?)
2) implement `throttle` manually, such that instead of the default of saving arguments for later delivery...it actually calls `event.persist()` and saves them for later delivery

Regardless, my pull request as it exists is not correct, and is "as broken" as it was before (though with less React console warnings :P )

ljharb:
I think perhaps it might be better to move `throttle` usage inside the `onKeyDown` handler?

mikelambert:
Going with your approach, there are still some issues with modifier-only keydowns "eating up" the throttle and keeping modifier+real_key events from being delivered. So I moved the throttle into `onKeyDown`, and omit modifier-only events, and then throw the rest through the throttler, which should work.

At this point, I also turned off the trailing edge. This means if you hold down the "tab" key for more than the key-repeat-delay and less than 300 ms, one tab will be delivered. Unlike if we support trailing edges (the original default), where it would send *two* tab events (ie, delivering the second tab button). This meant a "second event" could occur up to 300ms after the user let go of the key, which seems like a Bad Idea to me.

On the bright side, this also means we can let the React `SyntheticEvents` go back into the pool for re-use, and we don't have to worry about GC-ing these `persist`-ed objects either, which should be a bit better performance-wise as well.

I have tested this code with a bunch of attempted-fast tab and shift-tab in the storybook, and it seems to work without any issues.

erin-doyle:
This is awesome @mikelambert thanks for submitting this!  These warnings have been driving me crazy but I haven't had the time to dig in and look for a fix.

erin-doyle:
There is also a throttled `onKeyDown` in `DayPicker.js` which manages full keyboard navigation capabilities within the Calendar area.  There is an `onKeyDown` handler in the `DayPickerKeyboardShortcuts.js` as well though it's not being throttled currently but it's possible that was just a miss.  Can you take a look at those as well?

erin-doyle:
@mikelambert I stumbled upon this again.  I'm noticing some serious lag at time in the key event handling along with the SyntheticEvent warnings and I really wonder if they'd all be fixed with this.  Can you please see the feedback I left back on Nov 17 and make those changes?  It would be great to get this merged!

erin-doyle:
@ljharb what do you think about us merging this as is and I can follow up with making these changes in the remaining components that need them?

ljharb:
Either way this still needs a rebase; and since this PR was from prior to the react-with-styles change, it might not be a trivial one.

erin-doyle:
Ok well I'm actively working right now on trying to improve some issues with the keyboard navigation and I think this could be the fix to the problems I'm seeing.  So I've already copied these changes into the branch I'm working on and making the same changes to the other components.  If I get done before this has been rebased then we can supersede this with my PR.

ljharb:
@erin-doyle if you cherry-pick these commits into your PR, that would work just fine :-)

erin-doyle:
Ok so I take it back, I'm not sure if we should make these changes exactly to DayPicker.jsx and DayPickerKeyboardShortcuts.jsx.

First for DayPickerKeyboardShortcuts.jsx:
When adding a throttle to keyDown what it ends up doing is not being able to preventDefault on the long list of keys that need to be prevented but we really have to run through the function to determine what key it is and what action to take.  It just doesn't make sense to me at that point to try to throttle anything here (and it's not being throttled at all now without issue).

For DayPicker.jsx:
In my testing it seems almost as if the throttle is not even really working as it is now.  I can hold down keys and they aren't being throttled, I can increase the delay time and nothing changes.  If I make the same changes as are here in this PR on DateInput.jsx, having onKeyDown call the throttle function, it definitely is throttling, such that the experience is kind of laggy.  I am really starting to wonder if we should not use throttling at least here in this component.  I would expect that most users navigating through the dates using arrow keys will be clicking on them fairly quickly (if they are not using a screen reader and waiting for the reading) and expecting them to be responsive.  Reducing the delay to 200 is a decent compromise.

mikelambert:
Hi everyone, sorry for being MIA on the change request. :/ I am glad to hear Erin was able to try these changes out, and will defer to her experience since I'm not as familiar with `DayPicker` stuff. (And I'm happy to get this one change submitted as-is and let others deal with proper `DayPicker` fixes, or let someone else own this change and get the combined set of proper changes in.)

As far as this particular pull request, I just did a git merge to head, and there were no issues. I am not sure why `react-with-styles` would cause any merge issues, since this is unrelated to CSS?


Unfortunately, my attempts to rebase against `master` are failing. It seems there are multiple tag versions all being committed into `master` branch (16.0.1 and 15.5.2 in the same master commit log?), and my git-fu is not good enough to figure out how to rebase against multiple tags with differing versions somehow coexisting in the same commit log... ie, my `git rebase master -i` shows:
```
...
pick 94e1852 Version 16.0.0
pick 4d3b312 Add missing onKeyDown function to CalendarDay
pick e57d119 Version 16.0.1
pick 636f619 Revert "Merge pull request #866 from airbnb/maja-make-drp-ids-required"
pick 065b6a3 Version 15.5.2
pick f48b820 Version 15.5.3
```

ljharb:
@mikelambert try `git rebase upstream/master -i` instead (where "upstream" is this repo)

mikelambert:
Ahhh great, thanks, makes sense! Rebased, and now needs re-approval.

src/components/DateInput.jsx
```diff
@@ -126,15 +127,20 @@ class DateInput extends React.Component {
 
   onKeyDown(e) {
     e.stopPropagation();
+    if (!['Shift', 'Control', 'Alt', 'Meta'].includes(e.key)) {
```
ljharb: for perf, this should probably be an object literal or a Set defined at the module level.

mikelambert: Fixed.

PR #683, opened 2017-08-24
Introduce renderWeekHeaderDay prop

jseminck:
This is not ready to be merged as there are still missing tests. I can add those if the suggestion is approved.

This is an alternative solution to an existing open PR: https://github.com/airbnb/react-dates/pull/650 - That PR introduces just a `dayFormat` prop that is passed on to formatting the week day headers. However, I thought it's more consistent to have a `render()` function as the same option is already provided to `renderDay()` and `renderMonth()`.

The `renderWeekHeaderDay()` function takes in the day of the week as a prop (0-6). Users can then use this to format it how they want, and additionally apply any styles as they would like.

I'm not sure about the function name either... perhaps there's something better 😄


jseminck:
Hi! Any chance to get this or https://github.com/airbnb/react-dates/pull/650 merged? I know this implementation is still missing tests, but I'll definitely add them if this solution is approved.

jseminck:
Nothing additional. Just consistency with other `renderXXX` functions. It could be used to apply some different styling if someone wanted to.

But on the other hand, I understand that the simplicity of https://github.com/airbnb/react-dates/pull/650 is much nicer to have, especially when no one has a specific use-case to render something completely different.

At least my case would be solved with https://github.com/airbnb/react-dates/pull/650 as well 😄

majapw:
In that case, let's go with #650 for now and if we come up with a usecase that needs additional formatting we can expand it then. :) Thank you for your contribution. I'll review the other PR now so we can get that in!

Issue #696, opened 2017-08-29
IE 11 Tab Issues


humphries40:
I've noticed that in IE 11 if you have selected a date, then click on the input box again and try and use tab or shift-tab nothing occurs. You're just trapped with the calendar input open.  

 I was going to attempt this on the storybook, but it seems that is broken. Supposedly storybook has been updated to fix this issue [here](https://github.com/storybooks/storybook/issues/1352),  So you may need to update your storybook package to the latest release to allow IE users to use the storybook.

majapw:
Oh good call! We'll do the update asap.

humphries40:
Sweet then I can check if the issue is happening there as well, or just in our solution.

majapw:
Hi @humphries40, I updated storybook so hopefully we should be able to test in IE now.

humphries40:
Alright, testing around in IE I couldn't replicate the issue that we're seeing in our application. I'll close this bug.

sjain-mi9:
@majapw, I have a similar issue with tabbing on IE. I have 2 datepickers and it tabs out perfectly on chrome, but fails on IE. I need to switch from one datepicker to another one using tab. On the storybook, it takes 2 tabs to focus on the other datepicker.

PR #703, opened 2017-09-06
Bump storybook version to support IE and change default port to 6006

majapw:
This is partly to help support debugging of IE issues, support for which should be included in the latest storybook version.

See conversation here - https://github.com/airbnb/react-dates/issues/696

to: @humphries40 @ljharb @airbnb/webinfra


majapw:
@ljharb the port is mostly for my convenience due to the number of storybooks I usually have running simultaneously X_x

Issue #713, opened 2017-09-14
[a11y] Disabled dates aren't actually disabled in markup


backwardok:
When a date is disabled, CSS is applied to make it look disabled (`CalendarDay--blocked-calendar`, `CalendarDay--blocked-out-of-range`):
![image](https://user-images.githubusercontent.com/6530351/30406041-1bbe212a-98a4-11e7-8dfd-8c497f65c581.png)

However, this date is still clickable with a mouse (even if the mouse click doesn't do anything in the UI), can be navigated to with a keyboard (you can arrow through disabled dates), and doesn't indicate that those dates are disabled to a screen reader user (because the `button` itself is not actually disabled). 

Example where you can see this: http://airbnb.io/react-dates/?selectedKind=DRP%20-%20Day%20Props&selectedStory=blocks%20fridays&full=0&down=1&left=1&panelRight=0&downPanel=storybook%2Factions%2Factions-panel

backwardok:
@majapw can you think of a reason why someone would _not_ want to have the `disabled` attribute on the `button` underlying the date?

Note: this likely means messing with how next/previous day handling works for months (and what does this mean if you open a month in the past where the past is disabled? Or should you just not be allowed to do that either?

majapw:
I think a couple of things are being conflated here.

1) I am surprised that the VO does not read out that the date is unavailable. That's like specifically what https://github.com/airbnb/react-dates/blob/master/src/components/CalendarDay.jsx#L113 is for. So that is a bug.

2) We can definitely disable the button for a disabled date. That was just an oversight on my part.

> this likely means messing with how next/previous day handling works for months
3) I don't understand why these things would be related. We have talked about possibly disabling month navigation when it would take you to exclusively blocked dates (https://github.com/airbnb/react-dates/issues/339 and I think there's another issue as well), but that seems a separate discussion from disabling the blocked buttons?

erin-doyle:
So I've submitted a PR to address the unavailable dates not being read as such by screen readers.  

I did also locally code the disabling of the buttons for the unavailable dates.  However, upon testing, it occurred to me that if the buttons are disabled then those users depending on screen readers would be blocked from being able to navigate over those dates and nothing would read to them why they were being blocked from moving back or forward any further.  It works great for sighted people but I wonder if it would be a confusing user experience for those that are not.

majapw:
[...]

PR #715, opened 2017-09-14
Some improvements for screen readers [a11y]

erin-doyle:
Doing some screen reader testing on this and identified some things that could make the reading a little more clear/helpful:

1. The table was being interpreted as a data table versus a layout table and was reading extraneous info about the number of columns and rows in the current cell happened to be.  The cell already nicely reads the full date and day of the week and therefore there's no need for the screen reader to associate the column headers with the cell.

1. The DayPickerRangeController did not seem to be properly setting the chooseAvailableStartDate phrase after first mounting so only after picking a start date and moving to the end date would the reader nicely read "it's available" for the date options.

1. I did also notice Issue #713 where unavailable dates are not reading as "unavailable" but I haven't figured out what's causing that yet and since it's already a known issue wanted to move forward with these changes before moving on to trying to help out with that one.

### DIFF
--- a/src/components/CalendarMonth.jsx
+++ b/src/components/CalendarMonth.jsx
@@ -132,10 +132,13 @@ export default class CalendarMonth extends React.Component {
 
     return (
       <div className={calendarMonthClasses} data-visible={isVisible}>
-        <table>
-          <caption className="CalendarMonth__caption js-CalendarMonth__caption">
-            <strong>{monthTitle}</strong>
-          </caption>
+        <div
+          id="CalendarMonth__caption"
+          className="CalendarMonth__caption js-CalendarMonth__caption"
+        >
+          <strong>{monthTitle}</strong>
+        </div>
+        <table role="presentation">
 
           <tbody className="js-CalendarMonth__grid">
             {weeks.map((week, i) => (
--- a/src/components/DayPicker.jsx
+++ b/src/components/DayPicker.jsx
@@ -825,6 +825,9 @@ export default class DayPicker extends React.Component {
       <div
         className={dayPickerClassNames}
         style={dayPickerStyle}
+        role="application"
+        aria-label="Calendar"
+        aria-describedby="CalendarMonth__caption"
       >
         <OutsideClickHandler onOutsideClick={onOutsideClick}>
           <div
--- a/src/components/DayPicker.jsx
+++ b/src/components/DayPicker.jsx
@@ -826,8 +826,7 @@ export default class DayPicker extends React.Component {
         className={dayPickerClassNames}
         style={dayPickerStyle}
         role="application"
-        aria-label="Calendar"
-        aria-describedby="CalendarMonth__caption"
+        aria-label={phrases.calendarLabel}
       >
         <OutsideClickHandler onOutsideClick={onOutsideClick}>
           <div
--- a/src/components/DayPickerRangeController.jsx
+++ b/src/components/DayPickerRangeController.jsx
@@ -132,6 +132,15 @@ const defaultProps = {
   isRTL: false,
 };
 
+const getChooseAvailableDatePhrase = (phrases, focusedInput) => {
+  if (focusedInput === START_DATE) {
+    return phrases.chooseAvailableStartDate;
+  } else if (focusedInput === END_DATE) {
+    return phrases.chooseAvailableEndDate;
+  }
+  return phrases.chooseAvailableDate;
+};
+
 export default class DayPickerRangeController extends React.Component {
   constructor(props) {
     super(props);
@@ -172,6 +181,22 @@ export default class DayPickerRangeController extends React.Component {
     this.onMultiplyScrollableMonths = this.onMultiplyScrollableMonths.bind(this);
     this.getFirstFocusableDay = this.getFirstFocusableDay.bind(this);
     this.setDayPickerRef = this.setDayPickerRef.bind(this);
+  }
+
+  componentDidMount() {
+    const { focusedInput } = this.props;
+    const { phrases } = this.state;
+
+    // initialize phrases
+    // set the appropriate CalendarDay phrase based on focusedInput
+    const chooseAvailableDate = getChooseAvailableDatePhrase(phrases, focusedInput);
+
+    this.setState({
+      phrases: {
+        ...phrases,
+        chooseAvailableDate,
+      },
+    });
   }
 
   componentWillReceiveProps(nextProps) {
@@ -356,12 +381,7 @@ export default class DayPickerRangeController extends React.Component {
 
     if (didFocusChange || phrases !== this.props.phrases) {
       // set the appropriate CalendarDay phrase based on focusedInput
-      let chooseAvailableDate = phrases.chooseAvailableDate;
-      if (focusedInput === START_DATE) {
-        chooseAvailableDate = phrases.chooseAvailableStartDate;
-      } else if (focusedInput === END_DATE) {
-        chooseAvailableDate = phrases.chooseAvailableEndDate;
-      }
+      const chooseAvailableDate = getChooseAvailableDatePhrase(phrases, focusedInput);
 
       this.setState({
         phrases: {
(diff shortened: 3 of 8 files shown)


src/components/CalendarMonth.jsx
```diff
@@ -132,10 +132,13 @@ export default class CalendarMonth extends React.Component {
...
-          <caption className="CalendarMonth__caption js-CalendarMonth__caption">
-            <strong>{monthTitle}</strong>
-          </caption>
+        <div
+          id="CalendarMonth__caption"
+          className="CalendarMonth__caption js-CalendarMonth__caption"
+        >
+          <strong>{monthTitle}</strong>
+        </div>
+        <table role="presentation">
```
erin-doyle: Added the role of presentation here to stop the screen reader from reading all of the details of the columns and rows of the table [WAI-ARIA use of presentation role](https://www.w3.org/TR/2009/WD-wai-aria-practices-20090224/#kbd_layout_impact_mode_css2).

src/components/CalendarMonth.jsx
```diff
@@ -132,10 +132,13 @@ export default class CalendarMonth extends React.Component {
...
-        <table>
-          <caption className="CalendarMonth__caption js-CalendarMonth__caption">
-            <strong>{monthTitle}</strong>
-          </caption>
+        <div
+          id="CalendarMonth__caption"
+          className="CalendarMonth__caption js-CalendarMonth__caption"
+        >
+          <strong>{monthTitle}</strong>
+        </div>
```
erin-doyle: Changed `<caption>` to `<div>` as apparently this is a structural type element that cues screen readers that this is a data vs layout type table (see NOTE on "layout tables" and the caption element here: [H39: Using caption elements to associate data table captions with data tables](https://www.w3.org/TR/WCAG20-TECHS/H39)).

See in DayPicker.jsx that it now points here for the `aria-describedby` and thus continues to read the title as it was before with the `<caption>`.

src/components/DayPicker.jsx
```diff
@@ -825,6 +825,9 @@ export default class DayPicker extends React.Component {
       <div
         className={dayPickerClassNames}
         style={dayPickerStyle}
+        role="application"
```
erin-doyle: Added a role of application here as it seemed to match up with the WAI-ARIA example shown here: [For application landmarks with static prose](https://www.w3.org/TR/2009/WD-wai-aria-practices-20090224/#kbd_layout_static).

src/components/DayPickerRangeController.jsx
```diff
@@ -174,6 +183,22 @@ export default class DayPickerRangeController extends React.Component {
     this.setDayPickerRef = this.setDayPickerRef.bind(this);
   }
 
+  componentDidMount() {
+    const { focusedInput } = this.props;
+    const { phrases } = this.state;
+
+    // initialize phrases
+    // set the appropriate CalendarDay phrase based on focusedInput
+    const chooseAvailableDate = getChooseAvailableDatePhrase(phrases, focusedInput);
```
erin-doyle: It seemed like the chooseAvailableDate phrase was only getting set on update of the focus or the phrases so I thought adding the setting of it in `componentDidMount` would address initialization.

test/components/SingleDatePickerInput_spec.jsx
```diff
@@ -21,41 +21,45 @@ describe('SingleDatePickerInput', () => {
   describe('clear date', () => {
     describe('props.showClearDate is falsey', () => {
       it('does not have .SingleDatePickerInput__clear-date class', () => {
-        const wrapper = shallow(<SingleDatePickerInput showClearDate={false} />);
+        const wrapper = shallow(<SingleDatePickerInput id="date" showClearDate={false} />);
```
erin-doyle: All the changes here were just to clean up some prop type warnings that were bugging me.  Hope that's ok!

majapw:
This is awesome! I will take a look at the code today. :) Thanks!

FYI @backwardok

src/components/DayPicker.jsx
```diff
@@ -825,6 +825,9 @@ export default class DayPicker extends React.Component {
...
+        aria-label="Calendar"
```
backwardok: We should make sure that this is something that can be localized

src/components/DayPicker.jsx
```diff
@@ -825,6 +825,9 @@ export default class DayPicker extends React.Component {
...
+        aria-describedby="CalendarMonth__caption"
```
backwardok: I'm curious if this could be confusing - since it will be most likely just read when the user comes to the calendar the first time, the description will always be the first month. They will come to an area that will read something like, "Calendar, application, September". Seems like a more appropriate description would be something like "Date picker"

erin-doyle: Hmm well, going back and testing this again to be sure, it seems as if it's not even reading the aria-describedby.  I must have assumed that when I heard "September 2017" it was the aria-describedby but really it was just the reading of the selected CalendarDay.  Upon changing the text it's now more obvious that it's not being read.  Perhaps it's just not supported by the browser screen reader combinations I'm testing on (Mac with VoiceOver on Chrome, Firefox, Safari)?  I so far can't find a good reason for why it's not being read... 

I wonder if it's really necessary or if the aria-label here and the aria-describedby on the input telling the user how to interact with the calendar is enough information?  If you think a describedby is crucial here let me know and I'll keep working on why it's not being read, otherwise I can just remove it.

erin-doyle: @majapw and @backwardok, I just wanted to bump this and ask again, what are your thoughts about whether to skip adding an aria-describedby on the role="application" div or to figure out why it's not being read and fix it.  If the answer is to add it and fix it, what should the text be?

backwardok: Sorry for the delay! I think it would be fine to bypass adding the aria-describedby here. If we were to have some sort of visual text that also explained the calendar, or there was text that explained some calendar instructions specifically for screen reader users, then that might be useful, but I think it's probably ok without it.

erin-doyle: No worries!  I have added the aria-label text to the localizable phrases and removed the aria-describedby.  

As for having something visual and read out loud explaining the calendar instructions, there already is the keyboard shortcuts modal which is opened by entering ? or clicking on the ? at the bottom of the calendar.  The instructions for getting to the keyboard shortcuts modal is read via the aria-describedby on the DateInput:

`Press the down arrow key to interact with the calendar and
  select a date. Press the question mark key to get the keyboard shortcuts for changing dates.`

So the instructions seem pretty clear and complete from my testing as a visually able, keyboard only, and screen-reader user.

ljharb:
@erin-doyle would you mind rebasing on the command line and force pushing, so as to remove any merge commits?

erin-doyle:
oh jeez I forget to rebase every time...  Ok it should be cleaned up now.

src/components/DayPickerRangeController.jsx
```diff
@@ -174,6 +183,22 @@ export default class DayPickerRangeController extends React.Component {
     this.setDayPickerRef = this.setDayPickerRef.bind(this);
   }
 
+  componentDidMount() {
```
backwardok: Since the determination of which phrase to show is based off of the props and doesn't rely on the actual state of the component (waiting for the input to have focus), you can do this logic in the constructor and set `phrases` to the new object in `this.state`. 

If you put that logic in `componentDidMount`, which gets called after `render`,  setting `state` will trigger another render.

ljharb: (it'd then also need the same logic in `componentWillReceiveProps`)

erin-doyle: Yep that totally makes sense, I've moved that logic to the constructor and removed `componentDidMount`.  As for `componentWillReceiveProps` that logic was already there.  The bit I added was to make sure that phrase was being initialized at component load time instead of having to wait until props changed.

src/components/DayPickerRangeController.jsx
```diff
@@ -166,10 +166,17 @@ export default class DayPickerRangeController extends React.Component {
 
     const { currentMonth, visibleDays } = this.getStateForNewMonth(props);
 
+    // initialize phrases
```
ljharb: anything that depends on props in the constructor also must happen in `componentWillReceiveProps`.

ljharb: whoops, sorry, i saw your message.

PR #722, opened 2017-09-22
Convert react-dates to use react-with-styles

majapw:
This change moves react-dates onto Airbnb's modern best practices of CSS in your Javascript! 

Because this is again a huge change, it is best viewed on a commit by commit level (which I've tried to split up by component). 

While this transformation of the code is complete, there are still a number of things to be done. 

- [X] Documentation! How do you use react-with-styles?
- [ ] More customization! I think we should identify overrides that might be useful in the theme since you can't go about overriding everything willy-nilly anymore
- [X] ~I think we might want to export a CSS-styles version of the code as well as the vanilla version~
- [X] TESTS DON'T WORK ANYMORE. I need to fix that.
- [X] Once tests are running, probably some of them will be broken
- [X] there is a weird flash before the height of the calendar month grid is set now... we should address that.

to: @ljharb @airbnb/webinfra @lencioni @adamrneary @sdjidjev


majapw:
@lencioni yup! That is on my TODO. :)

src/components/CalendarDay.jsx
```diff
@@ -118,12 +116,44 @@ export default class CalendarDay extends React.Component {
       height: daySize - 1,
     };
 
+    const useDefaultCursor =
```
gabergg: As we discussed, maybe worth converting these modifiers to boolean properties or something similar. Might be an easier interface + more intuitive.

src/shapes/withStylesPropTypes.js
```diff
@@ -0,0 +1,8 @@
+import PropTypes from 'prop-types';
```
gabergg: Why pull this out of react-with-styles?

css/styles.css
```diff
@@ -0,0 +1,18876 @@
...
+.DateRangePicker_picker {
+  z-index: 1;
+  background-color: #fff;
+  position: absolute
+}
+.DateRangePicker_picker__rtl {
+  direction: rtl
+}
+.DateRangePicker_picker__directionLeft {
+  left: 0px
```
ljharb: also that zero values have units

css/styles.css
```diff
@@ -0,0 +1,18876 @@
...
+.DateRangePicker_picker__rtl_5.DateRangePicker_picker__rtl_5.DateRangePicker_picker__rtl_5.DateRangePicker_picker__rtl_5.DateRangePicker_picker__rtl_5 {
+  direction: rtl
+}
+.DateRangePicker_picker__rtl_6.DateRangePicker_picker__rtl_6.DateRangePicker_picker__rtl_6.DateRangePicker_picker__rtl_6.DateRangePicker_picker__rtl_6.DateRangePicker_picker__rtl_6 {
+  direction: rtl
+}
+.DateRangePicker_picker__rtl_7.DateRangePicker_picker__rtl_7.DateRangePicker_picker__rtl_7.DateRangePicker_picker__rtl_7.DateRangePicker_picker__rtl_7.DateRangePicker_picker__rtl_7.DateRangePicker_picker__rtl_7 {
+  direction: rtl
+}
+.DateRangePicker_picker__rtl_8.DateRangePicker_picker__rtl_8.DateRangePicker_picker__rtl_8.DateRangePicker_picker__rtl_8.DateRangePicker_picker__rtl_8.DateRangePicker_picker__rtl_8.DateRangePicker_picker__rtl_8.DateRangePicker_picker__rtl_8 {
```
ljharb: also it seems like all of these should be collapsed - ie, `a { blah } b { blah }` should always be `a, b { blah }`

src/components/CalendarDay.jsx
```diff
@@ -174,6 +174,7 @@ class CalendarDay extends React.Component {
 CalendarDay.propTypes = propTypes;
 CalendarDay.defaultProps = defaultProps;
 
+export { CalendarDay as PureCalendarDay };
```
gabergg: Could Pure be a misleading term here? Maybe overloaded

initialize.js
```diff
@@ -0,0 +1,4 @@
+// eslint-disable-next-line import/no-unresolved
```
ljharb: this override doesn't need to be here if the build process has ran, does it?

package.json
```diff
@@ -4,9 +4,9 @@
   "description": "A responsive and accessible date range picker component built with React",
   "main": "index.js",
   "scripts": {
-    "build": "npm run clean && npm run build:js && npm run build:css && npm run build:svg",
+    "build": "npm run clean && npm run build:js && npm run build:svg && mkdir lib/css && npm run build:css -- --optimize ",
```
ljharb: should `--optimize` just go in build:css? or is it useful to have it separate?

also, the `mkdir -p` should go in `prebuild:css` instead, and we should also have a script that rimrafs lib/css prior to building, just in case

src/components/CalendarMonthGrid.jsx
```diff
@@ -91,25 +98,32 @@ function getMonths(initialMonth, numberOfMonths, withoutTransitionMonths) {
...
-export default class CalendarMonthGrid extends React.Component {
+class CalendarMonthGrid extends React.Component {
   constructor(props) {
     super(props);
     const withoutTransitionMonths = props.orientation === VERTICAL_SCROLLABLE;
     this.state = {
       months: getMonths(props.initialMonth, props.numberOfMonths, withoutTransitionMonths),
     };
 
+    this.calendarMonthHeights = new Array(props.numberOfMonths);
```
ljharb: eesh, `new Array` creates a sparse array, which is terribad for performance. does this need to start out as more than `[]`?

src/components/CalendarMonthGrid.jsx
```diff
@@ -166,6 +189,20 @@ export default class CalendarMonthGrid extends React.Component {
     this.container = ref;
   }
 
+  setMonthHeight(height, i) {
+    if (this.calendarMonthHeights[i]) {
+      if (i === 0) {
+        this.calendarMonthHeights.pop();
+        this.calendarMonthHeights.unshift(height);
```
ljharb: the mutation, it makes me sad. is there no way to avoid or simplify all this?

src/components/DayPicker.jsx
```diff
@@ -395,24 +329,23 @@ export default class DayPicker extends React.Component {
...
+      }
 
-    // The first CalendarMonth is always positioned absolute at top: 0 or left: 0
-    // so we need to transform it to the appropriate location before the animation.
-    // This behavior is because we would otherwise need a double-render in order to
-    // adjust the container position once we had the height the first calendar
-    // (ie first draw all the calendar, then in a second render, use the first calendar's
-    // height to position the container). Variable calendar heights, amirite? <3 Maja
-    this.translateFirstDayPickerForAnimation(translationValue);
+      const newMonthHeight = this.calendarMonthHeights.slice(0, numberOfMonths)
```
ljharb: `const newMonthHeight = Math.max(0, ...this.calendarMonthHeights.slice(0, numberOfMonths));` (the 0 is only necessary because otherwise the max would be `-Infinity` if the array is empty)

src/components/DayPicker.jsx
```diff
@@ -424,14 +357,20 @@ export default class DayPicker extends React.Component {
...
-      this.isVertical() ? -this.getMonthHeightByIndex(1) : -this.dayPickerWidth;
+      this.isVertical() ? -this.calendarMonthHeights[1] : -calendarMonthWidth;
 
-    if (isRTL && this.isHorizontal()) {
-      translationValue = 0;
+    if (this.isHorizontal()) {
+      if (isRTL) {
+        translationValue = 0;
+      }
+      const newMonthHeight = this.calendarMonthHeights.slice(2)
```
ljharb: `const newMonthHeight = Math.max(0, ...this.calendarMonthHeights.slice(2));`

src/components/DayPicker.jsx
```diff
@@ -442,6 +381,23 @@ export default class DayPicker extends React.Component {
...
+
+    let firstVisibleMonthIndex = 1;
+    if (monthTransition === PREV_TRANSITION) {
+      firstVisibleMonthIndex -= 1;
+    } else if (monthTransition === NEXT_TRANSITION) {
+      firstVisibleMonthIndex += 1;
+    }
+
+    const verticalScrollable = orientation === VERTICAL_SCROLLABLE;
+    if (verticalScrollable) firstVisibleMonthIndex = 0;
```
ljharb: `if (orientation === VERTICAL_SCROLLABLE) return 0;`?

src/components/DayPicker.jsx
```diff
@@ -457,12 +413,17 @@ export default class DayPicker extends React.Component {
...
-    this.calendarMonthGrid = ref;
+  setCalendarMonthHeights(calendarMonthHeights) {
+    const { numberOfMonths } = this.props;
+    const firstVisibleMonthIndex = this.getFirstVisibleIndex();
+
+    this.calendarMonthHeights = calendarMonthHeights;
+    this.calendarMonthGridHeight = calendarMonthHeights.reduce((maxHeight, height, i) => {
+      const isVisible =
+        (i >= firstVisibleMonthIndex) && (i < firstVisibleMonthIndex + numberOfMonths);
+      return isVisible ? Math.max(maxHeight, height) : maxHeight;
```
ljharb: `Math.max(...calendarMonthHeights.filter((, i) => (i >= firstVisibleMonthIndex) && (i < firstVisibleMonthIndex + numberOfMonths))`?

src/components/DayPickerKeyboardShortcuts.jsx
```diff
@@ -128,42 +105,61 @@ export default class DayPickerKeyboardShortcuts extends React.Component {
       ? phrases.hideKeyboardShortcutsPanel
       : phrases.showKeyboardShortcutsPanel;
 
+    const bottomRight = (buttonLocation === BOTTOM_RIGHT);
```
ljharb: all these parens are redundant

src/utils/calculateDimension.js
```diff
@@ -0,0 +1,32 @@
...
+  const style = (!borderBox || withMargin) ? window.getComputedStyle(el) : null;
+
+  // Offset includes border and padding
+  const { offsetWidth, offsetHeight } = el;
+  let size = axis === 'width' ? offsetWidth : offsetHeight;
+
+  // Get the inner size
+  if (!borderBox) {
+    size -= (
+      parseFloat(style[`padding${axisStart}`]) +
```
ljharb: does this need to pivot based on RTL?

stories/DateRangePicker.js
```diff
@@ -83,7 +83,6 @@ storiesOf('DateRangePicker (DRP)', module)
     moment.locale('fa');
     return (
       <DateRangePickerWrapper
-        placeholder="تقویم فارسی"
```
ljharb: isn't the goal here to have farsi/arabic in the example? why remove the placeholder?

src/shapes/withStylesPropTypes.js
ljharb: yeah that's actually a good question; can't we import from there instead?

src/components/CalendarDay.jsx
ljharb: wait, why are we exporting this at all? best practice is to only use `.dive()` to get at the unwrapped component

stories/DateRangePicker.js
majapw: This didn't actually do anything. `placeholder` is not a valid prop on the DRP and the text itself translates to Farsi Calendar which is not a meaningful value for the `startDatePlaceholderText` or the `endDatePlaceholderText` props.

src/shapes/withStylesPropTypes.js
majapw: `react-with-styles` doesn't actually have this file. :P We fake it in our own internal repo. Probably the right thing to do is to add it to `react-with-styles`

src/components/CalendarDay.jsx
majapw: because we are spying on prototype methods. `dive` doesn't solve that issue afaik, does it?

src/components/CalendarMonthGrid.jsx
majapw: I can make a new array instead.

src/components/CalendarMonthGrid.jsx
majapw: I was doing this because I didn't want to assume the order that the `CalendarMonth` heights would come back in. As such, I'm creating the array with the known length and then setting `this.calendarMonthHeights[i] = ...` as the heights come back. Is there a better way to do this? I suppose I could use an object instead but that seems... silly

package.json
majapw: `--optimize` is quite slow (like pretty damn slow), so I wanted a fast version of `build:css` that would run as part of the `storybook:css` script.

initialize.js
majapw: it's true, but... it hasn't at the time of the lint call.

css/styles.css
```diff
@@ -0,0 +1,18876 @@
+.DateRangePicker {
+  position: relative;
+  display: inline-block
```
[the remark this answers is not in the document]
majapw: This file is autogenerated and that process still needs some improvements --- namely on collapsing style definitions as well as best practices for styling. X_x 

I think I have also committed the unoptimized version here which I ... probably don't want?

I think I should just `.gitignore` this file tbh.

css/styles.css
majapw: again this is due partly to the autogeneration of this file and i need to... probably clean it up a bit more/read more into clean-css options.

css/styles.css
majapw: again this is due partly to the autogeneration of this file and i need to... probably clean it up a bit more/read more into clean-css options.

stories/DateRangePicker.js
ljharb: 👍

src/components/CalendarMonthGrid.jsx
ljharb: you don't have to set the initial length, JS will grow it for you as you set `i` - but also you could do `Array.from({ length: props.numberOfMonths })` if you want a dense array of `undefined` values.

css/styles.css
ljharb: the output of the css file is totes a nonblocker, but it'd be nice to file issues somewhere to document these kinds of things

css/styles.css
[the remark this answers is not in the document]
ljharb: yeah that's a good idea, good call

src/shapes/withStylesPropTypes.js
ljharb: aha, yes let's please do that asap :-D

src/components/CalendarDay.jsx
ljharb: you have me there.

any way we could avoid the spying?

it'd be ugly but we could `const Pure = shallow(<Wrapped />).dive().type(); sinon.spy(Pure.prototype, 'foo')`

src/components/CalendarDay.jsx
majapw: I think we can do this in a follow-up PR (and one that is slightly less overwhelming to read)

src/components/CalendarMonthGrid.jsx
majapw: oh, i did not know that you could do 
```
const foo = [];
foo[3] = 'bar';
```
and that would work. 

I think it just goes against all my computer science degree sensibilities. :P

lencioni: If the values of this array are all positive ints, you might want to use a typed array.

majapw: a what now

lencioni: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Typed_arrays

But you'd probably need to have a regular array fallback for older browsers. Probably outside the scope of this PR, but possibly worth looking into for the future.

src/utils/calculateDimension.js
majapw: I do not believe so (or at the least, `react-dates` currently still works as expected with the `isRTL` prop). However, I am pegging to the old version of RWS-IA for this change and will come back to the question of RTL support with the new interface in a follow-up PR. :)

majapw:
@ljharb @gabergg I believe I've addressed all of your comments.

src/theme/DefaultTheme.js
```diff
@@ -0,0 +1,149 @@
...
+  primaryShade_4: '#b2f1ec',
+  primary_dark: '#008489',
+
+  secondary: '#007a87',
+
+  yellow: '#ffe8bc',
+  yellow_dark: '#ffce71',
+};
+
+export default {
```
majapw: One of the things I was thinking about was that... if we're importing this into a project that uses many `react-with-styles` packages and we need to register the theme, we should probably namespace the theme so that there are no collisions. Something like:
```
export default {
  reactDates: { 
    color: ...,
    zIndex: ...,
    ...
  },
};
```
@gabergg @lencioni @ljharb what do you think?

lencioni: Yep, I think it should definitely be namespaced.

.storybook-css/addons.js
```diff
@@ -0,0 +1,4 @@
+/* eslint-disable import/no-extraneous-dependencies, import/extensions */
```
ljharb: why are these deps extraneous?

scripts/buildCSS.js
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env node
+
+const fs = require('fs');
+const CleanCSS = require('clean-css');
+
+const CSSInterface = require('react-with-styles-interface-css').default;
```
ljharb: wait; this is a bug in the css interface. There should never be a `default` property on `module.exports` for a package's entry point :-/

It's possible your addition of `registerMaxSpecificity` caused this; if so, maybe we should remove it and have it be under a different entry point (`require('react-with-styles-interface-css/registerMaxSpecificity')`)

scripts/buildCSS.js
```diff
@@ -0,0 +1,35 @@
...
+const compileCSS = require('react-with-styles-interface-css/compile').default;
+
+const registerInterfaceWithDefaultTheme = require('../src/utils/registerInterfaceWithDefaultTheme').default;
+
+const args = process.argv.slice(2);
+const optimizeForProduction = args.includes('-o') || args.includes('--optimize');
+
+require('../test/_helpers/ignoreSVGStrings');
+
+registerMaxSpecificity(11);
```
ljharb: why is this turned up to eleven?

(no but seriously, why the magic number?)

scripts/buildCSS.js
```diff
@@ -0,0 +1,35 @@
...
+registerInterfaceWithDefaultTheme(CSSInterface);
+
+const DateRangePickerPath = './src/components/DateRangePicker.jsx';
```
ljharb: don't you want to pull in the babel output, not the raw code?

scripts/buildCSS.js
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/env node
+
+const fs = require('fs');
+const CleanCSS = require('clean-css');
+
+const CSSInterface = require('react-with-styles-interface-css').default;
+const registerMaxSpecificity = require('react-with-styles-interface-css').registerMaxSpecificity;
+const compileCSS = require('react-with-styles-interface-css/compile').default;
```
ljharb: also here

src/components/CalendarMonthGrid.jsx
```diff
@@ -166,6 +187,18 @@ export default class CalendarMonthGrid extends React.Component {
     this.container = ref;
   }
 
+  setMonthHeight(height, i) {
+    if (this.calendarMonthHeights[i]) {
+      if (i === 0) {
+        this.calendarMonthHeights = [height, ...this.calendarMonthHeights.slice(0, -1)];
+      } else if (i === this.calendarMonthHeights.length - 1) {
+        this.calendarMonthHeights = [...this.calendarMonthHeights.slice(1), height];
```
ljharb: it's slightly more efficient here to do `this.calendarMonthHeights.slice(1).concat(height)`

src/components/CalendarMonthGrid.jsx
```diff
@@ -166,6 +187,18 @@ export default class CalendarMonthGrid extends React.Component {
     this.container = ref;
   }
 
+  setMonthHeight(height, i) {
+    if (this.calendarMonthHeights[i]) {
+      if (i === 0) {
+        this.calendarMonthHeights = [height, ...this.calendarMonthHeights.slice(0, -1)];
```
ljharb: it's slightly more efficient here to do `[height].concat(this.calendarMonthHeights.slice(0, -1))`

src/components/DayPicker.jsx
```diff
@@ -898,3 +813,110 @@ export default class DayPicker extends React.Component {
 
 DayPicker.propTypes = propTypes;
 DayPicker.defaultProps = defaultProps;
+
+export { DayPicker as PureDayPicker };
```
ljharb: i think i called this out before; this is just to spy on instance methods in tests, yes?

src/components/SingleDatePicker.jsx
```diff
@@ -479,3 +472,82 @@ export default class SingleDatePicker extends React.Component {
 
 SingleDatePicker.propTypes = propTypes;
 SingleDatePicker.defaultProps = defaultProps;
+
+export { SingleDatePicker as PureSingleDatePicker };
```

src/theme/DefaultTheme.js
ljharb: I'm confused, why should it be namespaced?

This module exports a theme directly; wouldn't whoever registered the theme be able to install it under a key themselves?

test/.eslintrc
```diff
@@ -10,8 +10,11 @@
   },
   "extends": "airbnb",
   "rules": {
+    "import/no-extraneous-dependencies": [2, {
+      "devDependencies": true
+    }],
     "indent": [2, 2, {
-		"MemberExpression": "off"
-	}],
+  		"MemberExpression": "off"
```
ljharb: these two lines have tab chars; please make them spaces

src/theme/DefaultTheme.js
majapw: But the components themselves would be expecting the `color`, `zIndex`, etc. keys directly and there'd be no way to change that.

.storybook-css/addons.js
majapw: They're never used directly!

majapw: (but they are necessary)

scripts/buildCSS.js
majapw: https://github.com/airbnb/react-with-styles-interface-css/pull/8 should fix some things!

scripts/buildCSS.js
majapw: Same https://github.com/airbnb/react-with-styles-interface-css/pull/8

scripts/buildCSS.js
majapw: I should leave a comment. The number comes from the maximum number of styles passed into any `css` call in this repo

scripts/buildCSS.js
majapw: I don't... think so. At least, that's not what the css interface uses as its tests.

src/components/DayPicker.jsx
majapw: Yes

src/components/SingleDatePicker.jsx
majapw: Also yes

src/theme/DefaultTheme.js
ljharb: gotcha. maybe it'd be better to have `withStyles` support a "namespace" option or something, but this is fine for now i guess.

.storybook-css/addons.js
ljharb: They're imported here; "extraneous" means they're not in package.json. If they're necessary, why can't they be in package.json?

majapw: So the thing is... they come with `@storybook/react` but they have namespacing that makes eslint think they're separate packages. idk man, i blame storybook for this one, but this is what you have to do to get them to work.

majapw:
@ljharb @gabergg @lencioni I am fishing for a stamp! :) I think I have addressed all comments.

.storybook-css/addons.js
ljharb: aren't they separate packages? https://www.npmjs.com/package/@storybook/addon-actions and https://www.npmjs.com/package/@storybook/addon-links

majapw: I don't believe they were when I first put together this addons.js file/updated storybook in this package. It's possible this has changed in newer versions.

This is literally just a copy of what's in `.storybook` which is the default config.

scripts/buildCSS.js
```diff
@@ -0,0 +1,36 @@
+#!/usr/bin/env node
+
+const fs = require('fs');
+const CleanCSS = require('clean-css');
+
+const compileCSS = require('react-with-styles-interface-css/compile');
+
+const registerMaxSpecificity = require('react-with-styles-interface-css/dist/utils/registerMaxSpecificity').default;
```
ljharb: oh noes, there's still dot defaults? :-(

adamchenwei:
what is the progress here??? are we pursuing styled-component now? pretty confused, I tried to get this react-date running with` with-react-style` but the component seems just dont have any style, even that's pretty much how the storybook from airbnb setup for DateRangePicker is... what else I should be configuring to have the style?

Btw the symptom is also that, I see dynamic class name applied to the airbnb's own storybook, but by using same config I copied from the airbnb's DateRangePicker story, I observed all css is missing.... anything I should be added to the style without apply a whole ccss sheet? I thought withStyle should work as is no?

majapw:
@adamchenwei this PR was merged in in October and has been in active use since then... Have you read through the README, specifically this section: https://github.com/airbnb/react-dates#initialize?

There's also a bit more context on the transition to `react-with-styles` in https://github.com/airbnb/react-dates#advanced. 

I'd love to answer more specific questions if you have them in a new issue. :)

adamchenwei:
@majapw thanks for the reply! I did do above. so the only thing I forgot and I found is to loading the actual css. I thought it suppose to work out of the book with `withStyle` but I see ticket that is work in progress toward that, which is great. Thanks.

adamchenwei:
@majapw actually sorry... I read the sections now, its not what I thought it was lol, I must misread... but it says react-date no longer depends on css, but I still had to even I used withStyle.... I think something else must went wrong.... I need to figure this out. thanks. Never actually knew react-dates no longer depends on css anymore lol wonder why its still avaialble for import, and no ... warning for deprecating css import? that would be great.... just a thought

majapw:
@adamchenwei can you please open an issue so we can continue this discussion there where it's a bit more discoverable? :)

PR #739, opened 2017-10-02
Added Fn call when month change

phsantiago:
When I was implementing this component, I missed this function to reflect month transitions in others parts of my state.

### DIFF
--- a/src/components/DayPicker.jsx
+++ b/src/components/DayPicker.jsx
@@ -62,7 +62,7 @@ const propTypes = forbidExtraProps({
   onNextMonthClick: PropTypes.func,
   onMultiplyScrollableMonths: PropTypes.func, // VERTICAL_SCROLLABLE daypickers only
   onMonthChange: PropTypes.func,
-  
+
   // month props
   renderMonth: PropTypes.func,
 
@@ -583,9 +583,9 @@ export default class DayPicker extends React.Component {
         );
       }
     }
-    
+
     onMonthChange(newMonth);
-    
+
     this.setState({
       currentMonth: newMonth,
       monthTransition: null,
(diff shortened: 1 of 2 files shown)


majapw:
I'm confused on why we would need the `onMonthChange` callback when we already have the `onPrevMonthClick` and `onNextMonthClick` methods?

phsantiago:
@majapw with `onPrevMonthClick` and `onNextMonthClick` we need to calculate outside the Calendar component how many months has passed, with sums and subtractions. It's easier has this props telling us in what month we are.

phsantiago:
@ljharb Thank you for the feedback. I've made some tweaks, can you review please?

majapw:
@phsantiago If you look at https://github.com/airbnb/react-dates/blob/master/src/components/DayPickerRangeController.jsx#L521 and https://github.com/airbnb/react-dates/blob/master/src/components/DayPickerRangeController.jsx#L545, we actually do pass the current month back to the props pass through the `DateRangePicker`. Something similar happens for the `SingleDatePicker`.

Are you using the `DayPicker` component directly?

phsantiago:
@majapw Wow sorry, you are sure! I'm using `DayPicker` directly, didn't knew about [those](https://github.com/airbnb/react-dates/blob/master/src/components/DayPickerRangeController.jsx#L521) [lines](https://github.com/airbnb/react-dates/blob/master/src/components/DayPickerRangeController.jsx#L545).

majapw:
@phsantiago unless you are doing something dramatically different from picking a range or picking a single date, I would recommend using one of those directly. :) In any case, I don't think that this is necessarily the right strategy to add this prop.

PR #748, opened 2017-10-06
Move calendar month caption div outside of table

majapw:
In a bad rebase, I left this div inside the table which caused... problems.

to: @ljharb


majapw:
😆

PR #753, opened 2017-10-06
Changes css style specificity to 0

majapw:
While having specificity be quite high is a necessity in having `react-with-styles-interface-css` be fully operable, it is not actually necessary (with one small tweak) for `react-dates` itself. In order to have our exported css file be less of a trash-fire, I am reducing the specificity to zero.

to: @ljharb @lencioni

### DIFF
--- a/src/components/CalendarDay.jsx
+++ b/src/components/CalendarDay.jsx
@@ -126,9 +126,9 @@ class CalendarDay extends React.Component {
       modifiers.has('selected-start') ||
       modifiers.has('selected-end');
 
-    const hoveredSpan =
+    const hoveredSpan = !selected && (
       modifiers.has('hovered-span') ||
-      modifiers.has('after-hovered-start');
+      modifiers.has('after-hovered-start'));
 
     const isOutsideRange = modifiers.has('blocked-out-of-range');
 
--- a/scripts/buildCSS.js
+++ b/scripts/buildCSS.js
@@ -13,9 +13,7 @@ const optimizeForProduction = args.includes('-o') || args.includes('--optimize')
 
 require('../test/_helpers/ignoreSVGStrings');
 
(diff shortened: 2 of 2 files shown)


src/components/CalendarDay.jsx
```diff
@@ -126,9 +126,9 @@ class CalendarDay extends React.Component {
       modifiers.has('selected-start') ||
       modifiers.has('selected-end');
 
-    const hoveredSpan =
+    const hoveredSpan = !selected && (
```
lencioni: Is this related?

majapw: Yes! So basically specificity is what generates something like:
```
.CalendarDay, 
.CalendarDay_1.CalendarDay_1,
.CalendarDay_2.CalendarDay_2.CalendarDay_2,
```
and then the appropriate class name is applied based on the position `styles.CalendarDay` takes in the `css` call. 

In this particular method, we were relying on a particular order of `hoveredSpan` vs. `selected` to do styling overrides. By not applying the `hoveredSpan` class to `selected` days, we get around this necessity. As a result, this allows us to remove the specificity requirement and make the css file we ship with `react-dates` much smaller.

In general, we'd recommend using a high specificity value with RWS-ICSS (it defaults to 20), but in this small package, it's not strictly necessary.

ljharb: So then this change means that selected days no longer have a hover effect?

majapw: There is still a hover effect on selected days as defined by https://github.com/airbnb/react-dates/blob/4a8d379a2d23c1e0c265545b8ac7f2e78ab3a607/src/components/CalendarDay.jsx#L259-L263. The hovered span styling are completely different and before we were relying on the ordering of styles to apply selected over hovered. However, this change means that those styles will never be applied at the same time which will solve this issue.

PR #759, opened 2017-10-09
Add eslint-plugin-react-with-styles, add root: true to .eslintrc

lencioni:
These eslint rules will help protect this project from bad
react-with-styles usage.

@majapw

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,5 +1,12 @@
 {
-  "extends": "airbnb",
+  "extends": [
+    "airbnb",
+    "plugin:react-with-styles/recommended"
+  ],
+
+  "plugins": [
+    "react-with-styles"
+  ],
 
   "env": {
     "browser": true,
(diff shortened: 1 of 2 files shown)


.eslintrc
```diff
@@ -8,8 +11,13 @@
 
   "rules": {
     "no-mixed-operators": [2, { "allowSamePrecedence": true }],
+
     "jsx-a11y/no-static-element-interactions": 1, // TODO: enable
+
     "react/no-did-mount-set-state": 0, // necessary for server-rendering
+
+    "react-with-styles/no-unused-styles": 2,
```
ljharb: do these come in with a recommended config at all, or only enabled one at a time?

lencioni: One of them is in recommended config. The other is not yet.

ljharb: Should we use the recommended config then, and only add the rule that differs?

lencioni: Sure thing, I'll push an update shortly.

PR #767, opened 2017-10-12
Add missing customCloseIcon prop/usage to SingleDatePicker.

mmaday:
Properly pass the customCloseIcon property from the SingleDatePicker to the SingleDatePickerInput.


majapw:
LGTM as well!

PR #771, opened 2017-10-13
For for Issue #713 Unavailable Date phrase not reading on Screen Readers

erin-doyle:
Fix for Issue #713 by updating the way in which modifiers is checked for existence of the BLOCKED_MODIFIER now that modifiers is a Set and no longer an object.

### DIFF
--- a/src/components/CalendarDay.jsx
+++ b/src/components/CalendarDay.jsx
@@ -101,15 +101,11 @@ class CalendarDay extends React.Component {
 
     if (!day) return <td />;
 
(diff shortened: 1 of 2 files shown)


erin-doyle:
No problem, I'm happy I was able to find it!  This was one of those sneaky bugs where for the case of available dates it was coincidentally working fine, where the check for the blocking modifier was returning false, it just happened to be for the wrong reason.  So it's easy to see how it was able to fly under the radar. 😉

Issue #632, opened 2017-07-21
Implement verticalHeightFactor prop


vemv:
From DatePicker.jsx source:

```javascript
    // this is a kind of made-up value that generally looks good. we'll
    // probably want to let the user set this explicitly.
    const verticalHeight = 1.75 * calendarMonthWidth;
```

I guess that the user-configurable value should be the 1.75 - setting `verticalHeight` directly seems more error-bound.

Could you please implement this prop?

Cheers - Victor

ljharb:
Seems reasonable.

vemv:
This one was fixed by https://github.com/airbnb/react-dates/pull/773

PR #773, opened 2017-10-15
add verticalHeight prop for DayPicker

ingiulio:
Addresses #632 

👋

These changes allow the user of DayPicker to specify a vertical height (applied only when on vertical orientation), while keeping the current behavior (an arbitrary size) as the default if no height is specified.

This is actually slightly different than what the issue's title suggests, but imho it seems easier for the user to just have full control on the height rather than trying different "factors". Open to different opinions obviously!

### DIFF
--- a/src/components/DayPicker.jsx
+++ b/src/components/DayPicker.jsx
@@ -55,6 +55,7 @@ const propTypes = forbidExtraProps({
   hideKeyboardShortcutsPanel: PropTypes.bool,
   daySize: nonNegativeInteger,
   isRTL: PropTypes.bool,
+  verticalHeight: nonNegativeInteger,
 
   // navigation props
   navPrev: PropTypes.node,
@@ -99,6 +100,7 @@ export const defaultProps = {
   hideKeyboardShortcutsPanel: false,
   daySize: DAY_SIZE,
   isRTL: false,
+  verticalHeight: null,
 
   // navigation props
   navPrev: null,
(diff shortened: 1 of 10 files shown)


src/components/DayPicker.jsx
```diff
@@ -55,6 +55,7 @@ const propTypes = forbidExtraProps({
   hideKeyboardShortcutsPanel: PropTypes.bool,
   daySize: nonNegativeInteger,
   isRTL: PropTypes.bool,
+  verticalHeight: PropTypes.number,
```
ljharb: should this be a nonNegativeInteger?

ingiulio: sounds good, I can't think of a case where it would in fact make sense to use a negative height 🤔

ingiulio:
@majapw ok! I'll add it there too.

Btw, is the `Changes unknown [etc]` message from coveralls something new?

ljharb:
@ingiulio try rebasing on the latest master, on the command line?

majapw:
This PR will be ready to go once you can set the `verticalHeight` prop in the `SingleDatePicker`, `DateRangePicker`, `DayPickerSingleDateController`, and `DayPickerRangeController`.

I also reran the pending test.

ingiulio:
Done, thank you! 
Do you think the stories for `SingleDatePicker ` and `DateRangePicker ` are useful?  I guess they help verify the props are passed through properly, but if they don't add much value please let me know and I'll just remove them.

ingiulio:
Thank you! :)

Issue #774, opened 2017-10-16
Right to left keyboard navigation 


sag1v:
Hi, when i created [454](https://github.com/airbnb/react-dates/pull/454) for the `RTL` support ,i forgot to take care for the keyboard navigation. Now it is reversed.  
It's been a while since I've looked in the source code and can't find the time to do another PR on this in the near future.  
Would love to get some help on that.

PR #775, opened 2017-10-16
Fixed RTL navigations

sag1v:
This is due to #774

### DIFF
--- a/src/components/DayPicker.jsx
+++ b/src/components/DayPicker.jsx
@@ -248,7 +248,7 @@ class DayPicker extends React.Component {
 
     this.setState({ withMouseInteractions: false });
 
-    const { onBlur } = this.props;
+    const { onBlur, isRTL } = this.props;
     const { focusedDate, showKeyboardShortcuts } = this.state;
     if (!focusedDate) return;
 
@@ -271,7 +271,11 @@ class DayPicker extends React.Component {
         break;
       case 'ArrowLeft':
         e.preventDefault();
-        newFocusedDate.subtract(1, 'day');
+        if (isRTL) {
+          newFocusedDate.add(1, 'day');
+        } else {
+          newFocusedDate.subtract(1, 'day');
+        }
         didTransitionMonth = this.maybeTransitionPrevMonth(newFocusedDate);
         break;
       case 'Home':
@@ -292,7 +296,11 @@ class DayPicker extends React.Component {
         break;
       case 'ArrowRight':
         e.preventDefault();
-        newFocusedDate.add(1, 'day');
+        if (isRTL) {
+          newFocusedDate.subtract(1, 'day');
+        } else {
+          newFocusedDate.add(1, 'day');
+        }
         didTransitionMonth = this.maybeTransitionNextMonth(newFocusedDate);
         break;
       case 'End':


majapw:
Hi @sag1v, I think we'll have to update the phrases that get read out by the screen reader as well. Do the keyboard shortcuts still make sense with this update?

sag1v:
Hi @majapw, what do you mean by keyboard shortcuts? Iv'e only updated the logic inside the switch statement of the key press for right and left arrows, if we on RTL then the next focused day is calculated in reversed

sag1v:
@majapw Any further thoughts about this PR?

src/components/DayPicker.jsx
```diff
@@ -265,7 +265,7 @@ class DayPicker extends React.Component {
         break;
       case 'ArrowLeft':
         e.preventDefault();
-        newFocusedDate.subtract(1, 'day');
+        !isRTL ? newFocusedDate.subtract(1, 'day') : newFocusedDate.add(1, 'day');
```
majapw: I think our linter rules require this to be:
```
if (isRTL) {
  newFocusedDate.add(1, 'day');
} else {
  newFocusedDate.subtract(1, 'day');
}
```

src/components/DayPicker.jsx
```diff
@@ -286,7 +286,7 @@ class DayPicker extends React.Component {
         break;
       case 'ArrowRight':
         e.preventDefault();
-        newFocusedDate.add(1, 'day');
+        !isRTL ? newFocusedDate.add(1, 'day') : newFocusedDate.subtract(1, 'day');
```
majapw:
```
if (isRTL) {
  newFocusedDate.subtract(1, 'day');
} else {
  newFocusedDate.add(1, 'day');
}
```

ljharb:
Also, please rebase on the command line so as to remove any merge commits :-)

sag1v:
@majapw 

> Can you also update the Home and End shortcuts?  
What kind of update? `Home` & `End`  are not affected by `RTL` as far as i know

ljharb:
Ah, you’d have to clone your *fork* and rebase your branch there. No worries tho, we can do it for you prior to merging.

majapw:
@sag1v While on keyboards that have a home and end button, i agree that it's not as much of an issue, but on keyboards where the functionality is forced by `fn + left` or `fn + right`, it feels like it might make sense to switch them? 

It might be a minority of keyboards for those accessing the RTL version of the datepicker, so maybe it doesn't make sense and home pointing to beginning of the week/end pointing to the end of the week is fine.

sag1v:
@majapw Thanks, i see your point. If there is anyway to determine if a Home or End buttons aren't available then we can write the condition for that. Otherwise i think we should merge this PR to at least support the Arrow keys on `RTL` until  we can find a way to support fn + Arrows.

ljharb:
I believe on a Mac, `fn` + left/right is indistinguishable from "home"/"end"?

sag1v:
@ljharb 

> I believe on a Mac, fn + left/right is indistinguishable from "home"/"end"?


It seems [it's not that different than other brands](https://www.macworld.co.uk/how-to/mac/find-home-end-buttons-on-mac-3618360/):

PR #777, opened 2017-10-16
Bump react-with-styles-interface-css version

majapw:
Previously `react-with-styles-interface-css` included both the interface and the compiler. Because the compiler was doing some build step shenanigans to do compilation on the fly, it included a regular dependency on `babel-register` to do some babel compilation on the fly. This, of course, got passed down to consumers of `react-dates` which wasn't great. `react-with-styles-interface-css` split into ``react-with-styles-interface-css` and `react-with-styles-interface-css-compiler` in v3, and this PR addresses that change.

to: @ljharb

### DIFF
--- a/scripts/.eslintrc
+++ b/scripts/.eslintrc
@@ -1,0 +1,7 @@
+{
+  "rules": {
+    "import/no-extraneous-dependencies": [2, {
+      "devDependencies": true
+    }],
+  }
+}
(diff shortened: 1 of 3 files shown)


scripts/.eslintrc
```diff
@@ -0,0 +1,7 @@
+{
+  "rules": {
+    "import/no-extraneous-dependencies": [2, {
+      "devDependencies": true
```
ljharb: this can even be configured in the root .eslintrc, with glob-based configuration :-D

PR #780, opened 2017-10-19
Fix some apparent a11y issues with react-dates

majapw:
I recently pulled in the latest version of `react-dates` into our internal repos and discovered a few accessibility issues through our automatic testing.

1. The caption id was not unique! Namely it was the same on each month which was causing some problems. It also was missing the corresponding `aria-labelledby` attribute the table itself.
2. There was a text-contrast issue with the focus background color, so I modified the default styling to match what is currently used on airbnb.com. I still need to go back and make sure everything is customizable. :)
3. I took the opportunity to address a lot of queries in the issues on this repo and make the input a real input. 

to: @ljharb @gabergg @backwardok @airbnb/webinfra

### DIFF
--- a/src/components/CalendarMonth.jsx
+++ b/src/components/CalendarMonth.jsx
@@ -172,7 +172,6 @@ class CalendarMonth extends React.Component {
         data-visible={isVisible}
       >
         <div
-          id="CalendarMonth__caption"
           ref={this.setCaptionRef}
           {...css(
             styles.CalendarMonth_caption,
@@ -182,7 +181,10 @@ class CalendarMonth extends React.Component {
           <strong>{monthTitle}</strong>
         </div>
 
-        <table {...css(styles.CalendarMonth_table)} role="presentation">
+        <table
+          {...css(styles.CalendarMonth_table)}
+          role="presentation"
+        >
           <tbody ref={this.setGridRef}>
             {weeks.map((week, i) => (
               <tr key={i}>
--- a/src/components/DateInput.jsx
+++ b/src/components/DateInput.jsx
@@ -13,7 +13,6 @@ const propTypes = forbidExtraProps({
   id: PropTypes.string.isRequired,
   placeholder: PropTypes.string, // also used as label
   displayValue: PropTypes.string,
-  inputValue: PropTypes.string,
   screenReaderMessage: PropTypes.string,
   focused: PropTypes.bool,
   disabled: PropTypes.bool,
@@ -37,7 +36,6 @@ const propTypes = forbidExtraProps({
 const defaultProps = {
   placeholder: 'Select Date',
   displayValue: '',
-  inputValue: '',
   screenReaderMessage: '',
   focused: false,
   disabled: false,
@@ -90,7 +88,6 @@ class DateInput extends React.Component {
 
     if (focused && isFocused) {
       this.inputRef.focus();
-      this.inputRef.select();
     } else {
       this.inputRef.blur();
     }
@@ -149,7 +146,6 @@ class DateInput extends React.Component {
       id,
       placeholder,
       displayValue,
-      inputValue,
       screenReaderMessage,
       focused,
       showCaret,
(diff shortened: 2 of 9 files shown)


src/components/CalendarMonth.jsx
```diff
@@ -182,7 +183,11 @@ class CalendarMonth extends React.Component {
           <strong>{monthTitle}</strong>
         </div>
 
-        <table {...css(styles.CalendarMonth_table)} role="presentation">
+        <table
+          {...css(styles.CalendarMonth_table)}
+          role="presentation"
+          aria-labelledby={`CalendarMonth_caption__${toISOMonthString(month)}`}
```
backwardok: since this has `role="presentation"`, the label here doesn't mean as much (and also possibly won't even get read)

gabergg: Maybe `CalendarMonth_caption__${toISOMonthString(month)}` in a var just to make sure they both get updated together?

majapw: hmmm, does this belong in the DayPicker then? hmmm. But the DayPicker contains all the months, so perhaps it needs a different label? 

Trying to follow up on https://github.com/airbnb/react-dates/pull/715/files#r139019568

backwardok: I don't know that it's necessary, and I think even in general case, I don't know that the month name is as meaningful as a label outside of this table. I would imagine that the container for the picker may benefit from a label like "Date picker" or "Select a date", but the month is less valuable information for the higher level container

majapw: I'll just kill the id then

majapw:
PTAL @backwardok @gabergg

src/components/DateRangePickerInput.jsx
```diff
@@ -43,7 +43,6 @@ const propTypes = forbidExtraProps({
   onQuestionMark: PropTypes.func,
 
   startDate: PropTypes.string,
-  startDateValue: PropTypes.string,
   endDate: PropTypes.string,
   endDateValue: PropTypes.string,
```
ljharb: did you mean to remove this prop also?

majapw: Yes I did!

majapw:
@ljharb the reason for removing the `inputValue`, `startDateValue`, and `endDateValue` props is that they were essentially a way to have the `input` element value to be in iso format even as the display format was something nicer. The effect was that a server-side submit of a form (rare, i know) would have a nice iso formatted date submitted.

Of course now that I have axed the concept of the fake input and real input (which is much better for accessibility reason), this is no longer necessary and the value of the `input` element is just the formatted date value when it exists.

majapw:
@ljharb @backwardok @gabergg could I get a stamp on this? I think I've addressed concerns.

src/components/DateInput.jsx
```diff
@@ -90,7 +88,6 @@ class DateInput extends React.Component {
 
     if (focused && isFocused) {
       this.inputRef.focus();
-      this.inputRef.select();
```
ljharb: Can you elaborate in why the autoselection was removed?

backwardok: imo this is actually better for a11y. It's not expected that by default, moving to a text input will select everything in it.

src/components/DateInput.jsx
```diff
@@ -181,6 +176,8 @@ class DateInput extends React.Component {
           {...css(
             styles.DateInput_input,
             readOnly && styles.DateInput_input__readOnly,
+            focused && styles.DateInput_input__focused,
+            disabled && styles.DateInput_input__disabled,
           )}
           aria-label={placeholder}
```
backwardok: probably for another PR, but this shouldn't be set to `placeholder`. If a developer wanted to pass in whatever they set for the `placeholder` as both the label and the placeholder text (which I'd argue that they shouldn't), that should be an explicit decision and not a default one.

src/components/DateInput.jsx
majapw: So before, when you focused on the input, it was select the whole text so that when you typed anything, it would just replace whatever was there. This seemed like a good approach when there was no cursor to indicate where you were in the input. However, with a cursor, it seems unnecessary behavior.

src/components/DateInput.jsx
majapw: Yeah, let's do that in a follow-up! :)

majapw:
I blame jordan! :P

PR #781, opened 2017-10-20
Fix documentation theming spread

richardvenneman:
This commit fixes the theming override example. The spread wasn't correct as the output in the current example becomes:

screen shot 2017-10-20 at 13 56 45

With this change, the object merges correctly:

screen shot 2017-10-20 at 13 56 14

NOTE: my editor automatically stripped some leading spaces as well.

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ Right now, the easiest way to tweak `react-dates` to your heart's contents is to
 .CalendarDay__highlighted_calendar {
   background: #82E0AA;
   color: #186A3B;
-} 
+}
 
 .CalendarDay__highlighted_calendar:hover {
   background: #58D68D;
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -315,25 +315,27 @@ import DefaultTheme from 'react-dates/lib/theme/DefaultTheme';
...
-      backgroundColor: '#82E0AA',
-      backgroundColor_active: '#58D68D',
-      backgroundColor_hover: '#58D68D',
-      color: '#186A3B',
-      color_active: '#186A3B',
-      color_hover: '#186A3B',
+  reactDates: {
+    ...DefaultTheme.reactDates,
+    color: {
+      ...DefaultTheme.color,
```
majapw: Does this not need to be `DefaultTheme.reactDates.color` as well? 

Also thanks for the catch! :) 

@lencioni does theme registration work by overriding everything or by merging? I guess it wouldn't really matter since this theme is namespaced anyways....

richardvenneman: You're totally right! I just pushed the change. I have that in our app's code as well so not sure why I didn't commit it like that 🤔

lencioni: @majapw It simply overrides everything See here: https://github.com/airbnb/react-with-styles/blob/e17dea580ed45b5d8a4d7974354e94e8b6706671/src/ThemedStyleSheet.js#L8-L13

PR #787, opened 2017-10-23
Replace svg-loader with babel-plugin-inline-react-svg

lencioni:
I want to do some optimizations on how react-dates is built, to help
reduce the bundle size impact of using this project. The first step I
see is reducing the size of each of the components, which we can achieve
by avoiding building them with webpack.

From what I can tell, webpack was only brought in to support getting
SVGs in these components. We can pretty easily replace this with a Babel
transform that will achieve the same result, but with less complexity in
our build, and a little less overhead in the resulting components.

Before this change:

```
~/react-dates ❯❯❯ du -hs lib/components
420K    lib/components
```

After this change:

```
~/react-dates ❯❯❯ du -hs lib/components
348K    lib/components
```

Closes #765

@airbnb/webinfra


lencioni:
Yep, I tested all of those things and everything seemed to be just great.

lencioni:
@majapw I had to rebase which removed your review. Can you give this another look?

PR #789, opened 2017-10-23
Update consolidated-events usage to newer API

lencioni:
In v1.1.0, addEventListener returns a function that can be called to
remove the event listener.

https://github.com/lencioni/consolidated-events/blob/master/CHANGELOG.md#v110


lencioni:
Yep!

PR #796, opened 2017-10-25
fix(CalendarDay): apply styles for start and end modifiers

### LINKED ISSUE #786: Style selected end date
Hi!
I've noticed that the selected start date and the selected end date share the same css class names. We have the case that we need to set a different styling to the end date. Is it possible to set a different class name for the selected end date?

Or any other ideas how to select the selected end date with css?

We're already doing things like this:
```
/*
selected end date
*/
.CalendarDay__selected,
.CalendarDay__selected:hover,
td.CalendarDay__selected + td.CalendarDay__selected {
  ...
}

/*
selected start date
*/
.CalendarDay__selected:first-of-type,
:not(.CalendarDay__selected_span) + .CalendarDay__selected {
  ...
}
```
But it feels super hacky.

Thanks!

tagoro9:
Previous versions of the library would add the `CalendarDay--selected-start` and `CalendarDay--selected-end` class names when`selected-start` and `selected-end` modifiers were passed.

I think after the switch to `react-with-styles` these classes got lost and making it harder to modify the styles of the beginning and end of a range.

I'm not very familiar with `react-with-styles`, so I don't know if this is the proper way to get those class names set in the component.

I would also like to add unit tests for this, but I didn't see tests in `CalendarDay` regarding other modifiers and I'm not quite sure how to test this. I see that the wrapper gets the class names in the test, but they have what it seems a random modifier at the end. Also, the `react-with-styles` docs mention that `css(...)` returns an object with an opaque structure, so definitely, checking the value of the props added by this object doesn't seem a good testing option.

Fixes #786.

### DIFF
--- a/src/components/CalendarDay.jsx
+++ b/src/components/CalendarDay.jsx
@@ -142,6 +142,8 @@ class CalendarDay extends React.Component {
           hoveredSpan && styles.CalendarDay__hovered_span,
           modifiers.has('selected-span') && styles.CalendarDay__selected_span,
           modifiers.has('last-in-range') && styles.CalendarDay__last_in_range,
+          modifiers.has('selected-start') && styles.CalendarDay__selected_start,
+          modifiers.has('selected-end') && styles.CalendarDay__selected_end,
           selected && styles.CalendarDay__selected,
           isOutsideRange && styles.CalendarDay__blocked_out_of_range,
           daySizeStyles,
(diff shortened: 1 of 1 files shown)


majapw:
Hi @tagoro9, you *should* be able to test if the classes you expect get applied by running `npm run storybook:css` and inspecting the elements. The default storybook script uses aphrodite which is not quite... the class structure you are expecting.

tagoro9:
Thanks for the reply @majapw. I tested that the classes got added to the implementation in the story book. When I said I didn't know how to test it, I meant how to write a unit test that ensures that those classes get added. I should have explained that better in the description.

majapw:
Sweet. Yeah, in general, we don't want to write tests for the styling (because it would be testing the interface's internals) and the end-goal is to integrate with a tool like https://github.com/Galooshi/happo instead to check for visual regressions.

THIS IS ON OUR ... eventual roadmap. :) but i've just been swamped.

PR #799, opened 2017-10-25
Fixed onKeyDownQuestionMark for the SingleDatePicker

erin-doyle:
Being able to hit the question mark key in the input of a SingleDatePicker to display the Keyboard Shortcuts panel was not working.  Now it is all connected up and in parity with the DateRangePicker.

I also cleaned up some small naming inconsistencies on the DateRangePicker side and a couple copy/paste mistakes from DateRange to SingleDate, hope that's ok!

### DIFF
--- a/src/components/SingleDatePicker.jsx
+++ b/src/components/SingleDatePicker.jsx
@@ -108,10 +108,12 @@ class SingleDatePicker extends React.Component {
       dayPickerContainerStyles: {},
       isDayPickerFocused: false,
       isInputFocused: false,
+      showKeyboardShortcuts: false,
     };
 
     this.onDayPickerFocus = this.onDayPickerFocus.bind(this);
     this.onDayPickerBlur = this.onDayPickerBlur.bind(this);
+    this.showKeyboardShortcutsPanel = this.showKeyboardShortcutsPanel.bind(this);
 
     this.onChange = this.onChange.bind(this);
     this.onFocus = this.onFocus.bind(this);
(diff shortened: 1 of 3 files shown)


src/components/DateRangePickerInput.jsx
```diff
@@ -39,8 +39,8 @@ const propTypes = forbidExtraProps({
   onStartDateShiftTab: PropTypes.func,
   onEndDateTab: PropTypes.func,
   onClearDates: PropTypes.func,
-  onArrowDown: PropTypes.func,
-  onQuestionMark: PropTypes.func,
```
ljharb: this makes it a breaking change, unfortunately

erin-doyle: (doh) sorry I should have known that.  I can split that out, yes.

src/components/SingleDatePicker.jsx
```diff
@@ -211,20 +212,22 @@ class SingleDatePicker extends React.Component {
     });
 
     onFocusChange({ focused: false });
-    onClose({ startDate, endDate });
+    onClose({ date });
```
ljharb: hm, since `onClose` is user code, this is a breaking change as well.

erin-doyle: Bugger.  Ok I'll move it over!

ljharb:
This should be rebased; and unless @majapw objects, it'll be ready to merge!

erin-doyle:
Awesome thanks happy to help! 👍 🖖

PR #800, opened 2017-10-25
[Breaking] Make DateRangePicker key down handlers match SingleDatePicker's

erin-doyle:
Some minor clean up of function name inconsistencies between SingleDatePicker and DateRangePicker.  SingleDatePicker has `onKeyDownArrowDown` and `onKeyDownQuestionMark` and that's what DateInput has, so it seemed like DateRangePicker should match.


erin-doyle:
I checked the README and I don't see any impacts to it.

ljharb:
I think it'd be a good idea to do a release before merging in breaking changes

majapw:
v14.1.0 is out, so we can merge this in.

PR #816, opened 2017-11-01
[Fix] Pass onClose from SingleDatePicker to DayPickerSingleDateController

erin-doyle:
I just found a miss here and added it in.


erin-doyle:
Yep, I can do that.

erin-doyle:
Let me know if these were not what you were asking for.  Thanks!

Issue #818, opened 2017-11-01
README.md


eryue0220:
**Bug?**

Maybe not. But suggestion is real. I forked the repo and followed the steps as saying in the README.md:

> npm install
> npm run storybook

But when I opened the url `http://localhost:6001`, I met the error:

> Uncaught Error: Cannot resolve './lib/constants' in [path/to/local/directory]

And after checking out the `package.json`, I found that I should build first then execute `npm run storybook`. And I thought the `README.md` can update as follow:

> npm install
> npm run build && npm run storybook

Or any other ways that can let other people could easily set up in their local, and thus will reduce some misleadings and time, I think.

ljharb:
Yes, you're correct that the build process is required - it should probably be in package.json as `"prestorybook": "npm run build"`, though, rather than modifying the documentation.

majapw:
@ljharb Why does storybook require the build process now? It didn't before. Is this because of the esm changes?

majapw:
@ljharb @eryue0220 I think the right thing to do is not to have the build step before storybook but rather to update the components in https://github.com/airbnb/react-dates/tree/master/examples to rely on the src version of the constants file. 

If we have storybook relying on built files, that means changes to those files won't be updated automatically as you're developing which is a bad experience.

ljharb:
Pushed up a fix to use the constants from src.

If we can figure out how to forbid imports from outside `src` or `examples` or `node_modules` in the `examples` dir, we should.

PR #822, opened 2017-11-02
update npm script

eryue0220:
Add `npm run prestorybook` to build the project. And update the `npm run storybook`. More info see #818.


package.json
```diff
@@ -22,7 +22,8 @@
     "tests-only": "npm run mocha --silent test",
     "tests-karma": "karma start",
     "test": "npm run tests-only",
-    "storybook": "start-storybook -p 6006",
+    "prestorybook": "npm run build",
+    "storybook": "npm run prestorybook && start-storybook -p 6006",
```
ljharb: Npm runs “pre” scripts automatically for you; this line should be reverted.

ljharb:
Thanks! Now this just needs a rebase down to one commit :-)

ljharb:
Makes sense; that didn't occur to me. @eryue0220, thanks for the PR - let's go with @majapw's solution instead then.

Issue #557, opened 2017-06-12
Keyboard navigation issues


majapw:
cc: @backwardok for clarification on these issues

1. Down arrow to activate the calendar interferes with JAWS standard keyboard shortcuts
2. When opening the keyboard shortcuts panel, it does not take focus. This makes it somewhat difficult to exit when using some screen readers.

backwardok:
o.O @majapw I just noticed this because I'm terrible at checking github on my personal account.

The 1st point may not be too bad since I think the field should trigger focus/forms mode, which will accept keyboard shortcuts. 

The 2nd point is actually worse than described - because it doesn't take focus, it doesn't get read. This means screen reader users will have no idea what the shortcuts are, and it's not particularly easy to figure out how to get there since the months auto advance when trying to navigate out. Additionally, for sighted keyboard users, it's a bit confusing that you can't tab to the close button. You can close the keyboard shortcut menu by hitting the `esc` key, but then you've lost your previous spot and would need to go back to wherever you were again.

erin-doyle:
I'll take a shot at working on this. :crossed_fingers:

erin-doyle:
Ok PR submitted: https://github.com/airbnb/react-dates/pull/825.

yaroslav-perec:
Auto focus for keyboard shortcuts panel doesn't work, "react-dates": "^15.4.0"
OC: ubuntu 16.04;
Browser: Google Chrome Version 61.0.3163.100;

erin-doyle:
@yaroslav-perec can you describe this a little more?  What is not working about it?  What is it doing instead?

yaroslav-perec:
@erin-doyle  Sure, I use **SingleDatePicker**, it looks like this:

![selection_328](https://user-images.githubusercontent.com/30346881/34162371-2186c75e-e4dc-11e7-91c7-db31e02c0ac6.png)

It seems close button has focus, but when I press **Enter** key shortcuts panel popup window doesn't close, I know if i click **ESC** it will close, but if you use SingleDatePicker inside modal window (that often have the same logic for **ESC** key), you will close modal window too, I think we should have possibility to close the pop-up by pressing the **Enter** key when focus on Close button.

erin-doyle:
[...]

PR #825, opened 2017-11-04
[Fix] for Issue #557 moving focus to the keyboard shortcuts panel when opened

erin-doyle:
Modified aria attributes and focusing behavior so that the keyboard shortcuts panel receives focus, is read appropriately by screen readers, prevents focus from accidentally moving back to the other elements without use of the proper shortcuts, and then returns focus to the appropriate place when closed.

I tested this on Chrome and Safari with Mac VoiceOver and FireFox with no screen reader.  It seems that all the keyboard navigation and screen reading is working consistently across them all now.  Triggering the keyboard shortcuts panel opens, focuses and returns focus when closed appropriately now regardless of how its triggered.

_EDIT:_ also tested in Microsoft Edge along with NVDA.


src/components/DayPicker.jsx
```diff
@@ -234,13 +234,10 @@ class DayPicker extends React.Component {
   }
 
   componentDidUpdate(prevProps) {
-    const { isFocused, showKeyboardShortcuts } = this.props;
+    const { isFocused } = this.props;
     const { focusedDate } = this.state;
 
-    if (
-      (!prevProps.isFocused && isFocused && !focusedDate) ||
-      (!prevProps.showKeyboardShortcuts && showKeyboardShortcuts)
```
erin-doyle: I removed this line because it was pulling the focus back to the DayPicker after it had just been moved to the keyboard shortcuts panel.

src/components/KeyboardShortcutRow.jsx
```diff
@@ -38,7 +38,7 @@ function KeyboardShortcutRow({
         <span
           {...css(styles.KeyboardShortcutRow_key)}
           role="img"
-          aria-label={label}
+          aria-label={`${label} `}
```
erin-doyle: Adding in the space here inserts a space between the screen reader reading the label and then the {action} from the <div> below.  Previously it was reading like:

"Enter keySelect the date in focus"

Basically it was merging the last word of one to the first word of the next and making it hard to understand when read.

src/components/SingleDatePicker.jsx
```diff
@@ -404,6 +404,7 @@ class SingleDatePicker extends React.Component {
           renderCalendarInfo={renderCalendarInfo}
           isFocused={isDayPickerFocused}
           showKeyboardShortcuts={showKeyboardShortcuts}
+          onBlur={this.onDayPickerBlur}
```
erin-doyle: This was another important miss I found in functionality between the DateRangePicker but not the SingleDatePicker.  This was the missing piece to the focus returning back to the DateInput when that was where the keyboard shortcuts panel was triggered to open.

src/components/DayPickerKeyboardShortcuts.jsx
```diff
@@ -33,15 +33,91 @@ const defaultProps = {
   phrases: DayPickerKeyboardShortcutsPhrases,
 };
 
+function getKeyboardShortcuts(phrases) {
+  return [{
+    unicode: '↵',
+    label: phrases.enterKey,
+    action: phrases.selectFocusedDate,
+  },
```
ljharb: the indentation here is a little off; if it's going to start with `[{`, then each line needs to be `}, {` - otherwise the first `{` needs to be on its own line.

src/components/DayPickerKeyboardShortcuts.jsx
```diff
@@ -33,15 +33,91 @@ const defaultProps = {
...
+  ];
+}
+
 class DayPickerKeyboardShortcuts extends React.Component {
   constructor(...args) {
     super(...args);
 
-    this.onClick = this.onClick.bind(this);
+    this.keyboardShortcuts = getKeyboardShortcuts(this.props.phrases);
```
ljharb: this would also need to happen in `componentWillReceiveProps`.

erin-doyle: Moving this was not performance motivated, I just wanted to make render() a little more concise.  Sounds like I should just move it back.

ljharb: I think it's a great idea to move it; but if it's not in the render path, then it has to be updated in both of the places where props change.

erin-doyle: Ok I'll make that change in a separate changeset then 👍

src/components/DayPickerKeyboardShortcuts.jsx
```diff
@@ -120,7 +168,14 @@ class DayPickerKeyboardShortcuts extends React.Component {
           )}
           type="button"
           aria-label={toggleButtonText}
-          onClick={this.onClick}
+          onClick={this.onShowKeyboardShortcutsButtonClick}
+          onKeyDown={(e) => {
+            if (e.key === 'Enter') {
+              e.preventDefault();
+            } else if (e.key === 'Space') {
+              this.onShowKeyboardShortcutsButtonClick();
```
ljharb: this should probably pass the `e` through?

src/components/DayPickerKeyboardShortcuts.jsx
```diff
@@ -110,6 +157,7 @@ class DayPickerKeyboardShortcuts extends React.Component {
     return (
       <div>
         <button
+          id="showKeyboardShortcutsButton"
```
ljharb: it seems like this ID only exists so it can be targeted in tests; if so, please remove it - there's surely another way it can be located.

src/components/DayPickerKeyboardShortcuts.jsx
```diff
@@ -37,11 +37,46 @@ class DayPickerKeyboardShortcuts extends React.Component {
...
 
-  onClick() {
+  componentDidUpdate() {
+    this.handleFocus();
+  }
+
+  onKeyDown(e) {
+    const { closeKeyboardShortcutsPanel } = this.props;
+
+    e.stopPropagation();
```
ljharb: why does propagation need to be stopped here?

src/components/DayPickerKeyboardShortcuts.jsx
```diff
@@ -154,26 +210,25 @@ class DayPickerKeyboardShortcuts extends React.Component {
             </div>
 
             <button
+              id="hideKeyboardShortcutsButton"
```
ljharb: also this one

src/components/DayPickerKeyboardShortcuts.jsx
```diff
@@ -154,26 +210,25 @@ class DayPickerKeyboardShortcuts extends React.Component {
...
-              }}
+              onKeyDown={this.onKeyDown}
             >
               <CloseButton {...css(styles.DayPickerKeyboardShortcuts_closeSvg)} />
             </button>
 
-            <ul {...css(styles.DayPickerKeyboardShortcuts__list)}>
+            <ul
+              {...css(styles.DayPickerKeyboardShortcuts__list)}
+              id="DayPickerKeyboardShortcuts__description"
```
ljharb: i'm not sure if this is a pattern we already follow or not; but an ID must be unique on the page, and there should be able to be multiple date of these in the DOM at once - is there another way this can be linked?

src/components/KeyboardShortcutRow.jsx
ljharb: would adding `:` or similar instead of a space have the same effect?

src/components/DayPickerKeyboardShortcuts.jsx
erin-doyle: It's being done that way in DayPicker's `onKeyDown()` so figured it would probably be needed here as well.

src/components/DayPickerKeyboardShortcuts.jsx
erin-doyle: Ok I'll remove and update the test(s)

src/components/DayPickerKeyboardShortcuts.jsx
ljharb: i think it'd be more future-proof if it ever needed it in the future. Not a huge deal obv.

src/components/KeyboardShortcutRow.jsx
erin-doyle: Adding a `:` works great with Safari but in Chrome the punctuation is read so it sounds like:

"Enter key colon Select the date in focus"

I tried a `,` instead and that seems to work nicely and consistently in both browsers.  So I'll change this to a `,` instead of a space

ljharb: awesome!

erin-doyle:
I just want to follow up on this and make sure there weren't any more outstanding actions on my part that you may be waiting for.  I think I got it all but if not let me know!

src/components/DayPicker.jsx
majapw: hmmm, i don't remember. Focusing in the keyboard shortcuts panel may have evolved since I first integrated it and this may be a relic of the initial a11y change.

src/components/DayPickerKeyboardShortcuts.jsx
```diff
@@ -97,7 +70,98 @@ class DayPickerKeyboardShortcuts extends React.Component {
...
+
+      case 'Tab':
+      case 'Enter':
+      case 'Home':
+      case 'End':
+      case 'PageUp':
+      case 'PageDown':
+      case 'ArrowLeft':
+      case 'ArrowRight':
+        e.stopPropagation();
```
majapw: Can you clarify why you're stopping propagation here and below? and preventing default?

src/components/DayPickerKeyboardShortcuts.jsx
```diff
@@ -97,7 +70,98 @@ class DayPickerKeyboardShortcuts extends React.Component {
...
+        e.preventDefault();
+        break;
+
+      case 'ArrowUp':
+      case 'ArrowDown':
+        e.stopPropagation();
```
majapw: Can you leave a comment in here to explain what's going on? I'm not sure it's self-evident

src/components/KeyboardShortcutRow.jsx
```diff
@@ -38,7 +38,7 @@ function KeyboardShortcutRow({
         <span
           {...css(styles.KeyboardShortcutRow_key)}
           role="img"
-          aria-label={label}
+          aria-label={`${label},`}
```
majapw: ?

ljharb: https://github.com/airbnb/react-dates/pull/825#discussion_r148938046

erin-doyle: I can add a comment to the code on this as well.

src/components/DayPickerKeyboardShortcuts.jsx
erin-doyle: Yep no problem!

erin-doyle:
Looking in to build failures now, all I did was rebase on master to pick up any recent changes and then add comments...

erin-doyle:
Linting error:

111:1  error  Line 111 exceeds the maximum line length of 100  max-len

:woman_facepalming:

erin-doyle:
Yay!  🎉 😌

erin-doyle:
Would it be cool to go ahead and push a version update now that this is in?  Patch?  I can do it if y'all are busy but didn't want to assume in case you were holding off for something on its way.

majapw:
v15.1.0 is out!

Issue #828, opened 2017-11-08
Pass modifiers to renderDay to allow flexible CalendarDay display and styling?


ericls:
First of all, thanks for such a great library.

I'm trying to change how days are rendered based on the select/hover state of the calendar.
I'm aware that you can combine onMouseEnter, startDate, endDate, etc to achieve this, but it would be nice to have all those information passed to renderDay directly from modifiers.

majapw:
Yes! I actually have a local branch that does this but haven't had a chance
to merge it in. I'd love to accept a PR for this.

Rickkorsten:
@majapw @ljharb @schleyfox Is there someone who can show me an example how to use modifiers with the calendar?

thx

majapw:
@Rickkorsten I'm not sure that that is related to this issue (which I'm going to close since we merged in the feature). Can you open a new issue with more details on what exactly you're trying to do?

Thanks!

PR #829, opened 2017-11-08
pass day modifiers to renderDay function

ericls:
#828

~~Also exposes the constants?
Not sure if that's a good idea.I see most modifiers are defined/added in the `CalendarDay` component file but `BLOCKED_MODIFIER` is defined in `constants.js`.~~

### DIFF
--- a/src/components/CalendarDay.jsx
+++ b/src/components/CalendarDay.jsx
@@ -163,7 +163,7 @@ class CalendarDay extends React.Component {
           onClick={(e) => { this.onDayClick(day, e); }}
           tabIndex={tabIndex}
         >
-          {renderDay ? renderDay(day) : day.format('D')}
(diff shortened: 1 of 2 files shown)


src/index.js
```diff
@@ -18,3 +18,4 @@ export { default as isSameDay } from './utils/isSameDay';
 export { default as toISODateString } from './utils/toISODateString';
 export { default as toLocalizedDateString } from './utils/toLocalizedDateString';
 export { default as toMomentObject } from './utils/toMomentObject';
+export { default as constants } from './constants';
```
ljharb: This isn't necessary; someone who wants them can require/import from `react-dates/lib/constants` already.

ericls: Removed that line :)

Issue #832, opened 2017-11-09
gh-pages/storybook on master is currently down! :alarm:


majapw:
Working on it!

majapw:
Related to https://github.com/airbnb/react-dates/commit/7eb21918799a9fbec7be615fed9d447e4c4f7c84

Storybook does not rely on the built files and ... I think it's not getting babel transformed in place as expected. @lencioni @ljharb ?

PR #833, opened 2017-11-09
babel transforms jsx files for storybook

majapw:
This is a fix for https://github.com/airbnb/react-dates/issues/832

The removal of the explicit `import React`s from the icon files didn't take into account that those files were not being babel transformed for storybook. This addresses the issue.

(I'm gonna merge through so master can be unfucked, but I'm opening this PR for posterity)

FYI @ljharb @lencioni


ljharb:
LGTM

PR #842, opened 2017-11-14
Add ariaLabelFormat Prop into CalendarDay component

### LINKED ISSUE #778: Localize aria-label for CalendarDay button
Currently, the aria-label format for CalendarDay button is fixed like "Saturday, October 14, 2017".

https://github.com/airbnb/react-dates/blob/master/src/components/CalendarDay.jsx#L104

Is there any way to customize the format?
I guess I can do that by reparsing the text and format using moment but it seems not to be a proper approach.

Thanks!

koba04:
This is a PR to fix #778.
I've added `ariaLabelFormat` for `CalendarDay` and added 'dayAriaLabelFormat` for `CalendarMonth`, `CalendarMonthGrid` and `DayPicker` to pass through `ariaLabelFormat`

### DIFF
--- a/src/components/CalendarDay.jsx
+++ b/src/components/CalendarDay.jsx
@@ -24,6 +24,7 @@ const propTypes = forbidExtraProps({
   onDayMouseEnter: PropTypes.func,
   onDayMouseLeave: PropTypes.func,
   renderDay: PropTypes.func,
+  ariaLabelFormat: PropTypes.string,
 
   // internationalization
   phrases: PropTypes.shape(getPhrasePropTypes(CalendarDayPhrases)),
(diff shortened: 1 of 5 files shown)


src/components/CalendarDay.jsx
```diff
@@ -101,7 +104,7 @@ class CalendarDay extends React.Component {
 
     if (!day) return <td />;
 
-    const formattedDate = { date: `${day.format('dddd')}, ${day.format('LL')}` };
```
ljharb: hm, is there a reason that it was two separate format calls before?

koba04: I'm not sure... but the result is same.

koba04:
Thanks!
Currently, I'm using `DayPicker` directly, so I don't need that.
But it seems to make sense. I'll make a PR for that!

PR #843, opened 2017-11-15
Reset after-hovered-start in componentWillReceiveProps - Fixes #831

### LINKED ISSUE #831: `DayPickerRangeController` leaves ghost values when using a keyboard
http://airbnb.io/react-dates/?selectedKind=DayPickerRangeController&selectedStory=default&full=0&down=1&left=1&panelRight=0&downPanel=storybook%2Factions%2Factions-panel

Steps to reproduce:
1. Tab into calendar with keyboard
2. Select a date with the keyboard via space or enter. Notice that there's a green highlighted date right after the date chosen
3. Select an end date
4. Select a new date not in the existing range to select a new range

The date right after the start date is highlighted differently than the highlight for selected dates within the range. That highlight also doesn't go away when selecting a new range. This leaves highlighted dates on the calendar.

image

ghostdateskeyboard

wyattdanger:
after-hovered-start was being unset in onDayClick, which led to an issue
where the modifier was left as applied when changing dates via keyboard
navigation. This moves that logic into componentWillReceiveProps so that
it works in both cases.

Fixes #831 

Gif of keyboard navigation and changing dates after the fix, for comparison with gif in the issue:
ghostvaluesgone

cc @backwardok @majapw

### DIFF
--- a/src/components/DayPickerRangeController.jsx
+++ b/src/components/DayPickerRangeController.jsx
@@ -206,23 +206,38 @@ export default class DayPickerRangeController extends React.Component {
       numberOfMonths,
       enableOutsideDays,
     } = nextProps;
+
+    const {
+      startDate: prevStartDate,
+      endDate: prevEndDate,
+      focusedInput: prevFocusedInput,
+      minimumNights: prevMinimumNights,
+      isOutsideRange: prevIsOutsideRange,
+      isDayBlocked: prevIsDayBlocked,
+      isDayHighlighted: prevIsDayHighlighted,
+      phrases: prevPhrases,
+      initialVisibleMonth: prevInitialVisibleMonth,
+      numberOfMonths: prevNumberOfMonths,
+      enableOutsideDays: prevEnableOutsideDays,
+    } = this.props;
+
     let { visibleDays } = this.state;
 
     let recomputeOutsideRange = false;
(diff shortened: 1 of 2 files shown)


src/components/DayPickerRangeController.jsx
```diff
@@ -258,6 +258,11 @@ export default class DayPickerRangeController extends React.Component {
     if (didStartDateChange) {
       modifiers = this.deleteModifier(modifiers, this.props.startDate, 'selected-start');
       modifiers = this.addModifier(modifiers, startDate, 'selected-start');
+      if (this.props.startDate) {
+        const startSpan = this.props.startDate.clone().add(1, 'day');
```
backwardok: I wonder if we should do something like
```javascript
const {
  startDate: prevStartDate,
  minimumNights: prevMinimumNights,
  ...
} = this.props
```
so that we're not referencing `this.props` so many times in this function

wyattdanger:
Oh, hello again @erin-doyle!

erin-doyle:
Isn't it a small React world @wyattdanger !  I think this looks good besides the suggestion to unpack this.prop.

wyattdanger:
@backwardok @erin-doyle alright, all this.props references in componentWillReceiveProps have been destructured in a second commit

PR #846, opened 2017-11-17
Use adapter helper

ljharb:
This uses `enzyme-adapter-react-helper` to handle both configuring enzyme, as well as ensuring the proper react version is installed.


ljharb:
(fixed the scripts in 96655cf75b7b34391c1110cfde204deb0b2a5e3a)

PR #851, opened 2017-11-23
Replace react-addons-shallow-compare with React.PureComponent

singingwolfboy:
[react-addons-shallow-compare](https://www.npmjs.com/package/react-addons-shallow-compare) is a legacy React addon, and is no longer maintained. Instead, we should use [`React.PureComponent`](https://reactjs.org/docs/react-api.html#reactpurecomponent).

Note that because `React.PureComponent` was introduced in React 15, this pull request also **drops support for React 0.14**. I realize this will be a controversial change, but I figured that I would propose it anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,6 @@ env:
   global:
     - TEST=true
   matrix:
-    - REACT=0.14
     - REACT=15
     - REACT=16
 sudo: false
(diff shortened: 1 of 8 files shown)


majapw:
Agreed with @ljharb. :) We won't be dropping 0.14 for a while.

developer239:
Really? 😄 `0.14`?

ljharb:
@developer239 yes, really. two major versions back.

PR #857, opened 2017-11-28
Add dayAriaLabelFormat into more places

koba04:
This PR is a follow-up for #842.
I've added `dayAriaLabelFormat` prop into `DayPickerRangeController` and `DayPickerSingleDateController` to pass the prop to `DayPicker`.

### DIFF
--- a/src/components/DayPickerRangeController.jsx
+++ b/src/components/DayPickerRangeController.jsx
@@ -80,6 +80,7 @@ const propTypes = forbidExtraProps({
   monthFormat: PropTypes.string,
   weekDayFormat: PropTypes.string,
   phrases: PropTypes.shape(getPhrasePropTypes(DayPickerPhrases)),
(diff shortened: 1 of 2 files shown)


koba04:
Thanks!

erin-doyle:
Sorry I'm very late to the game on this!  I would think this should also be exposed up through SingleDatePicker and DateRangePicker and noted in the README.  As it is right now it's only usable for those using the Controller versions.

majapw:
@erin-doyle I feel like we ended up doing that in a follow-up somewhere. ... did we not?

erin-doyle:
I'm not seeing it.  I think this PR was the "follow-up" PR to cover exposing this prop up the hierarchy further but I don't think it went far enough.

koba04:
> I would think this should also be exposed up through SingleDatePicker and DateRangePicker and noted in the README.

@erin-doyle I've made a PR for this! #984

Issue #859, opened 2017-11-28
How do I highlight "today" on the calendar? v12.6.0 upgrading to v15.1.0


uvtzxpm:
Using an old version (v12.6.0), this CSS worked:

```
.CalendarDay--today {
	box-shadow: inset 0 0 2px 1px rgba(0,0,255,1);
	font-weight: bold;
}
```

To give this effect:

![image](https://user-images.githubusercontent.com/24996578/33333872-0a036c04-d460-11e7-82f1-11ff529596f1.png)

How can I do the same thing using a recent version?

I have searched and looked at the v13+ changes, but the answer does not seem obvious. Thanks!

majapw:
Hmmm, it looks like this feature got removed by accident. Will add back!

erin-doyle:
This may have happened during the conversion to using `withStyles`.  Previously there was this:

```javascript
const className = cx('CalendarDay', {
    'CalendarDay--outside': isOutsideDay,
}, Array.from(modifiers, mod => `CalendarDay--${mod}`));
```

And within `modifiers` is "today", so it would have set the class `CalendarDay--today`.

majapw:
Put up a PR for a quick fix! :)

uvtzxpm:
@majapw Thanks for the fast fix. It works for me (class changed to `.CalendarDay__today`).

Do you think there will be a release containing this sometime this week?

majapw:
Yup! Working on a release right now :)

majapw:
v15.2.0 is out! :D

PR #861, opened 2017-11-28
Adds back today modifier

majapw:
Quick patch for https://github.com/airbnb/react-dates/issues/859

to: @erin-doyle @uvtzxpm @ljharb


erin-doyle:
Perhaps some tests on these classes getting added when expected would be nice.

majapw:
It's a bit difficult to write a test for this right now because it is a styling detail (and there's only one theme registered for the examples so if we have styling on `today` for one example, we have them for all. I'll think about this some more, but I think I'm gonna merge this in as is.

Issue #29, opened 2016-08-19
[Feature Request] Preset date ranges


ianks:
One thing that is necessary for an app that I am building is preset date ranges. For what I am doing, it is a bad UX to force users to pick specific dates always; oftentimes they want just the data from the last week/month/etc. I understand that for Airbnb this is not a valid use case, but for many other styles of it is. A good mockup for how this might look is from bootstrap-daterange-picker:

![calendar-with-presets](https://camo.githubusercontent.com/1083660a91a3d95a2dd4e8e5e74dfc9703fcdadd/687474703a2f2f692e696d6775722e636f6d2f4c62414d6633442e706e67)

Would the maintainers be open to accepting a PR with the feature if it was not part of the default configuration?

mikesparr:
@ianks you can build your own either as dropdown next to picker, links, or get fancy and add to div.

Simple implementation re-using the change callback

~~~
<a href="#" onClick={() => this.onDatesChange({startDate: moment.utc().subtract(1, 'year'), endDate: moment.utc()})}>Year</a> | &nbsp;
<a href="#" onClick={() => this.onDatesChange({startDate: moment.utc().subtract(3, 'months'), endDate: moment.utc()})}>Qtr</a> | &nbsp;
<a href="#" onClick={() => this.onDatesChange({startDate: moment.utc().subtract(1, 'month'), endDate: moment.utc()})}>Mo</a> | &nbsp;
<a href="#" onClick={() => this.onDatesChange({startDate: moment.utc().subtract(1, 'week'), endDate: moment.utc()})}>Week</a> | &nbsp;
<a href="#" onClick={() => this.onDatesChange({startDate: moment.utc().subtract(1, 'day'), endDate: moment.utc()})}>Day</a> 
~~~

<img width="1201" alt="screen shot 2017-01-16 at 2 37 02 pm" src="https://cloud.githubusercontent.com/assets/5553105/21999865/70f3e25a-dbf9-11e6-8456-3544def7cbb3.png">

sharq1:
Check out our extension:
https://github.com/streamfoundations/react-dates-presets/

![687474703a2f2f73747265616d666f756e646174696f6e732e6769746875622e696f2f72656163742d64617465732d707265736574732f64656d6f2f64656d6f2e676966](https://user-images.githubusercontent.com/1241239/27733871-a8b7edd2-5d98-11e7-8a35-73a63f69f09a.gif)

liubko:
[...]

PR #862, opened 2017-11-28
Add DateRangePicker example with presets

majapw:
Somewhat of a fix for #29 

So I am loath to introduce this as a separate prop because it is already possible to accomplish with the current API (and the current API allows for full customization of however you'd want to make this look). This adds an example for how exactly you might set up preset date ranges using `renderCalendarInfo` and a regular ol' `DateRangePicker`.

I think I still need to get this working with the css storybook, but here's a screenshot of the aphrodite storybook example:
screen shot 2017-11-28 at 12 06 54 pm

### DIFF
--- a/.storybook-css/config.js
+++ b/.storybook-css/config.js
@@ -67,6 +67,7 @@ function loadStories() {
   require('../stories/DayPickerRangeController');
   require('../stories/DayPickerSingleDateController');
   require('../stories/DayPicker');
+  require('../stories/PresetDateRangePicker');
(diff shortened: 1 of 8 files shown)


erin-doyle:
This is a great idea to provide an example of how to do this as it sounds like there are enough people wanting to and I agree the API supports it just fine out of the box. 👍

majapw:
I think I should add some usage details! and probably a link to the example. I will do that.

PR #863, opened 2017-11-28
Allow react-portal v4

### LINKED ISSUE #860: react-portal v4?
v4 is out and supports only React 16+, but react-dates does not allow it to be used. Is there a way to add support for v4 here?

lencioni:
This is the version you will want to use if you are on React 16. It
looks like the API is the same, so it should be safe to allow this.

Fixes #860 

Note: I haven't tested this at all, so not sure if it actually works.

### DIFF
--- a/package.json
+++ b/package.json
@@ -118,7 +118,7 @@
     "prop-types": "^15.5.10",
     "react-addons-shallow-compare": "^15.5.2",
     "react-moment-proptypes": "^1.5.0",
-    "react-portal": "^3.1.0",
+    "react-portal": "^3.1.0 || ^4.0.0",
     "react-with-styles": "^2.2.0",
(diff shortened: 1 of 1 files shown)


majapw:
> Note: I haven't tested this at all, so not sure if it actually works.

:|

majapw:
Hmm, it looks like something is not going well with the tests. It seems like Enzyme is unable to parse the `react-portal` export as a selector.

majapw:
Because the API has in fact changed between the versions, I'm not sure that we can do this. :/

lencioni:
Thanks for the investigation. 

> Because the API has in fact changed between the versions, I'm not sure that we can do this. :/

That's supremely disappointing.

Issue #849, opened 2017-11-21
Option to disable animation


wachunga:
It appears there is no way to disable the transitions/animation when navigating through months. I would like to use this library but the animation doesn't jive with my company's style guide.

I've tried overriding `transition-duration` etc but the navigation breaks after navigating back a month or two (using SingleDatePicker). The [inline styles in CalendarMonthGrid](https://github.com/airbnb/react-dates/blob/master/src/components/CalendarMonthGrid.jsx#L255) are also difficult to work with.

Could eg `animated` be a prop? Perhaps I've overlooked a simpler solution?

ljharb:
A prop to disable animations/transitions makes sense to me.

majapw:
This seems like a good idea to me too! I think it's been floated before.
Happy to accept a PR.

majapw:
Released in 15.3.0! :)

Issue #834, opened 2017-11-09
Reduce transition duration


M-Willett:
First of all, awesome job on this! Very nice and flexible. 

Our users feel that the transition duration is too slow. The component waits for the end of the transition before being able to skip ahead without delay. I had a quick search through the list of issues but surprisingly no one has reported this.

I've tried setting a `transition-duration: .1s !important` property which creates a weird card-shuffling effect. I then tried `transition: none !important` to remove completely but since the pagination relies on the `transitionEnd` it breaks the pagination. (I've found and tried `transformValue`  as a string but this didn't change anything afaik.)

Is there/could there be a default prop for setting this?

majapw:
I think that's probably a good idea! I would support this added prop.

majapw:
Released in 15.3.0! :)

PR #865, opened 2017-11-29
Add transitionDuration prop

majapw:
Fix for https://github.com/airbnb/react-dates/issues/834 and also for https://github.com/airbnb/react-dates/issues/849

This change allows you to set the `transitionDuration` manually (including setting it to 0).

FYI: @M-Willett @wachunga

### DIFF
--- a/src/components/CalendarMonthGrid.jsx
+++ b/src/components/CalendarMonthGrid.jsx
@@ -50,6 +50,7 @@ const propTypes = forbidExtraProps({
   firstDayOfWeek: DayOfWeekShape,
   setCalendarMonthHeights: PropTypes.func,
   isRTL: PropTypes.bool,
+  transitionDuration: nonNegativeInteger,
 
   // i18n
   monthFormat: PropTypes.string,
@@ -78,6 +79,7 @@ const defaultProps = {
   firstDayOfWeek: null,
   setCalendarMonthHeights() {},
   isRTL: false,
+  transitionDuration: 200,
 
   // i18n
   monthFormat: 'MMMM YYYY', // english locale
@@ -166,11 +168,17 @@ class CalendarMonthGrid extends React.Component {
   }
 
   componentDidUpdate(prevProps) {
-    const { isAnimating, onMonthTransitionEnd, setCalendarMonthHeights } = this.props;
+    const {
+      isAnimating,
+      transitionDuration,
+      onMonthTransitionEnd,
+      setCalendarMonthHeights,
+    } = this.props;
 
     // For IE9, immediately call onMonthTransitionEnd instead of
-    // waiting for the animation to complete
-    if (!this.isTransitionEndSupported && isAnimating) {
+    // waiting for the animation to complete. Similarly, if transitionDuration
+    // is set to 0, also immediately invoke the onMonthTransitionEnd callback
+    if ((!this.isTransitionEndSupported || !transitionDuration) && isAnimating) {
       onMonthTransitionEnd();
     }
 
@@ -232,6 +240,7 @@ class CalendarMonthGrid extends React.Component {
       styles,
       phrases,
       dayAriaLabelFormat,
+      transitionDuration,
     } = this.props;
 
     const { months } = this.state;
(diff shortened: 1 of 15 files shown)


wachunga:
Nice work, @majapw.

src/components/DateRangePicker.jsx
```diff
@@ -79,6 +79,7 @@ const defaultProps = {
   isRTL: false,
   firstDayOfWeek: null,
   verticalHeight: null,
+  transitionDuration: 0.2,
```
ljharb: does this need a real defaultProp value? If this is set to `undefined`, then it will delegate to the `defaultProp` in the DayPickerRangeController instead, rather than duplicating it here.

src/components/DayPicker.jsx
```diff
@@ -102,6 +103,7 @@ export const defaultProps = {
   daySize: DAY_SIZE,
   isRTL: false,
   verticalHeight: null,
+  transitionDuration: 0.2,
```
ljharb: also here; if this is `undefined`, it will delegate to `CalendarMonthGrid` rather than duplicating the default prop value here

src/components/DayPickerRangeController.jsx
```diff
@@ -121,6 +122,7 @@ const defaultProps = {
   renderCalendarInfo: null,
   firstDayOfWeek: null,
   verticalHeight: null,
+  transitionDuration: 0.2,
```
ljharb: also here

src/components/DayPickerSingleDateController.jsx
```diff
@@ -102,6 +103,7 @@ const defaultProps = {
   firstDayOfWeek: null,
   daySize: DAY_SIZE,
   verticalHeight: null,
+  transitionDuration: 0.2,
```
ljharb: and here

src/components/SingleDatePicker.jsx
```diff
@@ -72,6 +72,7 @@ const defaultProps = {
   daySize: DAY_SIZE,
   isRTL: false,
   verticalHeight: null,
+  transitionDuration: 0.2,
```
ljharb: ditto

src/components/CalendarMonthGrid.jsx
```diff
@@ -253,8 +262,8 @@ class CalendarMonthGrid extends React.Component {
           isVertical && styles.CalendarMonthGrid__vertical,
           isVerticalScrollable && styles.CalendarMonthGrid__vertical_scrollable,
           isAnimating && styles.CalendarMonthGrid__animating,
-          isAnimating && {
-            transition: 'transform 0.2s ease-in-out',
+          isAnimating && transitionDuration && {
+            transition: `transform ${transitionDuration}s ease-in-out`,
```
lencioni: What do you think about using `ms` here instead of `s` and then making transitionDuration an integer? Also, should the unit be part of the prop name for ultimate clarity? `transitionDurationMS` or something like that?

lencioni: Also, are folks going to want to control the easing? I often find ease-out to be nicer than ease-in-out for example.

majapw: I'll make the change to ms. I am adding the change to transition duration because it's been requested but it might make sense to control the easing as well. We can do that in a follow-up PR.

lencioni: For sure, but it might be worth considering what the ideal API for this would be before merging in this PR. e.g. do you want it to be 2 separate props, 1 prop that takes an options object, or a string prop that controls all of the transition options at the same time? I think probably 2 separate props, but you would probably know what is better for this project than I.

majapw: That is fair. I think it would be clearest and make the most sense to add a `transitionEasing` prop that would accept valid options as per the specification. It seems like it would be more confusing to have one prop cover both options.

majapw:
@ljharb restamp me pls? :)

src/components/CalendarMonthGrid.jsx
```diff
@@ -78,6 +79,7 @@ const defaultProps = {
   firstDayOfWeek: null,
   setCalendarMonthHeights() {},
   isRTL: false,
+  transitionDurationMS: 200,
```
ljharb: i don't know if the "MS" prefix is actually helpful here; durations in JS are almost always in milliseconds already (like setTimeout)

majapw: Well except for the fact that we were previously doing the transition in seconds... I think it is still useful metadata to have in the prop name.

ljharb: previously, we weren't accepting a transition at all, so nobody knew the unit. That in a previous iteration of the PR it took seconds isn't relevant.

ljharb: Good call; it should be documented regardless.

majapw:
I change the name back to `transitionDuration` and clarified the unit in the README. PTAL @erin-doyle @ljharb

Issue #326, opened 2017-02-11
required id property in DateInput, SingleDatePicker, SingleDatePickerInput


Artur-A:
Hi,

DateInput, SingleDatePicker and SingleDatePickerInput has `id` property, which is required.
I cannot see why it is done as mandatory field. 

Is it done on purpose or the field can be optional?

See [src/components/DateInput.jsx#L7](https://github.com/airbnb/react-dates/blob/master/src/components/DateInput.jsx#L7)

and
```
        <input
          id={id}
          name={id}
```

ljharb:
It's done on purpose. For accessibility, all inputs must be both wrapped in a `<label>` that it is *also* linked to with `id`/`htmlFor`. IDs must be unique on the page, so a reusable component can't safely generate them. Typically, you'd want to pass them down from the server, since that's the only place you know for sure a component is only being used once on the page.

(also for accessibility, inputs shouldn't be used outside of forms)

Artur-A:
Thank you Jordan.

Got the idea. 
But there are still some confusion in my head. Hope to clear it up:

* label requires `id` for assistive technologies (like screen readers). At the same time `aria-label` is already used for the same purpose in `DateInput.jsx`. And the field `aria-label` does not require `id` at all.

* for a form posting `name` field is required, but it can be done as an optional field and has `name` property in control instead of `id` duplication. Some date controls are not placed in the form => `name` can be optional.

* `DateRangePicker` control uses __optional__ `startDateId` and `endDateId` props and feels them with default string ids. The second instance of DateRangePicker on the screen without ids overriding will break DOM id uniqueness. Under the hood `startDateId` and `endDateId` are forwarded to `DateInput.id`. `DateRangePicker` ids are made _optional_, but `DateInput, SingleDatePicker, SingleDatePickerInput` ids are made as _mandatory_ fields.

* it seems that `id` can be optional

ljharb:
You're right, `DateRangePicker` should have `startDateId` and `endDateId` props required as well. Thanks for pointing out the omission!

h0jeZvgoxFepBQ2C:
This should be renamed to something more obvious, needed to browse the web to understand why startDateId is required. Also it shouldnt be required imo.

h0jeZvgoxFepBQ2C:
Also I dont use any labels around or a form, I just use the component via onChange callback, so there are no labels at all.

ljharb:
[...]

PR #866, opened 2017-11-29
Remove startDateId and endDateId default prop values

majapw:
Fix for https://github.com/airbnb/react-dates/issues/326

The ids were made required but their default prop values were not removed and as such the `required` aspect was useless. This addresses that issue.

This is technically breaking.

### DIFF
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -46,9 +46,7 @@ const defaultProps = {
   focusedInput: null,
 
   // input related props
-  startDateId: START_DATE,
   startDatePlaceholderText: 'Start Date',
-  endDateId: END_DATE,
   endDatePlaceholderText: 'End Date',
   disabled: false,
   required: false,


ljharb:
@majapw why is this breaking? If the props are required, then the defaultProp was never needed.

We definitely shouldn't worry about someone who was ignoring propType warnings.

majapw:
@ljharb the props have been required for a while, but because default props were also provided, then people wouldn't see an error when not using their own ids. if we remove the default props, then people who were previously not seeing an error will now see one.  that was my motivation for marking it as breaking.

ljharb:
Hmm, I guess that's technically breaking, but I also think people ignoring propType warnings are OK to break :-p

JcBernack:
None of the documentation is updated to reflect this change, it still says "required: no". Also I can't find documentation what `startDateId` and `endDateId` actually even do or why I would want to set them. In the readme they are listed under OPTIONAL props (note the caps).

> but because default props were also provided, then people wouldn't see an error when not using their own ids

Noone was ignoring an error, because there never was one. Removing the defaultProp absolutely was a breaking change. /rant 😉

erin-doyle:
It looks like we have a few propType warnings when running the tests now.  Not a big deal but kind of annoying:

```
Warning: Failed prop type: The prop `startDateId` is marked as required in `withStyles(DateRangePicker)`, but its value is `undefined`.
    in withStyles(DateRangePicker)
Warning: Failed prop type: The prop `endDateId` is marked as required in `withStyles(DateRangePicker)`, but its value is `undefined`.
    in withStyles(DateRangePicker)
Warning: Failed prop type: The prop `startDateId` is marked as required in `DateRangePicker`, but its value is `undefined`.
    in DateRangePicker
Warning: Failed prop type: The prop `endDateId` is marked as required in `DateRangePicker`, but its value is `undefined`.
    in DateRangePicker
```

ljharb:
Fair enough, my bad.

@majapw, should we revert this and publish a v16.0.2, and then rerevert it for a v17?

majapw:
That seems reasonable to me! It'd be good to add a blurb about the id as
well in the README.

I am sans computer for ten days so would appreciate if you did that @ljharb.

Sorry for the warnings!

ljharb:
No prob, I'll take care of it today (revert + publish + put up rerevert PR)

ljharb:
Whoops, this is actually in v15. I'll revert it in v15.5.2 tonight/tomorrow, but leave it in v16 (and add the docs).

ljharb:
v15.5.2 is published; i'll follow up with a docs PR for v16 this weekend.

PR #869, opened 2017-11-30
Add borderless option to the `DayPicker` variations

majapw:
Sometimes you don't want the DayPicker to have a border (or you want to have full control of the border prop yourself)! This adds that option.

This is one PR in what will probably be a series of style props to help get airbnb.com onto the latest and greatest. Hopefully these new additions will help others customize to their heart's content as well! :) 

to: @ljharb @erin-doyle @noratarano @airbnb/webinfra

### DIFF
--- a/README.md
+++ b/README.md
@@ -262,6 +262,7 @@ The following is a list of other *OPTIONAL* props you may provide to the `DayPic
   renderCalendarInfo: PropTypes.func,
   onOutsideClick: PropTypes.func,
   keepOpenOnDateSelect: PropTypes.bool,
+  noBorder: PropTypes.bool,
 
   // navigation related props
   navPrev: PropTypes.node,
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -262,6 +262,7 @@ The following is a list of other *OPTIONAL* props you may provide to the `DayPic
   renderCalendarInfo: PropTypes.func,
   onOutsideClick: PropTypes.func,
   keepOpenOnDateSelect: PropTypes.bool,
+  borderless: PropTypes.bool,
```
lencioni: I generally like to name props like this `noFoo` instead of `fooless` because I think it is easier to be more consistent that way.

majapw: sounds reasonable to me!

majapw:
PTAL @lencioni

ljharb:
(linking to #870)

Issue #538, opened 2017-05-29
SingleDatePicker: Vertical Spacing Between Picker and Input Can't be Given a Value in Ems


ethanresnick:
Hi, I'm using the SingleDatePicker component and want to position the date picker widget a certain number of ems away from the input, rather than a certain number of pixels, because (per #537) my input changes font-size/height at different breakpoints. However, if I set `$react-dates-spacing-vertical-picker` in ems, rather than pixels, Sass complains because that variable is added to another variable (`$react-dates-width-tooltip-arrow`) which is set in pixels, and Sass obviously can't do math between ems and px. If I set `$react-dates-width-tooltip-arrow` in ems as well, the Sass error goes away, but the tooltip arrow breaks; specifically, the white covering that hides most of the tooltip arrow (all but the part that will visually become the 'edge') is mispositioned. See below, where the white arrow covering is overlapping the prior month button:

<img width="244" alt="screen shot 2017-05-29 at 2 41 26 am" src="https://cloud.githubusercontent.com/assets/471894/26539089/99a3df8e-4418-11e7-93eb-8b98666feaff.png">

So there are really two issues here:

1. That setting the arrow width in ems doesn't seem to work, which is probably just a bug. (In my example, I've set `$react-dates-width-tooltip-arrow: 1.25em;` and `$react-dates-spacing-vertical-picker: 3.225em;`.)

2. More broadly, it might be nice if different variables could be set in different units. This is supported in browsers with `calc()`, of course, so the Sass could simply output calc() expressions. (Unless react-dates targets non-calc-supporting browsers?) Alternatively, since `calc()` makes the output a little longer and maybe has some performance cost, react-dates could maybe use a sass mixin that would output a precomputed value when the units are compatible and a calc expression when they aren't?

ljharb:
react-dates does indeed support browsers without calc, but regardless, calc has abysmal performance, so we'd want to avoid that at all costs if we can.

It seems like perhaps the proper solution here is to always use font-sensitive units here (em) instead of font-insensitive units (px). @majapw?

majapw:
@ljharb we can do that.

PR #883, opened 2017-12-04
Add verticalSpacing prop

majapw:
This prop allows for greater customization of the space in between the picker and the inputs. Additionally, if that value is less than the height of the fang then we hide the fang. To accomplish this, I ended up revamping the fang so that it would use an svg instead of pseudo-selectors. The reasoning for this is that because I had to dynamically position the fang relative to the inputs (based on a prop), I had to leverage inline styles to do so. Unfortunately, the `::before` and `::after` selectors don't allow for that. 

I also renamed caret to fang everywhere... which is like ¯\\\_(ツ)_/¯ but consistency? 

This does not, however, solve https://github.com/airbnb/react-dates/issues/538... so maybe expanding the prop to allow for `em` values as well would be useful. However, this would maybe make the fang show/hide logic a bit more complicated (and have to be manual???). What are your thoughts? Maybe that belongs in a follow-up PR?

to: @airbnb/webinfra @ljharb @noratarano @ethanresnick @gabergg @erin-doyle

### DIFF
--- a/src/components/DateInput.jsx
+++ b/src/components/DateInput.jsx
@@ -1,12 +1,24 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import { forbidExtraProps } from 'airbnb-prop-types';
+import { forbidExtraProps, nonNegativeInteger } from 'airbnb-prop-types';
 import { css, withStyles, withStylesPropTypes } from 'react-with-styles';
 import throttle from 'lodash/throttle';
 import isTouchDevice from 'is-touch-device';
 
+import getInputHeight from '../utils/getInputHeight';
 import openDirectionShape from '../shapes/OpenDirectionShape';
-import { OPEN_DOWN, OPEN_UP } from '../constants';
+import {
+  OPEN_DOWN,
+  OPEN_UP,
+  FANG_HEIGHT_PX,
+  FANG_WIDTH_PX,
+  DEFAULT_VERTICAL_SPACING,
+} from '../constants';
+
+const FANG_PATH_TOP = `M0,${FANG_HEIGHT_PX} ${FANG_WIDTH_PX},${FANG_HEIGHT_PX} ${FANG_WIDTH_PX / 2},0z`;
+const FANG_STROKE_TOP = `M0,${FANG_HEIGHT_PX} ${FANG_WIDTH_PX / 2},0 ${FANG_WIDTH_PX},${FANG_HEIGHT_PX}`;
+const FANG_PATH_BOTTOM = `M0,0 ${FANG_WIDTH_PX},0 ${FANG_WIDTH_PX / 2},${FANG_HEIGHT_PX}z`;
+const FANG_STROKE_BOTTOM = `M0,0 ${FANG_WIDTH_PX / 2},${FANG_HEIGHT_PX} ${FANG_WIDTH_PX},0`;
 
 const propTypes = forbidExtraProps({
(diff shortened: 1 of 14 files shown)


src/components/DateRangePicker.jsx
```diff
@@ -338,6 +343,10 @@ class DateRangePicker extends React.Component {
       <CloseButton {...css(styles.DateRangePicker_closeButton_svg)} />
     );
 
+    const inputHeight = parseInt(reactDates.font.input.lineHeight, 10)
```
ljharb: this logic is duplicated; can it be shared?

src/components/SingleDatePicker.jsx
```diff
@@ -361,12 +364,18 @@ class SingleDatePicker extends React.Component {
...
       transitionDuration,
+      verticalSpacing,
+      theme: { reactDates },
     } = this.props;
     const { dayPickerContainerStyles, isDayPickerFocused, showKeyboardShortcuts } = this.state;
 
     const onOutsideClick = (!withFullScreenPortal && withPortal) ? this.onClearFocus : undefined;
     const closeIcon = customCloseIcon || (<CloseButton />);
 
+    const inputHeight = parseInt(reactDates.font.input.lineHeight, 10)
```
ljharb: this logic is duplicated; can it be shared?

src/components/DateInput.jsx
```diff
@@ -148,28 +161,34 @@ class DateInput extends React.Component {
...
+      theme: { reactDates: { font, spacing } },
     } = this.props;
 
     const value = displayValue || dateString || '';
     const screenReaderMessageId = `DateInput__screen-reader-message-${id}`;
 
-    const withCaret = showCaret && focused;
+    const withFang = showFang && focused;
+
+    const inputHeight = parseInt(font.input.lineHeight, 10)
```
ljharb: this logic is triplicated, can it be shared?

src/components/DateInput.jsx
```diff
@@ -196,6 +215,31 @@ class DateInput extends React.Component {
           aria-describedby={screenReaderMessage && screenReaderMessageId}
         />
 
+        {withFang &&
```
ljharb: please surround multiline jsx with parens

kashifshamaz21:
@majapw I'm trying to use `verticalSpacing` prop in `DateRangePicker.jsx`, but for some reason I'm seeing that the prop's value is being received as 0 inside the render function of DateRangePicker.

```
<DateRangePicker
          displayFormat="DD/MM/YYYY"
          startDate={this.props.startDate} // momentPropTypes.momentObj or null,
          startDateId="startDate" // PropTypes.string.isRequired,
          endDate={this.props.endDate} // momentPropTypes.momentObj or null,
          endDateId="endDate" // PropTypes.string.isRequired,
          onDatesChange={this.props.dateRangeUpdated}
          focusedInput={this.props.focusedInput} // PropTypes.oneOf([START_DATE, END_DATE]) or null,
          onFocusChange={this.props.onFocusChange} // PropTypes.func.isRequired,
          hideKeyboardShortcutsPanel={true}
          isOutsideRange={this.isOutsideRange.bind(this)}
          verticalSpacing={10}
        />
```
Any pointers on what could be causing this issue? I looked at the code in `DateRangePicker.jsx` related to verticalSpacing, and i didn't find anything obvious which could lead to above behaviour.

Issue #349, opened 2017-02-26
Week picker


tleunen:
I have a use case of a date picker but instead of having to pick a single date, or a range of date, I'd need to pick an entire week.

The way I'm seeing this is hovering on a day of a week will highlight the entire week, and when clicking on any day of the week will select a range between the first and last day of the week.

With the current state, is is something possible? I checked the props and examples and it seems not but I'd like to be sure before starting working on it if you're interested in having it within react-dates.

daltones:
I guess this is currently possible.

In a common use case for `DayPicker`, we use `onDayMouseEnter` and `onDayMouseLeave` to save the hovered day, then `modifiers` with `hovered` state to mark visually that day.

You just need a `hovered` function that returns `true` for _every_ day of the same week of that saved hovered day (not just the exactly same day of the common case).

You can use the same approach for `onDayClick` and `modifiers` with `selection-span`.

pcatalina:
@tleunen did you finally arrive to pick a week by hovering over a date with react-dates ?

tleunen:
To be honest, I haven't taken the time to investigate and make it work. But as @daltones said, I believe it's possible without too much work.

derwaldgeist:
I need the same for selecting whole months and whole years. It would be perfect if the widget could be configured in a way that it selects either days, weeks, months, years or user-defined time-spans.

majapw:
It's def trivial to use the datepicker to click on a day and then select the week. It's a bit more challenging to hover the whole week when hovering over a single day. 

You could also, as @daltones said, roll your own custom DayPicker config to make it work.

a `DayPickerWeekController` might be an interesting component that we'd probably considering rolling into the library.

derwaldgeist:
Well, it’s not that trivial if you consider that there are different calendar types and different first days of a week. And yes, hovering should work, too, to give the user a consistent feedback. This is especially important if the widget auto-closes after selection.

jakeclements:
I've had a requirement for this functionality as well, I've had a chance to take a look at it and have submitted a PR.
Feedback & change requests are welcome

tleunen:
I see the PR has landed, so I'm closing the issue. Thanks again @jakeclements for your work. It will be very helpful!

sht5:
I managed to do this by overriding the renderCalendarDay method. i give it the relevant classes and add a hover action to keep a state of selected week which i highlight in the same method.
woohoooo!
https://codesandbox.io/s/5nk78l76x

PR #884, opened 2017-12-06
[Issue 349] Add single-click range picker

jakeclements:
This adds the feature requested in #349 by allowing a range prop to be added to the DayPickerRangeController.
The range prop requires an object with optional before / after keys, these expect a function that is passed a moment object of the hovered day.
Return a moment object from each function to specify the range.

Eg. Selecting 5 days (hovered day is not inclusive)
```js
range = {
  before: day => day.subtract(2, 'days'),
  after: day => day.add(2, 'days')
}
```

I've added storybook examples to show how this works across a couple of scenarios.
I'm happy to hear feedback on this and make changes to ensure it fits within your project

single-click range picker demo

### DIFF
--- a/examples/DayPickerRangeControllerWrapper.jsx
+++ b/examples/DayPickerRangeControllerWrapper.jsx
@@ -18,6 +18,8 @@ const propTypes = forbidExtraProps({
   autoFocusEndDate: PropTypes.bool,
   initialStartDate: momentPropTypes.momentObj,
   initialEndDate: momentPropTypes.momentObj,
+  startDateOffset: PropTypes.func,
+  endDateOffset: PropTypes.func,
 
   keepOpenOnDateSelect: PropTypes.bool,
   minimumNights: PropTypes.number,
@@ -53,6 +55,8 @@ const defaultProps = {
   autoFocusEndDate: false,
   initialStartDate: null,
   initialEndDate: null,
+  startDateOffset: undefined,
+  endDateOffset: undefined,
 
   // day presentation and interaction related props
   renderCalendarDay: undefined,
--- a/src/components/CalendarDay.jsx
+++ b/src/components/CalendarDay.jsx
@@ -129,6 +129,7 @@ class CalendarDay extends React.Component {
           styles.CalendarDay__default,
           isOutsideDay && styles.CalendarDay__outside,
           modifiers.has('today') && styles.CalendarDay__today,
+          modifiers.has('hovered-offset') && styles.CalendarDay__hovered_offset,
           modifiers.has('highlighted-calendar') && styles.CalendarDay__highlighted_calendar,
           modifiers.has('blocked-minimum-nights') && styles.CalendarDay__blocked_minimum_nights,
           modifiers.has('blocked-calendar') && styles.CalendarDay__blocked_calendar,
@@ -187,6 +188,12 @@ export default withStyles(({ reactDates: { color, font } }) => ({
       border: `1px double ${color.core.borderLight}`,
       color: 'inherit',
     },
+  },
+
+  CalendarDay__hovered_offset: {
+    background: color.core.borderBright,
+    border: `1px double ${color.core.borderLight}`,
+    color: 'inherit',
   },
 
   CalendarDay__outside: {
--- a/src/components/DayPickerRangeController.jsx
+++ b/src/components/DayPickerRangeController.jsx
@@ -18,6 +18,8 @@ import isBeforeDay from '../utils/isBeforeDay';
 import getVisibleDays from '../utils/getVisibleDays';
 import isDayVisible from '../utils/isDayVisible';
 
+import getSelectedDateOffset from '../utils/getSelectedDateOffset';
+
 import toISODateString from '../utils/toISODateString';
 import toISOMonthString from '../utils/toISOMonthString';
 
@@ -39,6 +41,8 @@ const propTypes = forbidExtraProps({
   startDate: momentPropTypes.momentObj,
   endDate: momentPropTypes.momentObj,
   onDatesChange: PropTypes.func,
+  startDateOffset: PropTypes.func,
+  endDateOffset: PropTypes.func,
 
   focusedInput: FocusedInputShape,
   onFocusChange: PropTypes.func,
@@ -92,6 +96,8 @@ const defaultProps = {
   startDate: undefined, // TODO: use null
   endDate: undefined, // TODO: use null
   onDatesChange() {},
+  startDateOffset: undefined,
+  endDateOffset: undefined,
 
   focusedInput: null,
   onFocusChange() {},
@@ -170,6 +176,7 @@ export default class DayPickerRangeController extends React.Component {
       'last-in-range': day => this.isLastInRange(day),
       hovered: day => this.isHovered(day),
       'hovered-span': day => this.isInHoveredSpan(day),
+      'hovered-offset': day => this.isInHoveredSpan(day),
       'after-hovered-start': day => this.isDayAfterHoveredStartDate(day),
     };
 
@@ -423,6 +430,8 @@ export default class DayPickerRangeController extends React.Component {
       onFocusChange,
       onClose,
       onDatesChange,
+      startDateOffset,
+      endDateOffset,
     } = this.props;
 
     if (e) e.preventDefault();
@@ -430,7 +439,15 @@ export default class DayPickerRangeController extends React.Component {
 
     let { startDate, endDate } = this.props;
 
-    if (focusedInput === START_DATE) {
+    if (startDateOffset || endDateOffset) {
+      startDate = getSelectedDateOffset(startDateOffset, day);
+      endDate = getSelectedDateOffset(endDateOffset, day);
+
+      if (!keepOpenOnDateSelect) {
+        onFocusChange(null);
+        onClose({ startDate, endDate });
+      }
+    } else if (focusedInput === START_DATE) {
       onFocusChange(END_DATE);
 
       startDate = day;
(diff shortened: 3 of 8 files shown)


examples/DayPickerRangeControllerWrapper.jsx
```diff
@@ -18,6 +18,10 @@ const propTypes = forbidExtraProps({
   autoFocusEndDate: PropTypes.bool,
   initialStartDate: momentPropTypes.momentObj,
   initialEndDate: momentPropTypes.momentObj,
+  range: PropTypes.shape({
+    before: PropTypes.func,
```
ljharb: This propType also allows an empty object; it seems like either one or both of these properties is required? If so, perhaps something like `or([justBefore, justAfter, bothBeforeAndAfter])` might be better?

src/components/DayPickerRangeController.jsx
```diff
@@ -39,6 +39,10 @@ const propTypes = forbidExtraProps({
   startDate: momentPropTypes.momentObj,
   endDate: momentPropTypes.momentObj,
   onDatesChange: PropTypes.func,
+  range: PropTypes.shape({
```
ljharb: please store this propType somewhere reusable, so it can be shared wherever it's needed.

src/components/DayPickerRangeController.jsx
```diff
@@ -413,36 +418,49 @@ export default class DayPickerRangeController extends React.Component {
   }
 
   onDayClick(day, e) {
-    const { keepOpenOnDateSelect, minimumNights, onBlur } = this.props;
+    const {
+      keepOpenOnDateSelect, minimumNights, onBlur, range,
```
ljharb: if the entire statement can't fit on one line, each property needs to be on its own line.

src/components/DayPickerRangeController.jsx
```diff
@@ -413,36 +418,49 @@ export default class DayPickerRangeController extends React.Component {
...
     if (e) e.preventDefault();
     if (this.isBlocked(day)) return;
 
     const { focusedInput, onFocusChange, onClose } = this.props;
     let { startDate, endDate } = this.props;
 
-    if (focusedInput === START_DATE) {
-      onFocusChange(END_DATE);
+    if (range) {
+      startDate = range.before ? range.before(day.clone()) : day.clone();
```
ljharb: there's 4 clones here; are all 4 needed?

src/components/DayPickerRangeController.jsx
```diff
@@ -413,36 +418,49 @@ export default class DayPickerRangeController extends React.Component {
...
+      endDate = range.after ? range.after(day.clone()) : day.clone();
+      if (!keepOpenOnDateSelect) {
+        onFocusChange(null);
+        onClose({ startDate, endDate });
+      }
+    }
 
-      startDate = day;
+    if (!range) {
```
ljharb: this seems like it should be an `else` attached to the previous conditional block?

src/components/DayPickerRangeController.jsx
```diff
@@ -457,51 +475,76 @@ export default class DayPickerRangeController extends React.Component {
...
-      modifiers = this.deleteModifier(modifiers, hoverDate, 'hovered');
-      modifiers = this.addModifier(modifiers, day, 'hovered');
 
-      if (startDate && !endDate && focusedInput === END_DATE) {
-        if (isAfterDay(hoverDate, startDate)) {
-          const endSpan = hoverDate.clone().add(1, 'day');
-          modifiers = this.deleteModifierFromRange(modifiers, startDate, endSpan, 'hovered-span');
-        }
+      if (range && (range.before || range.after)) {
+        const start = range.before ? range.before(day.clone()) : day.clone();
```
ljharb: this code seems repeated; can it live in somewhere shared?

examples/DayPickerRangeControllerWrapper.jsx
jakeclements: That's right, at least one is required, I can make an update here.

src/components/DayPickerRangeController.jsx
jakeclements: 👍

src/components/DayPickerRangeController.jsx
jakeclements: 👍

src/components/DayPickerRangeController.jsx
jakeclements: They're not, I'll clean this up.

src/components/DayPickerRangeController.jsx
jakeclements: You bet

src/components/DayPickerRangeController.jsx
jakeclements: 👍

jakeclements:
Appreciate the feedback @ljharb, super quick.
All sounds good to me!

jakeclements:
Thanks for taking the time to review @majapw!

> 1. Do we want to bubble up this functionality to the DateRangePicker as well?

While it's completely possible the functionality of it may be confusing. As far as I'm aware, the standard for the DateRangePicker is always to select the start and end date seperately (please correct me if I'm wrong here). 
If we were to introduce this I think it'd be helpful to specify it with another prop identifier other than `range`. 
It isn't immediately clear to me that this would create a single-click selection in this case.


> 2. Does the range functionality essentially mean you select a startDate and endDate with one click and is that... dramatically different enough to require its own picker?

That's correct. I've bundled the functionality into DayPickerRangeController because while the interaction is different, the outcome is still range selection.
If we were to create another picker I think that we may find ourselves repeating a lot of what has already been achieved in DayPickerRangeController.
We might be able to work on some of the terminology used to make this clearer. Open to hearing your thoughts here.


> 3. I find the name of the range prop and the getRangeDay method to be incredibly confusing.  

Naming things is hard, I agree. Let's work on this.


> I think there's also some overloading of what the hovered modifier actually does. ... something like rangeModifiers might be better?

Good call, I can update this so it doesn't affect normal hovering. I'll add a modifier that follows whatever we decide on for our naming conventions.


Happy to hear your feedback on the above.
I'll have more of a think about how we could word the language to make some of this functionality clearer and will add any suggestions as comments here.

src/components/CalendarDay.jsx
```diff
@@ -193,6 +194,12 @@ export default withStyles(({ reactDates: { color } }) => ({
     },
   },
 
+  CalendarDay__hovered: {
+    background: color.core.borderBright,
```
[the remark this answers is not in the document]
jakeclements: I've added borderBright to [defaultTheme.js](https://github.com/jakeclements/react-dates/blob/8b9ebb5d2819a6cacd42bf8aa3df0d9bfd05dd73/src/theme/DefaultTheme.js#L12).
Do any other files require an update for this?

src/components/CalendarDay.jsx
```diff
@@ -140,6 +140,7 @@ class CalendarDay extends React.Component {
           styles.CalendarDay_container,
           isOutsideDay && styles.CalendarDay__outside,
           modifiers.has('today') && styles.CalendarDay__today,
+          modifiers.has('hovered') && styles.CalendarDay__hovered,
```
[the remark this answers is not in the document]
jakeclements: I agree that this should be decoupled to let pseudo selectors handle the majority of the hovers.
Lets update this one once we've had a think about naming conventions.

src/components/CalendarDay.jsx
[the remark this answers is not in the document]
majapw: doh! i just missed it. :)

jakeclements:
Hey @majapw, what is your opinion on the following;
In place of `range` we could pass `selectRangeOnClick`. I think this conveys what the prop does in a much clearer manner.
It would also allow us to bubble up functionality to DateRangePicker (hopefully) without causing confusion.

```jsx
<DateRangePicker
  {...props}
  selectRangeOnClick={{
    beforeSelected: fn,
    afterSelected: fn,
  }}
/>
```

In place of the `hovered` modifier I would propose something along the lines of `active-range`, or `hovered-range` if active is too closely related to `selected`.

circlingthesun:
Ping?

jakeclements:
@circlingthesun, I'd be keen to continue the discussion and development on this issue in the new year. If you have any suggestions as to how the props could be named/structured to suit your applications I'd be interested to hear it.

circlingthesun:
@jakeclements, I really like this PR. So I'd be keen to help things move along.

I was looking to replace a custom date picker widget when I saw this. I need to be able to select an entire week by clicking on any day of said week.

I'd agree with @majapw that `getRangeDay` is a bit confusing (more so than `range`). Perhaps use `getDateOffset`?

Instead of range `selectRangeOnClick` or `range`, what if we use two props called `startDateOffset` and `endDateOffset`. This way the API is more flat and you don't need to provide a NOOP function if you don't want to offset the start or end?

Thoughts?

jakeclements:
Thanks for jumping in @circlingthesun, I like this. 
`offset` as the core language of this functionality sits well with me in differentiating it from the normal use of date-range.
Following this we could also create `hovered-offset` as the modifier.

In terms of `selectRangeOnClick`, it's currently possible to pass a single start or end into the object to create one offset.
I don't have a strong opinion on if this would be better suited to be passed as single object vs two props, happy to follow @majapw and yourself on this.

majapw:
I'll take a look tonight or tomorrow. :) I just got back from vacay.

Offset seems much clearer to me as the core word to use. Maybe even
`getSelectedDateOffset`?

examples/DayPickerRangeControllerWrapper.jsx
```diff
@@ -31,6 +35,7 @@ const propTypes = forbidExtraProps({
   orientation: ScrollableOrientationShape,
   withPortal: PropTypes.bool,
   initialVisibleMonth: PropTypes.func,
+  range: RangeShape,
```
majapw: offsetRange maybe? are we still using this?

src/shapes/RangeShape.js
```diff
@@ -0,0 +1,14 @@
+import PropTypes from 'prop-types';
```
majapw: Can we delete this?

src/utils/getRangeDay.js
```diff
@@ -0,0 +1,6 @@
+const defaultModifier = day => day;
```
majapw: Do we use this?

test/components/DayPickerRangeController_spec.jsx
```diff
@@ -1476,6 +1476,54 @@ describe('DayPickerRangeController', () => {
         );
       });
     });
+
+    describe('props.range', () => {
```
majapw: range => startDateOffset/endDateOffset

jakeclements:
Thanks @majapw, I kinda left this half finished yesterday. Will wrap it up today.

examples/DayPickerRangeControllerWrapper.jsx
jakeclements: Replaced with startDateOffset/endDateOffset

jakeclements:
@majapw, this has been updated, squash & rebased. Please let me know if there are any other required changes.

src/components/DayPickerRangeController.jsx
```diff
@@ -170,6 +176,7 @@ export default class DayPickerRangeController extends React.Component {
       'last-in-range': day => this.isLastInRange(day),
       hovered: day => this.isHovered(day),
       'hovered-span': day => this.isInHoveredSpan(day),
+      'hovered-offset': day => this.isInHoveredSpan(day),
```
majapw: Do we use the same method for hovered-span and for hovered-offset? does that cause any issues?

src/components/DayPickerRangeController.jsx
```diff
@@ -467,51 +484,78 @@ export default class DayPickerRangeController extends React.Component {
...
 
-        if (!this.isBlocked(day) && isAfterDay(day, startDate)) {
-          const endSpan = day.clone().add(1, 'day');
-          modifiers = this.addModifierToRange(modifiers, startDate, endSpan, 'hovered-span');
+        dateOffset = {
+          start,
+          end,
+        };
+
+        if (this.state.dateOffset && this.state.dateOffset.start && this.state.dateOffset.end) {
```
majapw: what about the scenario where we only have dateOffset.start or dateOffset.end?

src/components/DayPickerRangeController.jsx
```diff
@@ -467,51 +484,78 @@ export default class DayPickerRangeController extends React.Component {
...
-      modifiers = this.addModifier(modifiers, day, 'hovered');
 
-      if (startDate && !endDate && focusedInput === END_DATE) {
-        if (isAfterDay(hoverDate, startDate)) {
-          const endSpan = hoverDate.clone().add(1, 'day');
-          modifiers = this.deleteModifierFromRange(modifiers, startDate, endSpan, 'hovered-span');
-        }
+      if (hasOffset) {
+        const start = getSelectedDateOffset(startDateOffset, day);
+        const end = getSelectedDateOffset(endDateOffset, day, rangeDay => rangeDay.add(1, 'day'));
```
majapw: Can you explain the need for the function here?

src/components/DayPickerRangeController.jsx
```diff
@@ -522,12 +566,16 @@ export default class DayPickerRangeController extends React.Component {
...
   onDayMouseLeave(day) {
     const { startDate, endDate, minimumNights } = this.props;
-    const { hoverDate, visibleDays } = this.state;
+    const { hoverDate, visibleDays, dateOffset } = this.state;
     if (this.isTouchDevice || !hoverDate) return;
 
     let modifiers = {};
     modifiers = this.deleteModifier(modifiers, hoverDate, 'hovered');
 
+    if (dateOffset) {
```
majapw: do we need the checks for dateOffset.start or dateOffset.end here?

majapw:
Okay, I checked it out and it looks really good! Thanks so much for all the hard work! :)

jakeclements:
Thank you @majapw and @ljharb for trawling through my code and taking the time to review!

jonricaurte:
@jakeclements is it possible to have the selection be week to week? Like a start week and an end week? So when I click a week it will show the start date be the first date of the first selected week. Then when I hover over other weeks after the first week it will fill in all the weeks in between the first selected week and the week I am hovering over. Then when I select a second week the end date will be the last day of the second selected week and all weeks in between the first selected week and the second selected week will be filled in. Thanks!

jonricaurte:
This would be using "with current week range selection" in the storybook.

jakeclements:
@jonricaurte this functionality unfortunately isn't covered by this PR. It should be possible based on the work completed here and some modification to the existing codebase. It's probably worth opening an issue to see if this is something the Airbnb team is interested in including.

jonricaurte:
@jakeclements Thanks! Opened the following issue to track this: 

https://github.com/airbnb/react-dates/issues/1035

PR #885, opened 2017-12-06
Use theme font size in a meaningful way

majapw:
Previously we set the font size in the theme and then did nothing about it. This worked in the storybook and anywhere else where the page font size was 14px, but if you changed that value, everything in the calendar would inherit from it which could be weird. This fixes the font size to a default value of 14px. If you want something different, you should register a theme value of `inherit`, or override appropriately.

### DIFF
--- a/src/components/DayPickerKeyboardShortcuts.jsx
+++ b/src/components/DayPickerKeyboardShortcuts.jsx
@@ -214,17 +214,14 @@ class DayPickerKeyboardShortcuts extends React.Component {
 
         {showKeyboardShortcutsPanel &&
           <div
-            {...css(
-              styles.DayPickerKeyboardShortcuts_panel,
-              block && styles.DayPickerKeyboardShortcuts_panel__block,
-            )}
+            {...css(styles.DayPickerKeyboardShortcuts_panel)}
             role="dialog"
-            aria-labelledby="DayPickerKeyboardShortcuts__title"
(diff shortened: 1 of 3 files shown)


src/components/DayPickerKeyboardShortcuts.jsx
```diff
@@ -214,17 +214,14 @@ class DayPickerKeyboardShortcuts extends React.Component {
...
+            {...css(styles.DayPickerKeyboardShortcuts_panel)}
             role="dialog"
-            aria-labelledby="DayPickerKeyboardShortcuts__title"
-            aria-describedby="DayPickerKeyboardShortcuts__description"
+            aria-labelledby="DayPickerKeyboardShortcuts_title"
+            aria-describedby="DayPickerKeyboardShortcuts_description"
           >
             <div
               {...css(styles.DayPickerKeyboardShortcuts_title)}
-              id="DayPickerKeyboardShortcuts__title"
```
ljharb: why change the BEM __ to a single _?

majapw: This is the pattern I've been using throughout and this was a place I'd missed it. If you look at the rest of the app, I've been using `__` for modifiers and `_` for elements. The title, in this case, is an element, not a modifier.

majapw:
@ljharb I would argue that this is a patch because inherited styles from anything other than 14px look super weird (and kind of broken) without this change.

PR #886, opened 2017-12-06
Add `inputHeight` prop

majapw:
This allows for complete customization of the inputHeight via prop. It also cleans up some logic that was introduced with the `verticalSpacing` prop (although I'm not counting the deletion of the `getInputHeight` method as breaking because that change had not yet been released). 

I was considering changing the theme props related to input and displaytext padding ... but I'm not sure that it is necessary? They do technically apply but I guess they are not as meaningful? I would take thoughts on that.

to: @gabergg @ljharb @erin-doyle @noratarano @airbnb/webinfra


src/components/DateInput.jsx
```diff
@@ -169,21 +171,15 @@ class DateInput extends React.Component {
...
     const value = displayValue || dateString || '';
     const screenReaderMessageId = `DateInput__screen-reader-message-${id}`;
 
     const withFang = showCaret && focused;
 
-    const {
-      font: { input: { lineHeight } },
-      spacing: { inputPadding, displayTextPaddingVertical },
-    } = reactDates;
-    const inputHeight = getInputHeight({ lineHeight, inputPadding, displayTextPaddingVertical });
```
ljharb: i'm confused; how can all this logic be replaced by just passing down a prop? I'm fine with it being overrideable, but wouldn't the default need this same logic?

majapw: Not really because once you set the height to a fixed value, this padding doesn't affect it anymore. Although I guess that the `displayTextPadding` might still increase the whole thing... it's not really being used anymore, so this might be an opportunity just to axe it (it's kind of a weird pattern that doesn't make terribly much sense now that we have a real input).

majapw: Nope, I'm definitely wrong. Let me reinvestigate!

majapw:
@ljharb so (1) I think I'm going to get rid of the `inputPadding` theme variable because it doesn't make sense to have both that and the `displayTextPadding` with the new styling/structure. I will probably rename `displayTextPadding` as `inputPadding` and apply it to the actual input component. So the structure of `DateInput` will look like so [highly simplified]:
```
<div {...css(..., { height: props.inputHeight })}>
  <input {...css(..., { padding: theme.spacing.inputPadding })} />
</div>
```

1) `inputHeight` is the one true way to set height, set `overflow: 'hidden'` on the wrapping div. You can still adjust the padding but it doesn't affect the final height. 

2) The `inputHeight` prop sets the `height` css variable but you can still affect things with the padding. This requires bringing back the `getInputHeight` util which does something like `max(height, 2*padding`. 

I kind of prefer 1... but idk if that's *oversimplifying*.

majapw:
Closing! Don't think this is a level of customization we are going to pursue.

PR #888, opened 2017-12-07
Bump react-portal to v4.1.0

tajo:
I removed 2 tests since `isOpened` prop doesn't exist anymore.

Closes #887.

### DIFF
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -2,7 +2,7 @@ import React from 'react';
 import shallowCompare from 'react-addons-shallow-compare';
 import moment from 'moment';
 import { css, withStyles, withStylesPropTypes } from 'react-with-styles';
-import Portal from 'react-portal';
+import { Portal } from 'react-portal';
 import { forbidExtraProps } from 'airbnb-prop-types';
 import { addEventListener } from 'consolidated-events';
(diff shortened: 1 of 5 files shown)


src/components/DateRangePicker.jsx
```diff
@@ -280,7 +280,7 @@ class DateRangePicker extends React.Component {
 
     if (withPortal || withFullScreenPortal) {
       return (
-        <Portal isOpened>
+        <Portal>
```
ljharb: can you elaborate on what the effective change will be without `isOpened`?

brucewpaul:
also closes #860

src/components/DateRangePicker.jsx
brucewpaul: in v4, the portal is rendered on cDM, so I don't think the isOpened is necessary. If react-dates needs this info and more control over it (which it doesn't seem like it does), then we could use PortalWithState. @tajo should definitely confirm this though.

tajo:
v3's `<Portal isOpened>` = v4's `<Portal>` .

I think it's more common to show/hide part of UI this way (also one less prop to document):

```js
bool && <Foo />
```

than

```js
<Foo active={bool}/>
```

There was no technical reason why not to use the first case, thus the change in v4.

PR #905, opened 2017-12-18
[New] Added phrase for aria-label for the selected day

erin-doyle:
Added the setting of the CalendarDay aria-label with a dateIsSelected phrase when the day is the selected day.

### DIFF
--- a/src/components/CalendarDay.jsx
+++ b/src/components/CalendarDay.jsx
@@ -8,9 +8,9 @@ import moment from 'moment';
 
 import { CalendarDayPhrases } from '../defaultPhrases';
 import getPhrasePropTypes from '../utils/getPhrasePropTypes';
-import getPhrase from '../utils/getPhrase';
+import getCalendarDaySettings from '../utils/getCalendarDaySettings';
 
-import { BLOCKED_MODIFIER, DAY_SIZE } from '../constants';
+import { DAY_SIZE } from '../constants';
 
 const propTypes = forbidExtraProps({
   ...withStylesPropTypes,
@@ -107,47 +107,19 @@ class CalendarDay extends React.Component {
       renderDayContents,
       tabIndex,
       styles,
-      phrases: {
-        chooseAvailableDate,
-        dateIsUnavailable,
-        dateIsSelected,
-      },
+      phrases,
     } = this.props;
 
     if (!day) return <td />;
 
-    const daySizeStyles = {
-      width: daySize,
-      height: daySize - 1,
-    };
-
-    const useDefaultCursor = (
-      modifiers.has('blocked-minimum-nights')
-      || modifiers.has('blocked-calendar')
-      || modifiers.has('blocked-out-of-range')
-    );
-
-    const selected = (
-      modifiers.has('selected')
-      || modifiers.has('selected-start')
-      || modifiers.has('selected-end')
-    );
-
-    const hoveredSpan = !selected && (
-      modifiers.has('hovered-span')
-      || modifiers.has('after-hovered-start')
-    );
-
-    const isOutsideRange = modifiers.has('blocked-out-of-range');
-
-    const formattedDate = { date: day.format(ariaLabelFormat) };
-
-    let ariaLabel = getPhrase(chooseAvailableDate, formattedDate);
-    if (modifiers.has(BLOCKED_MODIFIER)) {
-      ariaLabel = getPhrase(dateIsUnavailable, formattedDate);
-    } else if (selected) {
-      ariaLabel = getPhrase(dateIsSelected, formattedDate);
-    }
+    const {
+      daySizeStyles,
+      useDefaultCursor,
+      selected,
+      hoveredSpan,
+      isOutsideRange,
+      ariaLabel,
+    } = getCalendarDaySettings(day, ariaLabelFormat, daySize, modifiers, phrases);
 
     return (
       <td
--- a/src/components/CalendarDay.jsx
+++ b/src/components/CalendarDay.jsx
@@ -110,16 +110,11 @@ class CalendarDay extends React.Component {
       phrases: {
         chooseAvailableDate,
         dateIsUnavailable,
+        dateIsSelected,
       },
     } = this.props;
 
     if (!day) return <td />;
-
-    const formattedDate = { date: day.format(ariaLabelFormat) };
-
-    const ariaLabel = modifiers.has(BLOCKED_MODIFIER)
-      ? getPhrase(dateIsUnavailable, formattedDate)
-      : getPhrase(chooseAvailableDate, formattedDate);
 
     const daySizeStyles = {
       width: daySize,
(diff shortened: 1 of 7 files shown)


src/defaultPhrases.js
```diff
@@ -25,13 +25,11 @@ const returnFocusToInput = 'Return to the date input field.';
 const keyboardNavigationInstructions = `Press the down arrow key to interact with the calendar and
   select a date. Press the question mark key to get the keyboard shortcuts for changing dates.`;
 
-// eslint-disable-next-line camelcase
 const chooseAvailableStartDate = ({ date }) => `Choose ${date} as your check-in date. It's available.`;
-
-// eslint-disable-next-line camelcase
```
erin-doyle: I removed these disable comments as these lines aren't failing that rule anymore

src/components/CustomizableCalendarDay.jsx
```diff
@@ -207,6 +202,15 @@ class CustomizableCalendarDay extends React.Component {
 
     const isOutsideRange = modifiers.has('blocked-out-of-range');
 
+    const formattedDate = { date: day.format(ariaLabelFormat) };
+
+    let ariaLabel = getPhrase(chooseAvailableDate, formattedDate);
```
ljharb: Can you think of any clean way to share this logic with the identical code in CalendarDay?

test/components/CalendarDay_spec.jsx
```diff
@@ -75,6 +76,23 @@ describe('CalendarDay', () => {
         expect(wrapper.prop('aria-label')).to.equal('chooseAvailableDate text');
       });
 
+      it('is formatted with the dateIsSelected phrase function when day is selected', () => {
+        const selectedModifiers = new Set(['selected', 'selected-start', 'selected-end']);
+
+        selectedModifiers.forEach((selectedModifier) => {
+          const modifiers = new Set().add(selectedModifier);
```
ljharb: `new Set([selectedModifier])`?

test/components/CalendarDay_spec.jsx
```diff
@@ -75,6 +76,23 @@ describe('CalendarDay', () => {
...
+
+          const wrapper = shallow(<CalendarDay
```
ljharb: Not a huge deal, but the `(` should end a line here, and the `)` should begin once - the rest of the indentation will follow that. (You may need to double-wrap the jsx in parens to satisfy the linter)

test/components/CustomizableCalendarDay_spec.jsx
```diff
@@ -77,6 +78,23 @@ describe('CustomizableCalendarDay', () => {
         expect(wrapper.prop('aria-label')).to.equal('chooseAvailableDate text');
       });
 
+      it('is formatted with the dateIsSelected phrase function when day is selected', () => {
+        const selectedModifiers = new Set(['selected', 'selected-start', 'selected-end']);
+
+        selectedModifiers.forEach((selectedModifier) => {
+          const modifiers = new Set().add(selectedModifier);
+
+          const wrapper = shallow(<CustomizableCalendarDay
```
ljharb: also here

src/defaultPhrases.js
ljharb: Thanks!

test/components/CalendarDay_spec.jsx
erin-doyle: Yep that works fine and I can change it, the use of `new Set().add(<whatever>);` is used heavily in this file so I was following in kind.

test/components/CalendarDay_spec.jsx
erin-doyle: Same as above comment, this is the style that is used throughout this entire file so I was just going with the flow, but I can certainly change it up.

src/components/CustomizableCalendarDay.jsx
erin-doyle: There's a lot more than just this duplicated between CalendarDay and CustomizableCalendarDay now.  I could put together a utility function somewhere though and call it from each to get the ariaLabel value.

test/components/CalendarDay_spec.jsx
ljharb: fair, then it's fine to keep it :-)

test/components/CalendarDay_spec.jsx
ljharb: the other one is fine, but this one goes against (the spirit, at least, of) the airbnb styleguide, so we should clean this pattern up.

erin-doyle: I'll do a quick sweep and clean them up.

src/components/CalendarDay.jsx
```diff
@@ -107,43 +107,19 @@ class CalendarDay extends React.Component {
...
-
-    const isOutsideRange = modifiers.has('blocked-out-of-range');
+    const {
+      daySizeStyles,
+      useDefaultCursor,
+      selected,
+      hoveredSpan,
+      isOutsideRange,
+      ariaLabel,
+    } = getCalendarDaySettings(day, ariaLabelFormat, daySize, modifiers, phrases);
```
erin-doyle: Not sure if you'll hate this idea or not but it was my stab at consolidating some of the duplicate code between CalenderDay and CustomizableCalendarDay in the render() function.

ljharb: i don't love it, but it's better than the duplication! :-D

erin-doyle: Ha!  Agreed!  I feel like there must be a design pattern that will solve for all the duplication but it felt out of scope of this change.  I think this sort-of-icky utility function will do in the meantime. ;)

erin-doyle:
Rebase pushed.

erin-doyle:
I just want to give this one a bump before I rebase and force @ljharb to approve it again 😉 .

majapw:
Also leaving a reminder for myself that we need to translate the selected phrase for airbnb. :P

ljharb:
I think tiny well-tested utility methods is a good idea.

majapw:
That way both CalendarDay components could call the methods explicitly but the logic of what they did would be in one place?

ljharb:
Yep! Exactly. One source of truth means bugs are either less likely or more likely to be caught :-)

erin-doyle:
So are you saying you'd like something like:
```
getDaySizeStyles()
useDefaultCursorOnDay()
isDaySelected()
isDayHovered()
isDayOutsideRange()
getDayAriaLabel()
```

and would these each be their own file under utils?

I can go ahead and make those changes (in a separate PR preferably 😉).

erin-doyle:
I'm going to go ahead and rebase this now.

erin-doyle:
Ok @majapw sounds good and thanks!

erin-doyle:
Oh jeez, this still needs to be merged but now I need to rebase again. 😿 I'm going to do the rebase now and I'm sorry but I'm guessing it's going to require one of you approves again.  Once that's done I'll just merge it.

PR #910, opened 2017-12-20
Added "module" field in package.json for ESM

kamleshchandnani:
Added "module" field in package.json to support default import to be ESM for bundlers like webpack and rollup, which enables tree shaking out of the box.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,7 @@
   "version": "16.0.1",
   "description": "A responsive and accessible date range picker component built with React",
   "main": "index.js",
+  "module":"esm/index.js",
   "scripts": {
     "prebuild": "npm run clean",
     "build": "npm run build:cjs && npm run build:esm && npm run build:css -- --optimize ",


package.json
```diff
@@ -3,6 +3,7 @@
   "version": "16.0.1",
   "description": "A responsive and accessible date range picker component built with React",
   "main": "index.js",
+  "module":"esm/index.js",
```
ljharb: I'm a very very strong -1 for this change. This is a non-standard package.json field, and the fact that bundlers support it is highly toxic to the ecosystem.

Every time a package ships a "module" field, it makes it that much harder for a standard package.json field to ever exist.

kamleshchandnani: I agree! But tools like webpack and rollup still look up for "module". So IMO this will atleast reduce the bundle size for millions of consumer today, until the `.mjs` becomes standard and enabled in node by default?

ljharb: No amount of performance or bundle size is worth polluting the ecosystem, imo.

majapw:
I think unfortunately due to Jordan's point, we will be unable to accept this PR. Hopefully, if it becomes standard, then we will be able to make that change at that point. Thank you for contributing!

PR #916, opened 2017-12-22
[Fix] Keyboard nav issues

erin-doyle:
This is a combination of a few things:

- making the change requested in Issue #557 where we need the Enter key to be able to close the Keyboard Shortcuts Panel when open
- I found that the event.key value for the spacebar key is _not_ 'Space' at all 😳 but is either ' ' or 'Spacebar' on older browsers
- Added the throttle enhancements to the DayPicker that had been made to the DateInput in Pull Request #682 (fixing Issue #514)

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -26,4 +26,4 @@ export const FANG_WIDTH_PX = 20;
 export const FANG_HEIGHT_PX = 10;
 export const DEFAULT_VERTICAL_SPACING = 22;
 
-export const MODIFIER_KEY_NAMES = ['Shift', 'Control', 'Alt', 'Meta'];
+export const MODIFIER_KEY_NAMES = new Set(['Shift', 'Control', 'Alt', 'Meta']);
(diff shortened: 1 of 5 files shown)


src/constants.js
```diff
@@ -25,3 +25,5 @@ export const WEEKDAYS = [0, 1, 2, 3, 4, 5, 6];
 export const FANG_WIDTH_PX = 20;
 export const FANG_HEIGHT_PX = 10;
 export const DEFAULT_VERTICAL_SPACING = 22;
+
+export const MODIFIER_KEY_NAMES = ['Shift', 'Control', 'Alt', 'Meta'];
```
ljharb: Can this be a Set instead of an array?

erin-doyle: Yep I can do that.

erin-doyle:
I think this is a patch.  These are all just fixes, there isn't any breaking changes and nothing new really.  Now the Keyboard Shortcuts Panel will close using the Enter key where as before it was only Spacebar and Esc but that was really a miss, so I consider this fixing that not adding something new.

src/constants.js
erin-doyle: Done.  (was going to say "all set" 😝)

erin-doyle:
@majapw just want to bump this if you have a sec to take a look now that you're back from vacay. ;)

ljharb:
(it'll need a rebase also)

erin-doyle:
Pushed rebase. 👍

majapw:
Looking now! :)

Issue #904, opened 2017-12-18
openDirection="up" has fang in the wrong location


singingwolfboy:
This can be easily seen on the [official Storybook story for this case](http://airbnb.io/react-dates/?selectedKind=DRP%20-%20Calendar%20Props&selectedStory=open%20up):

![misplaced-fang](https://user-images.githubusercontent.com/132355/34110401-d955d5ae-e406-11e7-8325-3a5953f77271.png)

Notice how the top of the fang is aligned to the _input_, when it should be aligned to the _calendar_.

majapw:
Whoops! This is probably due to my recent refactoring. I can't get to a fix
until after the 25th, but the error is in the logic for fang placement in
the DateInput component. We'd accept a PR with the patch!

mhluska:
This tripped me up for a while. As a temporary fix, I'm passing `verticalSpacing={0}` to hide the caret.

majapw:
https://github.com/airbnb/react-dates/pull/925 should fix this!

singingwolfboy:
This appears to be fixed!

PR #925, opened 2017-12-28
Fix typo in DateInput.

codejunkienick:
This typo caused the wrong position of fang when using OPEN_UP constant for openDirection.


codejunkienick:
@ljharb we can test if svg element has 'bottom' or 'top' key in styles object.

ljharb:
Hmm, true, testing styles is probably the only way to cover it, and testing styles is generally not a good practice.

ljharb:
Let's get the tests passing, and then we can probably go forward as-is.

majapw:
Fixes https://github.com/airbnb/react-dates/issues/904

majapw:
Will merge this if travis ever feels better 🤒

PR #928, opened 2017-12-29
[Tests] include year in custom format string

ljharb:
Tests started failing on master as of 5 days before the end of the year.

Because the custom format string disregarded the year, it seems to break whenever "5 days from now" is in January.

I was able to get tests to pass by changing all of the "5"s to "1"s, which confirmed it's the year boundary causing the problem. I then reverted that and added the year into the custom format string, which fixed the issue.


ljharb:
Added some more cleanup and "improving coverage" commits as well.

ljharb:
Going to merge this prior to 2018 :-)

erin-doyle:
Nice catch and clean up (sorry I've been on vacation)!

PR #932, opened 2017-12-31
Requires startDateId, endDateId on DateRangePicker

iambenthompson:
Based on #866

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,7 +103,9 @@ Here is the minimum *REQUIRED* setup you need to get the `DateRangePicker` worki
 ```jsx
 <DateRangePicker
   startDate={this.state.startDate} // momentPropTypes.momentObj or null,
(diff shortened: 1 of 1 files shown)


erin-doyle:
It looks like we do indeed need something like this for SingleDatePicker.  The `id` prop is required but not shown in the example code.  It's also a little weird that we define a default value for `id` in SingleDatePicker even though it's required.

majapw:
@erin-doyle I did legitimately think we did that only for the DRP but it look like I was wrong ---https://github.com/airbnb/react-dates/blob/master/src/components/SingleDatePicker.jsx#L49

I guess maybe we need to go through this whole process for the SDP again.

majapw:
Since this is a README change, I'm going to merge through the PR failures (which have been addressed) and then add my tweaks directly on master.

PR #933, opened 2018-01-01
[Tests] fix more tests that fail when it’s January 1st

ljharb:
Found some more as a follow up to #928


majapw:
Thanks @ljharb! You're a hero. :)

Issue #939, opened 2018-01-03
Visually hidden previous/next months get read by screen readers


backwardok:
Steps to repro:
1. Open a date picker in Storybook (ex. http://airbnb.io/react-dates/?selectedKind=SingleDatePicker%20%28SDP%29&selectedStory=default&full=0&down=1&left=1&panelRight=0&downPanel=storybook%2Factions%2Factions-panel)
2. Turn on VoiceOver (or your screen reader of choice)
3. Navigate to the date picker input
4. Use arrow navigation (ctrl + opt + arrow in VoiceOver) to move to calendar and access dates

What's expected:
First date reached is the first date that is visible

What happens:
First date reached is the first date of the month before the visible month (so if this month is January, and that's the visible month, it goes to December).

Similarly, if you're on the last visible date, navigating beyond takes you to the next month (that's supposed to be invisible).

This could probably be solved by just setting `visibility:hidden` on the hidden months and then removing that styling when they're going to move over. Alternatively, you could use `aria-hidden` on the dates, but it's preferred to not use ARIA unless you have to.

majapw:
> This could probably be solved by just setting visibility:hidden on the hidden months and then removing that styling when they're going to move over. Alternatively, you could use aria-hidden on the dates, but it's preferred to not use ARIA unless you have to.

@backwardok can you explain why the discrepancy between the focused date and what's read is happening? I'm concerned that even if we apply `visibility: hidden` to the invisible months there will still be a problem in the following scenario:
![screen shot 2018-01-03 at 2 07 40 pm](https://user-images.githubusercontent.com/1383861/34542007-7975994a-f08f-11e7-8861-5f5af806f642.png)

where react-dates focused on the first available day if it exists. let me test that real quick.

majapw:
Okay, never mind, that does seem to work. Is it strange that our focus handling and the voiceover focus handling get out of sync?

backwardok:
If someone is navigating by traditional keyboard methods (tabbing around), then the focus would be on the "appropriate" date, though you'd still run into the situation where going to hidden dates would be possible. If you're navigating by screen reader browsing methods (going from node to node), then focus would end up on the inappropriate date. 

[...]

PR #940, opened 2018-01-03
Hide invisible first month using `visibility: hidden`

majapw:
Fix for #939 

FYI @erin-doyle @backwardok


majapw:
@backwardok yeah so this is in the `CalendarMonthGrid_month__hideForAnimation` styles which means it def gets removed when animating. I guess i should probably hide the last month too since this only applies to the first month.

majapw:
@backwardok @erin-doyle @ljharb can you take another look? 

The effect here is that both the first and last month (which are off-screen) are hidden from the nav. However, both have visibility during animation (otherwise things get weird). Thoughts?

backwardok:
Reasonable to me - I'll let someone else stamp

Issue #805, opened 2017-10-26
Cannot clear dates after manually changing date by highlighting text


sarahdmsi:
When manually entering a date, if you use backspace it seems to work fine. However, if you highlight and type to replace text, it seems to get stuck where you can no longer clear it. See the linked screenshot for an example.

https://www.dropbox.com/s/50hv1vk7v8kyvxa/React%20Dates%20Highlight%20Issue.mov?dl=0

majapw:
O_o that's really weird, it looks like it's not committing the value for some reason (ie not clearing the `dateString` state) but that's weird because it's closing the datepicker... which I would assume would only happen when it properly committed things. D: Maybe something got lost when we exposed the real input. 

This def needs more investigation. I can dig around a bit but would appreciate any help. :)

rakeshgunduka:
This bug isn't reproducible. I tried the way it was done in the [screen shot example given](https://www.dropbox.com/s/50hv1vk7v8kyvxa/React%20Dates%20Highlight%20Issue.mov?dl=0) by @sarahdmsi but ended with no success.

PR #943, opened 2018-01-07
Fix issue with clear dates after manually changing date

rokborf:
Issue:
https://github.com/airbnb/react-dates/issues/805

`dateString` should be kept only when there is no data in the `displayValue`, for edit purposes. Of course, If I understand it correctly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,7 +104,7 @@
     "rimraf": "^2.6.2",
     "safe-publish-latest": "^1.1.1",
     "sass-loader": "^6.0.6",
-    "sinon": "^4.1.3",
+    "sinon": "^4.1.4",
     "sinon-sandbox": "^1.0.2",
     "style-loader": "^0.19.0",
     "webpack": "^2.6.1"


src/components/DateInput.jsx
```diff
@@ -93,7 +93,7 @@ class DateInput extends React.Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    if (!this.props.displayValue && nextProps.displayValue) {
+    if (nextProps.displayValue) {
```
ljharb: The previous logic says “when a display value was not present, and you start passing one, clear the date string” - your new logic says “every time you render with a display value, even if it’s the same one as before, reset the date string”. I suspect this isn’t your intention?

rokborf: Actually, it is.
This is how you are defining `value` in the `render` method.
```javascript
```
So, basically, you are using `dateString` only if `displayValue` is empty or null.

The only thing that needs to be added to the condition is `!this.state.dateString` to prevent unnecessary `setState`.

So it will look like this:

```javascript
    if (!this.state.dateString && nextProps.displayValue) {
      this.setState({
        dateString: '',
      });
    }
```

src/components/DateInput.jsx
```diff
@@ -121,8 +121,7 @@ class DateInput extends React.Component {
     if (dateString[dateString.length - 1] === '?') {
       onKeyDownQuestionMark(e);
     } else {
-      this.setState({ dateString });
-      onChange(dateString);
+      this.setState({ dateString }, () => onChange(dateString));
```
ljharb: this is a good change regardless.

rokborf:
@ljharb 

### Why this happen:

**1. Changing date by editing without highlighting (works correctly)**
 - select date: `this.props.displayValue` = `e.target.value`, `this.state.dateString` = ''
 - deleting one number makes date invalid so `this.props.displayValue` becomes `null`, and `this.state.dateString` = `e.target.value`
 - After pressing number key date becomes valid. So `this.props.displayValue` = `null` && `nextProps.displayValue` = `e.target.value`
 - `componentWillReceiveProps` in `DateInput` sets `this.state.dateString` = ''
 - pressing clear button makes `this.props.displayValue` = null, `this.state.dateString` stays empty
 - in `render` `const value = displayValue || dateString || '';` value = '' 

 **2. Changing date by highlighting one number (works incorrectly)**
 - select date: `this.props.displayValue` = `e.target.value`, `this.state.dateString` = ''
 - highlighting one number and pressign number(**bug here**):
   - `onChange` in `DateInput` sets `this.state.dateString` = `e.target.value` and calls `this.props.onChange` that sets `this.props.displayValue` = `e.target.value`, but `this.state.dateString` stays uncleared, because condition in `componentWillReceiveProps` `!this.props.displayValue && nextProps.displayValue` === `false`
   - pressing clear button makes `nextProps.displayValue` = null, but `this.state.dateString` stil equals previous `value`
   - in `render` `const value = displayValue || dateString || '';` that makes value = dateString (previous value)

majapw:
Interesting. I'm going to check out this branch and play around with it. Thank you for the writeup!

majapw:
Huh, this does appear to fix this one specific bug. 

I'll write a regression test and merge this in.

I do think we'll need a more comprehensive solution for the date string soon.

PR #956, opened 2018-01-17
Add small property to docs

benrudolph:


majapw:
I feel like it would beneficial to automatically populate the README given how much I forget to add props. Maybe a postpublish script of some sort?

ljharb:
@majapw i recently wrote a PR that uses a tool like that; https://github.com/tc39/agendas/pull/298 for reference

majapw:
@ljharb cool! I'll dig into that :)

PR #957, opened 2018-01-17
Add regular option to DRP and SDP

majapw:
In a follow-up, I'd like to make regular the default and set light as an option (which makes more sense to me)

to: @ljharb @gabergg @noratarano @erin-doyle

### DIFF
--- a/src/components/DateInput.jsx
+++ b/src/components/DateInput.jsx
@@ -35,6 +35,7 @@ const propTypes = forbidExtraProps({
   showCaret: PropTypes.bool,
   verticalSpacing: nonNegativeInteger,
   small: PropTypes.bool,
+  regular: PropTypes.bool,
 
   onChange: PropTypes.func,
   onFocus: PropTypes.func,
@@ -60,6 +61,7 @@ const defaultProps = {
   showCaret: false,
   verticalSpacing: DEFAULT_VERTICAL_SPACING,
   small: false,
+  regular: false,
 
   onChange() {},
   onFocus() {},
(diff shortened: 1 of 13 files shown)


src/components/DateInput.jsx
```diff
@@ -288,7 +292,6 @@ export default withStyles(({
   },
 
   DateInput_input: {
-    fontWeight: 200,
```
ljharb: so this part is the breaking change?

majapw: Yes precisely

majapw:
@ljharb it was more that it seemed to me that a `regular` prop seems strange, whereas a light prop makes more sense when both are options (also matches our internal paradigms)

It's not strictly necessary to be breaking, it was just a preference

ljharb:
I'd rather add a `regular` prop now, ship that, and then in the next semver-major, remove `regular` and make it the default.

majapw:
@ljharb uh what?

ljharb:
@majapw iow, `light` would have the same behavior with true or false; `light` and `regular` would be mutually exclusive; and then later, we'd remove "regular", and make `light` function as expected.

majapw:
@ljharb That seems like the same amount of work as just switching them later on, it's just that the work is now instead of later.

ljharb:
It's the same amount of work for *us*, but it'll be less work for developers to migrate to the next semver-major, since those using `light` won't have to do anything, and those using `regular` will merely have to remove the prop.

Issue #747, opened 2017-10-06
SingleDatePicker focus problems on touch devices


comron:
Hello, 

Related to #560 -- Even after the recent change that allows `readonly` to be set and respected rather than default to `isTouchDevice`, things don't behave right on touch devices that also have keyboards, like a laptop or desktop with a touch screen.

Then problem seems to be that in `SingleDatePicker` there is focus logic that steals away the inputs focus and places focus on the `DayPicker` if isTouchDevice is true.

[https://github.com/airbnb/react-dates/blob/v13.0.0/src/components/SingleDatePicker.jsx#L182](https://github.com/airbnb/react-dates/blob/v13.0.0/src/components/SingleDatePicker.jsx#L182)

There is no way to short-circut this logic and even if you manage to set `readonly` correctly, you still are forced to use the `DayPicker` rather than the input on any touch device, even if you have a keyboard available.

Any ideas on how to proceed?

majapw:
Hmm, does hitting `esp` to go back to the input work? In general we want to encourage use of the calendar over typing in the inputs, so I'm not sure that this is the worse behavior. What would your preferred behavior be?

It's possible that we could remove this check when move focus to the daypicker and leave that for the portal case.

comron:
It doesn't look like there is anyway to get focus back to the input. The code seems to steal focus from the input whenever it gains focus if isTouchDevice is true.

I think the problem here is that "touch device" is too broad of a check, and specifically not allowing a user to type into the field if their browser responds to 'ontouchstart' seems overly strict. In this case the user doesn't even know they have a touch device, it just happens to be a PC  with a keyboard and mouse and a touch monitor. The user expects this form to work like any other web input, especially when typing specific far-future dates typing is the most intuitive interaction.

I think you're idea about only forcing the focus when presenting the DatePicker in a portal because of screen size is the better idea. At a minimum it would be nice to be in control of this behavior, even if the default doesn't change.

comron:
Was there any resolution for this? I would be happy to implement the change if one was proposed.

erin-doyle:
Maybe we need to work on making isTouchDevice more specific to mobile devices then?

ljharb:
We should be detecting features, not sniffing user agents - "mobile" isn't a feature, while "touch" is.

majapw:
@erin-doyle I think the problem is if we're not using `touch`, we usually end up using user-agent, which is unreliable. 

[...]

PR #958, opened 2018-01-17
Add keepFocusOnInput prop to maintain input focus even on touch devices

comron:
This adds a prop for SingleDatePicker and DateRangePicker. It defaults to false to maintain current behavior by default, but if true it will short circuit the logic that moves the focus to the DatePicker.

I'm not sure what level of testing the project requires, or if any changes to the storybook should be made. Let me know what else you'd like to see before merging.

Re: #747

### DIFF
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -65,6 +65,7 @@ const defaultProps = {
   block: false,
   small: false,
   regular: false,
+  keepFocusOnInput: false,
 
   // calendar presentation and interaction related props
   renderMonth: null,
@@ -188,10 +189,17 @@ class DateRangePicker extends React.Component {
   }
 
   onDateRangePickerInputFocus(focusedInput) {
-    const { onFocusChange, withPortal, withFullScreenPortal } = this.props;
+    const {
+      onFocusChange,
+      withPortal,
+      withFullScreenPortal,
+      keepFocusOnInput,
+    } = this.props;
 
     if (focusedInput) {
-      const moveFocusToDayPicker = withPortal || withFullScreenPortal || this.isTouchDevice;
+      const withAnyPortal = withPortal || withFullScreenPortal;
+      const moveFocusToDayPicker = withAnyPortal || (this.isTouchDevice && !keepFocusOnInput);
+
       if (moveFocusToDayPicker) {
         this.onDayPickerFocus();
       } else {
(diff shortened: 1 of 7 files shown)


src/components/DateRangePicker.jsx
```diff
@@ -187,10 +188,15 @@ class DateRangePicker extends React.Component {
...
+    const {
+      onFocusChange,
+      withPortal,
+      withFullScreenPortal,
+      keepFocusOnInput
+    } = this.props;
 
     if (focusedInput) {
-      const moveFocusToDayPicker = withPortal || withFullScreenPortal || this.isTouchDevice;
+      const moveFocusToDayPicker = !keepFocusOnInput && (withPortal || withFullScreenPortal || this.isTouchDevice);
```
majapw: One thing that's of note here is that means it's possible to keep focus on the input when using the portal options (with which the inputs are explicitly not visible). 

Maybe we want this to be `(withPortal || withFullScreenPortal || (!keepFocusOnInput && this.isTouchDevice))`? I'm not sure.

src/components/SingleDatePicker.jsx
```diff
@@ -192,9 +193,10 @@ class SingleDatePicker extends React.Component {
       onFocusChange,
       withPortal,
       withFullScreenPortal,
+      keepFocusOnInput
     } = this.props;
 
-    const moveFocusToDayPicker = withPortal || withFullScreenPortal || this.isTouchDevice;
+    const moveFocusToDayPicker = !keepFocusOnInput && (withPortal || withFullScreenPortal || this.isTouchDevice);
```
majapw: One thing that's of note here is that means it's possible to keep focus on the input when using the portal options (with which the inputs are explicitly not visible).

Maybe we want this to be (withPortal || withFullScreenPortal || (!keepFocusOnInput && this.isTouchDevice))? I'm not sure.

majapw:
Also it looks like we're gonna have to rebase! :) As for tests, it'd be good to add a quick mocha test for this behavior. I'm not sure that we need a story to demo the behavior but that's always nice.

src/components/DateRangePicker.jsx
```diff
@@ -188,10 +189,17 @@ class DateRangePicker extends React.Component {
...
+      onFocusChange,
+      withPortal,
+      withFullScreenPortal,
+      keepFocusOnInput,
+    } = this.props;
 
     if (focusedInput) {
-      const moveFocusToDayPicker = withPortal || withFullScreenPortal || this.isTouchDevice;
+      const withAnyPortal = withPortal || withFullScreenPortal;
+      const moveFocusToDayPicker = withAnyPortal || (this.isTouchDevice && !keepFocusOnInput);
```
majapw: 👍

PR #961, opened 2018-01-18
[Fixes #946] Move focus to datepicker when SingleDatePicker is set to readOnly

### LINKED ISSUE #946: SingleDatePicker onFocus when readOnly is true
This is related to the same code in #747 but a separate issue.
In the SingleDatePicker onFocus function, there is logic to determine whether focus should remain on the input when opened (so user can type in date) or move to the datepicker (for portal and touch device.)
When I have a readOnly input, I want to be able to capture subsequent interactions with the input so I can close the date picker. I can determine this in the onFocusChange function as long as the focus moves from the input to the date picker.
I am proposing that the onFocus event moves focus to the date picker in the case where readOnly is true.

aporlando:
Issue #946 proposes that when SingleDatePicker is set to readOnly, then its onFocus behavior should move the focus to the datepicker from the input (since there's no opportunity to type in a readOnly input).
This helps accomplish behavior to hook into subsequent clicks on a readOnly input to close the datepicker (on non-touch enabled devices).
Note that this is a separate issue from #958 which provides a hook to keep focus on the input.

### DIFF
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -204,6 +204,7 @@ class DateRangePicker extends React.Component {
   onDateRangePickerInputFocus(focusedInput) {
     const {
       onFocusChange,
+      readOnly,
       withPortal,
       withFullScreenPortal,
       keepFocusOnInput,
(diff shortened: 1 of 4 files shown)


majapw:
Does this actually end up closing the datepicker when you click on the input again?

aporlando:
No. It allows you to accomplish it in the onFocusChange event for readOnly inputs. Something like:
```
<SingleDatePicker readOnly
focused={this.stat.focused}
onFocusChange={({ focused }) => {
if (this.state.focused && focused) {
    // Already focused, so datepicker is open, focus event again means subsequent input interaction
    this.setState({ focused: false });
}
else this.setState({focused})
}}
/>```

PR #962, opened 2018-01-19
Quick fixes for `CustomizableCalendarDay`

majapw:
Whoops! Somehow I missed the `selected` styles the first time around.

Also, whenever `hasCustomStyles` ended up accidentally `null`, everything would break which was a downer.

@ljharb can you take a quick look?


majapw:
All styles-based unfortunately :/

ljharb:
@majapw sure but simply exercising this code would have thrown an error, no?

PR #979, opened 2018-01-25
Fix specificity issues with selected styles on the CustomizableCalendarDay

majapw:
This also adds a more thorough example for customization (with blocked dates and visible minimum nights).

It also fixes an issue where selected-start and selected-end styles don't supercede selected styles (as they should).

to: @ljharb @erin-doyle @gabergg @noratarano


gabergg:
LGTM!

PR #984, opened 2018-01-27
Add dayAriaLabelFormat into DatePickerRange and SingleDatePicker

koba04:
This PR is a follow-up for https://github.com/airbnb/react-dates/pull/857#issuecomment-359880717

I've add `dayAriaLabelFormat` into `DatePickerRange` and `SingleDatePicker` and have mentioned in README.

### DIFF
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -338,6 +338,7 @@ class DateRangePicker extends React.Component {
       customCloseIcon,
       onClose,
       phrases,
+      dayAriaLabelFormat,
       isRTL,
       weekDayFormat,
       styles,
@@ -417,6 +418,7 @@ class DateRangePicker extends React.Component {
           showKeyboardShortcuts={showKeyboardShortcuts}
           onBlur={this.onDayPickerBlur}
           phrases={phrases}
+          dayAriaLabelFormat={dayAriaLabelFormat}
           isRTL={isRTL}
           firstDayOfWeek={firstDayOfWeek}
           weekDayFormat={weekDayFormat}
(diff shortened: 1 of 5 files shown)


src/components/DateRangePicker.jsx
```diff
@@ -335,6 +335,7 @@ class DateRangePicker extends React.Component {
       customCloseIcon,
       onClose,
       phrases,
+      dayAriaLabelFormat,
```
majapw: We should have a defaultProp value for this one

src/components/SingleDatePicker.jsx
```diff
@@ -364,6 +364,7 @@ class SingleDatePicker extends React.Component {
       firstDayOfWeek,
       customCloseIcon,
       phrases,
+      dayAriaLabelFormat,
```
majapw: We should have a defaultProp value for this one

koba04:
@majapw 
I've added defaultProp for `dayAriaLabelFormat`.
But I didn't add defaultProp for `dayAriaLabelFormat` to components I've added `dayAriaLabelFormat` because `dayAriaLabelFormat` is passed as a `ariaLabelFormat` prop of `CalendarDay`.
`ariaLabelFormat` has a defaultProp(`dddd, LL`) so I thought defaultProp for `dayAriaLabelFormat` isn't necessary.

Should I add defaultProp for `dayAriaLabelFormat` in other components(e.g. `CalendarMonth`, `CalendarMonthGrid`, `DayPicker`, `DayPickerRangeController` and `DayPickerSingleDateController`)?

majapw:
@koba04 it would be good to be explicit, so in the case of CalendarMonth, CalendarMonthGrid, DayPicker, DayPickerRangeController and DayPickerSingleDateController... we don't need a defaultProp value of `"dddd, LL"`, but I would still put in a defaultProp value of `undefined`.

koba04:
CI was failed due to #995

koba04:
Thanks!

Issue #840, opened 2017-11-14
Configure location of Info Panel?


Drecula:
Hi,

Might we add a config so the Info Panel could be Top, Right, Bottom, or Left vs always at the bottom?

Thanks!

majapw:
Yes! I meant to respond to this when this was first opened, but this seems like a great idea.

varun1101:
I intend to start on this. Just wanted to know should this appear as a separate filter in Storybook or as an manual option similar to animation etc.

majapw:
Hi @varun1101, 

I think ideally that the DateRangePicker and SingleDatePicker would take a `calendarInfoPosition` prop and that it would behave very similarly to the `inputIconPosition` prop. It would be good to add like `INFO_POSITION_TOP`, `INFO_POSITION_BOTTOM`, `INFO_POSITION_LEFT`, and `INFO_POSITION_RIGHT` to the constants file and use those as enums. 

I would just add examples to https://github.com/airbnb/react-dates/blob/master/stories/DateRangePicker_calendar.js and https://github.com/airbnb/react-dates/blob/master/stories/SingleDatePicker_calendar.js in much the same way as the 'with info panel' story.

varun1101:
Thanks @majapw for the help.

mariepw:
Hi, 
Any updates on this? 
@varun1101 did you work on something? Or you want me to do the PR? 
Thanks.

varun1101:
Hi,
I did not get any time in particular to work on this. You can go ahead with it @mariepw . Thanks

Airboy:
Hi @mariepw,

Would be nice to update the documentation about the use of this feature ;-)
`calendarInfoPosition` is missing within the README file. Let me know if I can help with that.
Thanks.

mariepw:
@Airboy you're right, thanks. I'll try to make a PR to add it during the week-end :) Or you can do it if you want to ..

PR #989, opened 2018-01-30
Custom Info Panel position

mariepw:
Hi @majapw, Hi @ljharb, Hi all,
I implemented the changes proposed in #840 by adding prop 'calendarInfoPosition' to 'SingleDatePicker' and 'DateRangePicker'.
I also directly updated the stories, to test thoses changes.
Thanks.


src/components/DayPicker.jsx
```diff
@@ -753,6 +751,46 @@ class DayPicker extends React.Component {
...
+    const calendarInfoPositionLeft = calendarInfoPosition === INFO_POSITION_LEFT;
+    const calendarInfoPositionRight = calendarInfoPosition === INFO_POSITION_RIGHT;
+    const calendarInfoIsInline = calendarInfoPositionLeft || calendarInfoPositionRight;
+
+    const calendarInfo = renderCalendarInfo && (
+      <div
+        ref={this.setCalendarInfoRef}
+        {...css((calendarInfoIsInline) && styles.DayPicker_calendarInfo__horizontal)}
+      >
+        { renderCalendarInfo() }
```
ljharb: nit: jsx curly braces shouldn't have padding spaces inside them

src/components/DayPicker.jsx
```diff
@@ -753,6 +751,46 @@ class DayPicker extends React.Component {
...
+      width: this.isHorizontal() && wrapperHorizontalWidth,
+      height,
+    };
+
+    const dayPickerWrapperStyle = {
+      width: wrapperHorizontalWidth,
+    };
+
+    const dayPickerStyle = {
+      width: this.isHorizontal() && fullHorizontalWidth,
```
ljharb: can we cache the `isHorizontal` result in a var, so we don't have to call the function multiple times?

src/components/DayPicker.jsx
```diff
@@ -753,6 +751,46 @@ class DayPicker extends React.Component {
...
+
+      // These values are to center the datepicker (approximately) on the page
+      marginLeft: this.isHorizontal() && withPortal && -fullHorizontalWidth / 2,
+      marginTop: this.isHorizontal() && withPortal && -calendarMonthWidth / 2,
```
ljharb: rather than overloading `&&` for value selection, can these use explicit ternaries?

src/components/DayPicker.jsx
mariepw: Done

src/components/DayPicker.jsx
mariepw: Done.. but the function is called at other different places in the code... Will it be better to cache the result in a state?

src/components/DayPicker.jsx
mariepw: I changed it, but for the record, the implicit ternaries were already here before, I only change their position in the code..

src/components/DayPicker.jsx
ljharb: nah, that would cause a rerender. It's fine to call multiple times conceptually, but it's nice to minimize duplicate function calls within the body of a single function :-)

src/components/DayPicker.jsx
ljharb: Fair enough :-) thanks for leaving it cleaner than you found it!

mariepw: My pleasure ! :)

src/components/DayPicker.jsx
mariepw: So, it is okay like that ?

mariepw:
Hi @majapw , 

1 - Totally right, sorry for the "glitch", I changed the default info position.
2 - Still under investigation, but I am pretty sure it is because of the change of `display` on `wrapper` and `infoPanel` when they are inline.

mariepw:
I just found out where the issue is coming from... It is due to the use of `calculateDimension`.. Will find a way to fix this..

mariepw:
I'm still stuck with this problem... 

When I change : 
`     const calendarInfoWith = renderCalendarInfo && calendarInfoIsInline
      ? calculateDimension(this.calendarInfo, 'width', true, true)
       : 0; `
to : 
`     const calendarInfoWith = renderCalendarInfo && calendarInfoIsInline
      ? 300 // hardcoded random value
       : 0;
`
transition works well, and calendar info panel will always have the same width.. But I cannot just hardcode a value like this...

- I tried to get the width when `componentDidMount` and add it to a state, transition then works, but the width is not set properly... (`render` is called several times after `componentDidMount`).

- Then I tried : 
        - to recalculate the width in `componentWillUpdate`, and set the state again...
        - change the display of `DayPicker_calendarInfo__horizontal` and `DayPicker_wrapper__horizontal` from `table-cell` to `inline-block`
        - add `1px` to `fullHorizontalWidth` (to compensate white-space between the two `inline-block`)
This solution fix the problem of width for calendar info panel, but the transition is still buggy..

- I even tried to use `this.calendarInfo.getBoundingClientRect().width` instead of `calculateDimension(this.calendarInfo, 'width', true, true)`.. But it didn't change nothing.. 

So... Do you have any idea @majapw @ljharb  ?
Thanks a lot!

mariepw:
I finally fixed the issue..
Calculating the dimensions of calendar info panel, was creating a DOM repaint, this is what was breaking the CSS transition.. Adding a setTimeout to delay calculation, to wait until transition ends, fixed it. 
Everything works now, calendar info panel renders itself with expected width, and transition works. 
Let me know if it is ok.

src/components/DayPicker.jsx
```diff
@@ -242,6 +260,23 @@ class DayPicker extends React.Component {
...
+    if (this.calendarInfo) {
+      const { calendarInfoWidth } = this.state;
+      this.timeout = setTimeout(() => {
+        const calendarInfoPanelWidth = calculateDimension(this.calendarInfo, 'width', true, true);
+        if (calendarInfoWidth !== calendarInfoPanelWidth) {
+          this.setState({
+            calendarInfoWidth: calendarInfoPanelWidth,
+          });
+        }
+      }, 200);
```
majapw: Should we use `transitionDuration` here instead?

src/components/DayPicker.jsx
```diff
@@ -242,6 +260,23 @@ class DayPicker extends React.Component {
     return shallowCompare(this, nextProps, nextState);
   }
 
+  componentWillUpdate() {
+    // Calculating the dimensions trigger a DOM repaint which
+    // breaks the CSS transition.
+    // The setTimeout will wait until the transition ends.
+    if (this.calendarInfo) {
+      const { calendarInfoWidth } = this.state;
+      this.timeout = setTimeout(() => {
```
majapw: Can we name this `this.setCalendarInfoWidthTimeout` or something else meaningful

src/components/DayPicker.jsx
mariepw: Makes sense :) I didn't see the props...

mariepw:
Thanks a lot @majapw !!! Last changes are done now :)
I cannot take all the credits and have to say thanks to @onigoetz, because he gave me the lead about the DOM repaint... 
Anyway, yay it's done! Thanks :)

mariepw:
Hi @majapw do you know when when it will be possible to merge ?

majapw:
Sorry @mariepw, it slipped off my radar! :) Let me do a release today.

mariepw:
Yeeaah for the release, perfect !! Thanks a lot @majapw !

PR #990, opened 2018-01-30
Ensuring callbacks will execute after new state is ready.

iliran11:
`onNextMonthClick` and `onPrevMonthClick` callbacks were seemingly executing right after updating the state of `DayPickerSingleDateController.jsx`.

However, It is not ensured and might cause some unexpected behavior caused by callbacks executing before the state update was completed (happened in my work project).

it is also a documented pitfall on [react documentations](https://reactjs.org/docs/react-component.html#setstate).

> setState() does not always immediately update the component. It may batch or defer the update until later. This makes reading this.state right after calling setState() a potential pitfall.

### DIFF
--- a/src/components/DayPickerRangeController.jsx
+++ b/src/components/DayPickerRangeController.jsx
@@ -648,7 +648,6 @@ export default class DayPickerRangeController extends React.Component {
     }, () => {
       onNextMonthClick(newCurrentMonth.clone());
     });
-
   }
 
   onMultiplyScrollableMonths() {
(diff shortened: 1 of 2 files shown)


iliran11:
not that i'm aware of. i think it's the minimal required fix for now :)

majapw:
@iliran11 can you make this change in the `DayPickerRangeController` as well? Otherwise, looks good to me!

src/components/DayPickerRangeController.jsx
```diff
@@ -645,9 +645,10 @@ export default class DayPickerRangeController extends React.Component {
         ...newVisibleDays,
         ...this.getModifiers(nextMonthVisibleDays),
       },
+    }, () => {
+      onNextMonthClick(newCurrentMonth.clone());
     });
```
majapw: I think we need to delete this blank line

iliran11:
@majapw  - done! ;)

majapw:
Sweet, thanks!

PR #994, opened 2018-02-01
The min-nights modifiers should get cleared according to the previous prop value not the new one

majapw:
In addition, we don't seem to be removing the `blocked` modifier from these days at all. Whoops!

Here is a gif of the bug in action (not this only happens when you change the min nights value from a bigger value to a smaller value). 
min_nights

Now we will remove min nights blocking according to the previous value and also, any time the value changes (and reapply it for new >0 values). 

to: @amhunt @ljharb @erin-doyle

### DIFF
--- a/test/components/DayPickerRangeController_spec.jsx
+++ b/test/components/DayPickerRangeController_spec.jsx
@@ -696,6 +696,31 @@ describe('DayPickerRangeController', () => {
               expect(isSameDay(minimumNightsCalls[0].args[2], minimumNightsEndSpan)).to.equal(true);
             });
           });
+
+          describe('minimumNights changed', () => {
+            it('calls deleteModifierFromRange with start date + old min nights, and `blocked-minimum-nights`', () => {
+              const deleteModifierFromRangeSpy = sinon.spy(DayPickerRangeController.prototype, 'deleteModifierFromRange');
+              const startDate = today;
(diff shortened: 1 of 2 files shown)


erin-doyle:
👍 on a little more detail and test(s).

test/components/DayPickerRangeController_spec.jsx
```diff
@@ -696,6 +696,31 @@ describe('DayPickerRangeController', () => {
...
             });
           });
+
+          describe('minimumNights changed', () => {
+            it('calls deleteModifierFromRange with start date + old min nights, and `blocked-minimum-nights`', () => {
+              const deleteModifierFromRangeSpy = sinon.spy(DayPickerRangeController.prototype, 'deleteModifierFromRange');
+              const startDate = today;
+              const focusedInput = START_DATE;
+              const minimumNights = 5;
+              const wrapper = shallow(<DayPickerRangeController
```
ljharb: fwiw, the linter will soon require:
```jsx
const wrapper = shallow((
  <DayPickerRangeController
    startDate={startDate}
    focusedInput={focusedInput}
    minimumNights={minimumNights}
  />
));
```

majapw: we'll have to do a wholesale change of... most tests at that point. 😬

PR #995, opened 2018-02-03
Fix failed tests on Saturday

koba04:
Currently, some tests are failed on Saturday due to lack of stub functions for isLastDayOfWeek :)


majapw:
@koba04 we should probably stub `isFirstDayOfWeek` as well. I'll do that separately.

koba04:
Thanks!

PR #1004, opened 2018-02-04
Fix failed tests on Sunday

koba04:
@majapw You are right😄.  https://github.com/airbnb/react-dates/pull/995#issuecomment-362829942


koba04:
Thanks!

PR #1012, opened 2018-02-08
Changed OrientationShape to ScrollableOrientationShape in DateRangePickerShape

valeryq:


majapw:
Hi @valeryq there is not currently any support for the `VERTICAL_SCROLLABLE` on the DRP which is why this option is not included. 

In fact this is a screenshot of the DRP with orientation set to `VERTICAL_SCROLLABLE`:
screen shot 2018-02-08 at 10 35 15 am

Our currently recommendation for using that option is to apply it to the DayPickerRangeController and hook up your own inputs if needed. We may at some point in the future roll our implementation back up in the DRP but that hasn't happened yet.

valeryq:
@majapw Hi, okey, I get it. Thank you.

Issue #797, opened 2017-10-25
Uncaught TypeError: Cannot read property 'theme' of undefined


axe-z:
ThemedStyleSheet.js?17e6:33 Uncaught TypeError: Cannot read property 'theme' of undefined
    at Object.create (ThemedStyleSheet.js?17e6:33)
    at withStyles (withStyles.js?55c9:78)
    at Object.65 (DateInput.js?edf4:462)
    at __webpack_require__ (DateInput.js?edf4:21)
    at eval (DateInput.js?edf4:67)
    at eval (DateInput.js?edf4:70)
    at Object.<anonymous> (bundle.js:2838)
    at __webpack_require__ (bundle.js:20)
    at Object.34 (DateRangePickerInput.js?2b8e:333)
    at __webpack_require__ (DateRangePickerInput.js?2b8e:21)

JUST ON IMPORT 
import React from "react";
import moment from 'moment';

import { SingleDatePicker } from 'react-dates';
import 'react-dates/lib/css/_datepicker.css';

NOT SURE WHAT I'M DOING WRONG HERE... IT'S A FIRST USE.

majapw:
Hi @axe-z, if you look at the readme, as of v13, there is one additional step to get things working. Specifically, you need to `import 'react-dates/initialize';` before you use any of the components.

See https://github.com/airbnb/react-dates#initialize for more info!

PR #1020, opened 2018-02-13
update Initialize wording in README

jamesmosier:
I just starting using `react-dates` for the first time. I read over the first half of the README and quickly ran into the [issue described here](https://github.com/airbnb/react-dates/issues/797) related to `Cannot read property 'theme' of undefined`.

The reason for this was because after reading the "Initialize" section of the README, I did not feel it applied to my use case when setting up `react-dates`. The way I initially read it was: "If you are using CSS/classes to override our default theme, then you'll have to `react-dates/initialize` to do this" ... the way I looked it at, I just wanted to accept whatever the defaults were, no matter *how* the styles were accomplished. So I didn't include the import and it didn't work 🤷‍♂️ 

Thanks for the great project 🎉


README.md
```diff
@@ -45,7 +45,7 @@ Ensure packages are installed with correct version numbers by running:
 import 'react-dates/initialize';
 ```
 
-As of v13.0.0 of `react-dates`, this project relies on `react-with-styles`. If you want to continue using CSS stylesheets and classes, there is a little bit of extra set-up required to get things going. As such, you need to import `react-dates/initialize` to set up class names on our components. This import should go at the top of your application as you won't be able to import any `react-dates` components without it.
+As of v13.0.0, `react-dates` relies on [`react-with-styles`](https://github.com/airbnb/react-with-styles). Therefore, to get started, you need to import `react-dates/initialize` to set up class names on our components. This import should go at the top of your application as you won't be able to import any `react-dates` components without it.
```
ljharb: This advice isn’t the way to get started, it’s the way to get started *without react-with-styles* - using react-with-styles is the recommended approach.

majapw: I don't know if that's a fair assessment @ljharb. The `initialize` script uses `react-with-styles` as well and it *is* the recommended out of the box approach.

jamez14: Is there another way to organize the `Initialize` section of the README then? More of a "do it this way with `react-with-styles`, otherwise you must do it this other way". I'm not familiar with `react-with-styles`, other than jumping over to that repo to briefly check out its purpose, so I'd imagine many others are not using it out of the gate either.

PR #1021, opened 2018-02-14
remove react-addons-shallow-compare package and use from React.PureComponent

alirezavalizade:


ljharb:
This is a duplicate of #313, #565, and #851.

It would be a breaking change, and thus we can’t do it until we’ve dropped React 15 entirely (since < 15.3 iirc) doesn’t support PureComponent. That can’t happen until *at least* React 18 is out, and probably not even then.

PR #1026, opened 2018-02-15
update persian example - use jalali date in inputs - change direction to RTL

alirezavalizade:
screen shot 2018-02-15 at 10 45 50 am

### DIFF
--- a/examples/DateRangePickerWrapper.jsx
+++ b/examples/DateRangePickerWrapper.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 import PropTypes from 'prop-types';
 import momentPropTypes from 'react-moment-proptypes';
 import moment from 'moment';
+import momentJalaali from 'moment-jalaali';
 import omit from 'lodash/omit';
 
 import DateRangePicker from '../src/components/DateRangePicker';
@@ -15,6 +16,7 @@ const propTypes = {
   // example props for the demo
   autoFocus: PropTypes.bool,
   autoFocusEndDate: PropTypes.bool,
+  isPersianDateRangePicker: PropTypes.bool,
   initialStartDate: momentPropTypes.momentObj,
   initialEndDate: momentPropTypes.momentObj,
 
@@ -31,6 +33,7 @@ const defaultProps = {
   // example props for the demo
   autoFocus: false,
   autoFocusEndDate: false,
+  isPersianDateRangePicker: false,
   initialStartDate: null,
   initialEndDate: null,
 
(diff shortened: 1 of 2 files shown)


stories/DateRangePicker.js
```diff
@@ -84,8 +84,13 @@ storiesOf('DateRangePicker (DRP)', module)
   })
   .addWithInfo('non-english locale (Persian)', () => {
     moment.locale('fa');
+    momentJalaali.loadPersian({ dialect: 'persian-modern', usePersianDigits: true });
```
ljharb: Can you confirm whether this impacts moment usage at all?

examples/DateRangePickerWrapper.jsx
```diff
@@ -108,7 +111,14 @@ class DateRangePickerWrapper extends React.Component {
   }
 
   onDatesChange({ startDate, endDate }) {
-    this.setState({ startDate, endDate });
+    if (this.props.isPersianDateRangePicker) {
```
ljharb: Not a big deal, but maybe instead of this prop, momentJalaali could be passed in as a `stateDateWrapper` prop or something?

stories/DateRangePicker.js
alirezavalizade: It just impact on Persian example.

examples/DateRangePickerWrapper.jsx
alirezavalizade: So, You mean something like this? `stateDateWrapper={momentJalaali}`?
IDK, from my opinion it was cleaner way. What's the exact problem? In the naming of `isPersianDateRangePicker` prop?

ljharb: Persian might not be the only one this is needed for in the future; it seems like a more scalable approach.

Since it’s an example, it doesn’t matter much tho :-)

stories/DateRangePicker.js
ljharb: What i mean is, does this call affect moment, or only momentJalaai?

alirezavalizade: No, just momentJalaai.

examples/DateRangePickerWrapper.jsx
alirezavalizade: what's your suggestion exactly?

ljharb:
```jsx
const { stateDateWrapper } = this.props;
this.setState({ startDate: stateDateWrapper(startDate), endDate: stateDateWrapper(endDate) });
```

and then `stateDateWrapper` would default to `x => x`.

alirezavalizade: Ok Thanks, 
but this line would be,
`this.setState({
      startDate: startDate && stateDateWrapper(startDate),
      endDate: endDate && stateDateWrapper(endDate),
  });`
because when we pass momentJalaali in as stateDateWrapper prop and each one of values was equal with null or undefined, we get error because is an invalid date.

alirezavalizade:
No i think it is not until change `displayFormat: () => moment.localeData().longDateFormat('L')` in `defaultProps`.
It's affect on inputs and content of calendar, It was mix of English and Persian and i change it to Persian! (Esfand => اسفند)
screen shot 2018-02-20 at 6 05 39 pm

alirezavalizade:
Ping :|

Issue #1025, opened 2018-02-15
Disable hovered style on minimum nights when hovering selected check in date


angusfretwell:
Is there a way to disable the behaviour where the minimum number of nights are styled as hovered when hovering a selected check in date?

PR #1052, opened 2018-02-26
Add after hovered start span class name

angusfretwell:
Hey, I know this is a bit half baked but wanted to check if you'd be interested in merging it before I update the docs, etc.? As mentioned in #1025, we don't want the minimum nights to have the hovered style when hovering the selected start date. Adding this class name allows us to style those days the same as when the start date is not hovered.

### DIFF
--- a/test/utils/getCalendarDaySettings_spec.js
+++ b/test/utils/getCalendarDaySettings_spec.js
@@ -117,7 +117,7 @@ describe('getCalendarDaySettings', () => {
     });
 
     it('should be true when day is not selected and after-hovered-start', () => {
-      const modifiers = new Set(['hovered-span']);
+      const modifiers = new Set(['after-hovered-start']);
       const { hoveredSpan } = getCalendarDaySettings(
         testDay,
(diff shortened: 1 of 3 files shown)


test/utils/getCalendarDaySettings_spec.js
```diff
@@ -117,7 +117,7 @@ describe('getCalendarDaySettings', () => {
     });
 
     it('should be true when day is not selected and after-hovered-start', () => {
-      const modifiers = new Set(['hovered-span']);
+      const modifiers = new Set(['after-hovered-start']);
```
angusfretwell: Also noticed this test didn't match the description.

majapw:
@angusfretwell Sure! This seems fine to accept (esp since it's such a simple fix). The idea is just to have the `after-hover-start` class be explicit instead of bundled into `hovered-span`, right?

(also sorry for the delay in review :x)

ljharb:
@angusfretwell this was awaiting your response, but now you've deleted your fork so it can't ever be merged :-(

PR #1069, opened 2018-03-12
Add appendToBody and disableScroll props

### LINKED ISSUE #273: DateRangePicker cannot render correctly inside a scrolling container
Because `DateRangePicker__picker` is rendered adjacent to `DateRangePickerInput`, it cannot work correctly inside containers that scroll their overflow - it will either push the edges of the container outwards, or be clipped if you have something like `overflow-y: scroll; overflow-x: hidden;`. I believe I opened a ticket similar to this about five months ago, and it was "fixed" in #83.

@majapw Any update on this? I can provide an example if needed but it's quite simple to reproduce.

joaovieira:
**Fixes #273**

This adds the ability to have the day picker render outside of the scrolling container instead of being clipped.

It uses the same Portal as the portal variations, but positions the picker next to the inputs via hardware-accelerated CSS `translate` (keeping the `top`/`bottom` for relative positioning).

On top of that, I've decided to disable the scroll in this situation to avoid the picker being displayed when the input goes "out-of-bounds"/overflows the parent container, in case we scroll the picker with the input (as Tether/Popper does).

I've also decided to add a `disableScroll` with the same behaviour for non-detached pickers.

If you agree with this approach I'll continue to add tests.

**Note** One issue is that the anchors are rendered with the input instead of the picker, so the z-index context is different. I feel this is odd and the anchors should be part of the picker/popup. Though this is a much bigger change. Let me know what you think.

## Before

screen shot 2018-03-12 at 11 49 24

## After

react-dates__appendtobody

### Problem with a detached scrolling picker

screen shot 2018-03-12 at 11 55 01


joaovieira:
You can try it with `npm i joaovieira/react-dates#appendToBody-release`

joaovieira:
Great! Thanks for having a look @majapw. Will look into this later today:

- Add tests
- Document new props
- Investigate and start new PR to move anchors to popup

joaovieira:
@majapw I've added as much tests as I could. This a feature that is tied with the DOM, hence most of the tests require a browser and/or mounting (mocking all window/document would be too complex and/or tie the tests with the implementation even more). Thus, I've opened https://github.com/airbnb/react-dates/pull/1109 to fix Karma and ensure browser tests are mandatory in CI. Let me know your thoughts.

src/utils/disableScroll.js
```diff
@@ -0,0 +1,66 @@
+const getScrollingRoot = () => document.scrollingElement || document.documentElement;
+
+/**
+ * Recursively finds the scroll parent of a node. The scroll parrent of a node
+ * is the closest node that is scrollable. A node is scrollable if:
+ *  - it is allowed to scroll via CSS ('overflow-y' not visible or hidden);
+ *  - and its children/content are "bigger" than the node's box height.
+ *
+ * The root of the document always scrolls by default.
```
majapw: Thank you! This makes it super clear. :)

src/utils/getDetachedContainerStyles.js
```diff
@@ -0,0 +1,39 @@
...
+  if (openDirection === OPEN_UP) {
+    offsetY = -(window.innerHeight - referenceRect.bottom);
+  }
+
+  if (anchorDirection === ANCHOR_RIGHT) {
+    offsetX = -(window.innerWidth - referenceRect.right);
+  }
+
+  return {
+    transform: `translateX(${offsetX}px) translateY(${offsetY}px)`,
```
majapw: Would it be better to use `translate3D(${offsetX}px, ${offsetY}px, 0)` here for performance reasons?

joaovieira: You're right. Thought all CSS transforms were already hardware accelerated. I can do it. Thanks :)

majapw:
Sweet, I'll merge this in after tests pass!

PR #1095, opened 2018-03-28
export CustomizableCalendarDay styles, add story

a15n:
Previously if you wanted to update a single CSS style from CustomizableCalendarDay you'd need to (1) make the singular change and (2) duplicate all of the original CSS. 

With this change we can import the style definitions and update the single element using es6 spread operator or [update](https://github.com/kolodny/immutability-helper).

```
// before
const newDefaultStyles = {
  color: 'blue', // what we want to change
  // what we need to keep the same
  border: `1px solid ${color.core.borderLight}`,
  background: color.background,
  hover: {
    background: color.core.borderLight,
    border: `1px double ${color.core.borderLight}`,
    color: 'inherit',
  },
}


// after
import CustomizableCalendarDay, { defaultStyles, selectedStyles } from '../src/components/CustomizableCalendarDay';

const newDefaultStyles = {
  ...defaultStyles,
  color: 'blue',
};
```

### DIFF
--- a/src/components/CustomizableCalendarDay.jsx
+++ b/src/components/CustomizableCalendarDay.jsx
@@ -74,6 +74,109 @@ const propTypes = forbidExtraProps({
   phrases: PropTypes.shape(getPhrasePropTypes(CalendarDayPhrases)),
 });
 
+export const defaultStyles = {
+  border: `1px solid ${color.core.borderLight}`,
+  color: color.text,
+  background: color.background,
+
+  hover: {
+    background: color.core.borderLight,
+    border: `1px double ${color.core.borderLight}`,
+    color: 'inherit',
+  },
+};
+
+export const outsideStyles = {
+  background: color.outside.backgroundColor,
+  border: 0,
+  color: color.outside.color,
+};
+
+export const highlightedCalendarStyles = {
+  background: color.highlighted.backgroundColor,
+  color: color.highlighted.color,
+
+  hover: {
+    background: color.highlighted.backgroundColor_hover,
+    color: color.highlighted.color_active,
+  },
+};
+
+export const blockedMinNightsStyles = {
+  background: color.minimumNights.backgroundColor,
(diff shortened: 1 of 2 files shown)


src/components/CustomizableCalendarDay.jsx
```diff
@@ -291,45 +311,28 @@ class CustomizableCalendarDay extends React.Component {
...
-          modifiers.has('blocked-calendar') && blockedCalendarStyles,
-          hoveredSpan && hoveredSpanStyles,
-          modifiers.has('after-hovered-start') && afterHoveredStartStyles,
-          modifiers.has('selected-span') && selectedSpanStyles,
-          modifiers.has('last-in-range') && lastInRangeStyles,
-          selected && selectedStyles,
-          modifiers.has('selected-start') && selectedStartStyles,
-          modifiers.has('selected-end') && selectedEndStyles,
-          isOutsideRange && blockedOutOfRangeStyles,
+          getStyles(defaultStylesWithHover, isHovered),
```
a15n: couldn't think of a good intermediate name so I just execute everything inline

stories/DayPickerSingleDateController.js
```diff
@@ -206,6 +208,39 @@ storiesOf('DayPickerSingleDateController', module)
...
+      // extend and update styles with es6 spread operators
+      defaultStyles: {
+        ...defaultStyles,
+        color: 'blue',
+        hover: {
+          ...defaultStyles.hover,
+          color: 'blue',
+        },
+      },
+      // or immutability-helper's update
```
ljharb: i'm not sure this library needs calling out in an example

src/components/CustomizableCalendarDay.jsx
majapw: That seems reasonable to me

a15n:
Thanks for the review. Ready to merge

PR #1096, opened 2018-04-03
add verticalSpacing prop

amhunt:
This adds a verticalSpacing prop to DayPickerRangeController, which is passed all the way to `CalendarMonth`. This will allow users to add spacing between the rows in the month table. 

@majapw should we also add `horizontalSpacing` while we're at it? I don't currently have a use for it, but seems odd to only add one of the two. Maybe I should just add a `tableBorderSpacing` prop with two values: `vertical` and `horizontal`? 

NB: I also added `showInputs` as a declared prop for the wrapper. This was previously throwing a linting error.

screen shot 2018-04-02 at 5 05 58 pm

### DIFF
--- a/src/components/CalendarMonth.jsx
+++ b/src/components/CalendarMonth.jsx
@@ -161,6 +161,7 @@ class CalendarMonth extends React.Component {
       styles,
       phrases,
       dayAriaLabelFormat,
+      verticalSpacing,
     } = this.props;
 
     const { weeks } = this.state;
@@ -189,7 +190,10 @@ class CalendarMonth extends React.Component {
         </div>
 
         <table
-          {...css(styles.CalendarMonth_table)}
+          {...css(
+            !verticalSpacing && styles.CalendarMonth_table,
+            { borderSpacing: `0px ${verticalSpacing}px` },
+          )}
           role="presentation"
         >
           <tbody ref={this.setGridRef}>
--- a/src/components/CalendarMonth.jsx
+++ b/src/components/CalendarMonth.jsx
@@ -45,6 +45,7 @@ const propTypes = forbidExtraProps({
   renderDayContents: PropTypes.func,
   firstDayOfWeek: DayOfWeekShape,
   setMonthHeight: PropTypes.func,
+  verticalBorderSpacing: nonNegativeInteger,
 
   focusedDate: momentPropTypes.momentObj, // indicates focusable day
   isFocused: PropTypes.bool, // indicates whether or not to move focus to focusable day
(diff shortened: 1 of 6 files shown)


src/components/DayPicker.jsx
```diff
@@ -66,6 +66,7 @@ const propTypes = forbidExtraProps({
   verticalHeight: nonNegativeInteger,
   noBorder: PropTypes.bool,
   transitionDuration: nonNegativeInteger,
+  verticalSpacing: PropTypes.number,
```
ljharb: should this be non-negative, and an integer?

src/components/DayPickerRangeController.jsx
```diff
@@ -66,6 +66,7 @@ const propTypes = forbidExtraProps({
   hideKeyboardShortcutsPanel: PropTypes.bool,
   daySize: nonNegativeInteger,
   noBorder: PropTypes.bool,
+  verticalSpacing: PropTypes.number,
```
ljharb: also here

src/components/CalendarMonth.jsx
```diff
@@ -45,6 +45,7 @@ const propTypes = forbidExtraProps({
   renderDayContents: PropTypes.func,
   firstDayOfWeek: DayOfWeekShape,
   setMonthHeight: PropTypes.func,
+  verticalSpacing: nonNegativeInteger,
```
majapw: I think this name is somewhat overloaded as `verticalSpacing` in the case of the SDP and DRP signals the margin between the calendar dropdown and the inputs. Could we name this either `verticalBorderSpacing` to allude to the table or `verticalWeekSpacing` to allude to the calendar contents?

src/components/CalendarMonth.jsx
```diff
@@ -189,7 +192,11 @@ class CalendarMonth extends React.Component {
         </div>
 
         <table
-          {...css(styles.CalendarMonth_table)}
+          {...css(
+            !verticalSpacing && styles.CalendarMonth_table,
+            verticalSpacing && styles.CalendarMonth_verticalSpacing,
+            { borderSpacing: `0px ${verticalSpacing}px` },
```
majapw: Does this get weird when the verticalSpacing is undefined? Like it would try to apply a style of `0px undefinedpx` and then it would crap out. I think it would be better to do:
```
verticalSpacing && {
  borderCollapse: 'separate',
  borderSpacing: `0px ${verticalSpacing}px`,
},
```
or
```
verticalSpacing && styles.CalendarMonth_verticalSpacing,
verticalSpacing && { borderSpacing: `0px ${verticalSpacing}px` },
```

amhunt:
@majapw updated

majapw:
Sweet! Looks great to me! Will merge in and do a release.

PR #1106, opened 2018-04-06
[Issue 25] Unopinionated month and year selection support

GoGoCarl:
After tracking issue 25, which has been plaguing many users, myself included, I hope to have implemented, minimally, a stopgap solution to allow users to traverse months and years on the calendar, based on the excellent work contributed by @jvanderz22.

From the comments in the issue, it seems that the major sticking point for the past few months has been around design/UX. For this proposed solution, there is no design required from the maintainers; the caller is 100% responsible for the rendering.

To achieve this, this PR adds a single prop, `renderCaption`, which is expected to be a function that receives a data object as a param and returns a renderable custom caption (i.e. string, number, node).  The data object passed to `renderCaption` contains the following parameters:

* month -- a `moment` object for the current month
* onMonthSelect -- a callback function that takes two params
  * month - the `month` moment object
  * value - the integer value of the month to change to
* onYearSelect -- a callback function that takes two params
  * month - the month `moment` object
  * value - the integer value of the year to change to

Here is some sample code of how this could work with simple select inputs (this sample is also available in the storybook in this PR):
```
    <div style={{ display: 'flex', justifyContent: 'center' }}>
      <div>
        <select
          value={month.month()}
          onChange={(e) => { onMonthSelect(month, e.target.value); }}
        >
          {moment.months().map((label, value) => (
            <option value={value}>{label}</option>
          ))}
        </select>
      </div>
      <div>
        <select
          value={month.year()}
          onChange={(e) => { onYearSelect(month, e.target.value); }}
        >
          <option value={moment().year() - 1}>Last year</option>
          <option value={moment().year()}>{moment().year()}</option>
          <option value={moment().year() + 1}>Next year</option>
        </select>
      </div>
    </div>
```

Even if there is a solution later to present a pre-designed interface for selecting months and years, it would still be nice to have something like `renderCaption` to allow users to provide their own custom implementation.

As of this writing, this PR passes all tests and is updated to version 16.5.0.


majapw:
This is awesome, and I'll review it a bit more thoroughly today! :)

src/components/CalendarMonth.jsx
```diff
@@ -185,7 +194,8 @@ class CalendarMonth extends React.Component {
             verticalScrollable && styles.CalendarMonth_caption__verticalScrollable,
           )}
         >
-          <strong>{monthTitle}</strong>
+          {renderCaption && renderCaption({ month, onMonthSelect, onYearSelect })}
+          {!renderCaption && <strong>{monthTitle}</strong>}
```
ljharb: any reason not to make this be the default function for renderCaption?

src/components/CalendarMonthGrid.jsx
```diff
@@ -204,6 +216,32 @@ class CalendarMonthGrid extends React.Component {
...
+    const newMonth = currentMonth.clone();
+    const { orientation } = this.props;
+    const { months } = this.state;
+    const withoutTransitionMonths = orientation === VERTICAL_SCROLLABLE;
+    let initialMonthSubtraction = months.indexOf(currentMonth);
+    if (!withoutTransitionMonths) {
+      initialMonthSubtraction -= 1;
+    }
+    newMonth.set('month', newMonthVal).subtract(initialMonthSubtraction, 'months');
+    this.props.onMonthChange(newMonth);
```
ljharb: Please destructure this out, so the “this” value of the function isn’t the props object.

src/components/CalendarMonthGrid.jsx
```diff
@@ -204,6 +216,32 @@ class CalendarMonthGrid extends React.Component {
...
+    const newMonth = currentMonth.clone();
+    const { orientation } = this.props;
+    const { months } = this.state;
+    const withoutTransitionMonths = orientation === VERTICAL_SCROLLABLE;
+    let initialMonthSubtraction = months.indexOf(currentMonth);
+    if (!withoutTransitionMonths) {
+      initialMonthSubtraction -= 1;
+    }
+    newMonth.set('year', newYearVal).subtract(initialMonthSubtraction, 'months');
+    this.props.onYearChange(newMonth);
```
ljharb: Please destructure this out, so the “this” value of the function isn’t the props object.

src/utils/isNextMonth.js
```diff
@@ -0,0 +1,6 @@
+import moment from 'moment';
+
+export default function isNextMonth(a, b) {
+  if (!moment.isMoment(a) || !moment.isMoment(b)) return false;
```
ljharb: what non-moment values do you expect in these functions? I’d prefer a throw when it’s an unexpected value.

src/utils/isPrevMonth.js
```diff
@@ -0,0 +1,6 @@
+import moment from 'moment';
+
+export default function isPrevMonth(a, b) {
+  if (!moment.isMoment(a) || !moment.isMoment(b)) return false;
```
ljharb: Same here

src/components/CalendarMonth.jsx
GoGoCarl: Nope, sounds great, I'll update that.

src/components/CalendarMonthGrid.jsx
GoGoCarl: Gotcha 👍

src/utils/isNextMonth.js
GoGoCarl: I'll update this and the other spot, thanks.

GoGoCarl:
@ljharb I pushed a change to address a couple of your notes, but after some further review I left the type checks in the utils classes in place. The reason was because all the other similar utility classes, (i.e. https://github.com/airbnb/react-dates/blob/master/src/utils/isAfterDay.js) also do this check, so I think the code and behavior should remain consistent with what's already in place there. Thoughts?

ljharb:
Looks good; I’ll defer to @majapw about those functions (but consistency is probably better)

kevinnio:
My team is interested in having this PR merged. Can I have an estimate for when is this going to happen?

jgnieuwhof:
+1 on this, we have a team here waiting on this feature to go in, is there an estimate on when this will be merged?

gvincentlh:
We would **really** appreciate a merge as well 🙏

src/components/CalendarMonthGrid.jsx
```diff
@@ -204,6 +216,32 @@ class CalendarMonthGrid extends React.Component {
     onMonthTransitionEnd();
   }
 
+  onMonthSelect(currentMonth, newMonthVal) {
```
majapw: `currentMonth` is a momentObject, whereas `newMonthVal` is a zero-indexed integer?

src/components/DayPicker.jsx
```diff
@@ -419,6 +430,30 @@ class DayPicker extends React.Component {
     });
   }
 
+  onMonthChange(currentMonth) {
+    // Translation value is a hack to force an invisible transition that
```
majapw: :(

src/components/DayPickerSingleDateController.jsx
```diff
@@ -456,6 +484,10 @@ export default class DayPickerSingleDateController extends React.Component {
     return { currentMonth, visibleDays };
   }
 
+  setDayPickerRef(ref) {
```
majapw: Do we use this ref?

src/utils/isNextMonth.js
majapw: I think this is to allow one or both of a and b to be null (startDate and endDate are sometimes null)

src/utils/isNextMonth.js
```diff
@@ -0,0 +1,6 @@
...
+  return b.isSame(a.clone().add(1, 'month'), 'month');
```
majapw: I've found `isSame` to be slow even when it's got a 'month' or 'day' or whatever argument. Maybe let's do:
```
return b.month() === a.clone().add(1, 'month').month();
```

src/utils/isPrevMonth.js
```diff
@@ -0,0 +1,6 @@
...
+  return b.isSame(a.clone().subtract(1, 'month'), 'month');
```
majapw: I've found `isSame` to be slow even when it's got a 'month' or 'day' or whatever argument. Maybe let's do:
```
return b.month() === a.clone().subtract(1, 'month').month();
```

test/utils/isNextMonth__spec.js
```diff
@@ -0,0 +1,32 @@
+import moment from 'moment';
```
majapw: `isNextMonth_spec.js` (one underscore pls)

src/utils/isNextMonth.js
GoGoCarl: Only issue there is that a = 2017-01-01, b = 2018-02-01 would return true when it would be false. May not be a thing based on how the function is being called, but could be an issue down the line. To me, `isSame` may be slower, but it is safer.

Perhaps, if performance is a concern, checking a.year === b.year and a.month === b.month might help?

src/components/CalendarMonth.jsx
```diff
@@ -152,9 +158,12 @@ class CalendarMonth extends React.Component {
       onDayClick,
       onDayMouseEnter,
       onDayMouseLeave,
+      onMonthSelect,
+      onYearSelect,
       renderMonth,
       renderCalendarDay,
       renderDayContents,
+      renderCaption,
```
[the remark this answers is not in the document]
majapw: Let's leave renderMonth for right now and make the change to `renderMonthElement` in a follow-up PR (so as to separate out the breaking aspects)

src/utils/isNextMonth.js
majapw: good call!

src/components/DayPickerSingleDateController.jsx
majapw: This must be a relic of some old code that never got cleaned up

GoGoCarl:
All sounds good, thanks @majapw!

majapw:
\o/ Man i have been slow at getting to this. @GoGoCarl I don't want to put too much work on you randomly 2 months later :) so i'm gonna try to address my comments completely and get this merged in today.

I've noticed one small issue when the months change heights which I'll try to address before merging in. :)

GoGoCarl:
No worries, I'm just excited to see this getting closer to the finish line! If I can help out or debug just let me know, happy to help however I can.

majapw:
Hmm, I'm facing a weird issue where the height animation does not occur between non-neighboring months when using the custom caption. What's weird, is that neighboring months *do* trigger the animation, even when navigating via the caption. I think it has something to do with the transition time?

bkzl:
Hi 👋. I have already been using code from this PR as a patch in my app. There is one issue that you should be aware of. The caption is rendered "below" weekdays so in some cases (f.e. when custom components are used for select inputs) the weekdays overlap the content of the caption:

screen shot 2018-06-07 at 17 57 22

The way how the DOM tree looks like make it impossible to just use z-index. The current hacky workaround for me is to hide original weekdays with CSS and render them from `renderCaption` function:

```jsx
const captionRenderer = ({ month, onMonthSelect, onYearSelect }) => {
  const weekdays = moment.weekdaysMin();
  return (
    <div>
      <div className="DayPicker_weekHeader">
        <ul className="DayPicker_weekHeader_ul">
          {weekdays.map(day => (
            <li key={day} className="DayPicker_weekHeader_li">
              {day}
            </li>
          ))}
        </ul>
      </div>
    // ...
    </div>
  );
}
```

GoGoCarl:
@bkzl YES!! Thank you for mentioning this, I thought there was something else I wanted to flag but was convinced that it was the transition issue so I never thought about it again.

I, too, ran into the same issue and solved it the same way. I'll see if I can add a tweak to this that will help, unless you have some ideas on this @majapw

GoGoCarl:
Excellent! And sure @majapw will pull this down and have a look.

Also, I agree it may warrant its own issue. I also tried a few different solutions that I wasn't satisfied with enough to go with. It seems to be more of an issue of element ordering/positioning than z-index. But yeah, thinking caps needed.

GoGoCarl:
@majapw LGTM! I checked based on my notes in #1192 and resizes work as expected.

sauldeleon:
AHHHHH this is so close to the end! thanks both of you!

sauldeleon:
@GoGoCarl I have downloaded the branch to the package.json correctly and I have this working in my app, but there is a warning about OutsideClickHandler, could it be something about being fixed when updating this with master branch?

bkzl:
Yeah, the element ordering/positioning issue can defiantly be addressed in a separate PR and shouldn't block the release. Awesome work @GoGoCarl and @majapw! Thank you 👏

alexandrephilbertmondoin:
Hello, I'd like to test this out but I can't get it to work. I have inside my .jsx:

```
import 'react-dates-custom-month-year-navigation/initialize';
import { SingleDatePicker } from 'react-dates-custom-month-year-navigation';
import 'react-dates-custom-month-year-navigation/lib/css/_datepicker.css';
```
And I use the datepicker as I would normally do. All I see is the previous airbnb SingeDatePicker with no year selection. Do I need to pass a special argument when creating the picker?

sauldeleon:
Hi

I have downloaded the master branch of the library, as there is no TAG yet to include this PR.
As I have to build the library, I have found an error:

In the build process
```
> react-dates@16.7.0 build:css app/node_modules/react-dates
> node scripts/buildCSS.js "--optimize"

app/node_modules/react-dates/node_modules/react-outside-click-handler/build/OutsideClickHandler.js:42
  display: _propTypes2['default'].oneOf(Object.values(DISPLAY))TypeError: Object.values is not a function
    at Object.<anonymous> (app/node_modules/react-dates/node_modules/react-outside-click-handler/build/OutsideClickHandler.js:42:48)
    at Module._compile (module.js:570:32)
    at Module._extensions..js (module.js:579:10)
    at Object.require.extensions.(anonymous function) [as .js] (app/node_modules/react-dates/node_modules/babel-register/lib/node.js:152:7)
    at Module.load (module.js:487:32)
    at tryModuleLoad (module.js:446:12)
    at Function.Module._load (module.js:438:3)
    at Module.require (module.js:497:17)
    at require (internal/module.js:20:19)
    at Object.<anonymous> (app/node_modules/react-dates/node_modules/react-outside-click-handler/index.js:2:18)
```

This piece of code

```
var propTypes = (0, _airbnbPropTypes.forbidExtraProps)({
  children: _propTypes2['default'].node.isRequired,
  onOutsideClick: _propTypes2['default'].func.isRequired,
  disabled: _propTypes2['default'].bool,
  useCapture: _propTypes2['default'].bool,
  display: _propTypes2['default'].oneOf(Object.values(DISPLAY)),
})
```

Must be changed to

```
var propTypes = (0, _airbnbPropTypes.forbidExtraProps)({
  children: _propTypes2['default'].node.isRequired,
  onOutsideClick: _propTypes2['default'].func.isRequired,
  disabled: _propTypes2['default'].bool,
  useCapture: _propTypes2['default'].bool,
  display: _propTypes2['default'].oneOf(
    Object.keys(DISPLAY).map(function(key) {
      return DISPLAY[key]
    })
  ),
})
```

Or the build will crash.

There is another way of building the app or using this new feature?

Thanks in advance

ljharb:
@sauldeleon it sounds like you’re using an old version of node to do the build; try using node 8 or 10.

sauldeleon:
@ljharb thanks for the quick response.

You are totally right. Updating Node to the last version did the trick.

In addition, there is a warning regarding this PR

```
Warning: Failed prop type: DayPickerRangeController: unknown props found: onMonthChange, onYearChange
    in DayPickerRangeController
    in DateRangePicker
    in withStyles(DateRangePicker)
    in div
    in DateRangePickerWrapper
    in div
    in App
```

huchenme:
Any plan for next release? excited about this feature

majapw:
v17.0.0 coming v. soon (e.g. now :P)

sauldeleon:
WORKS like a charm! thanks @GoGoCarl  @majapw @ljharb !!!!

majapw:
@stahor posting on a closed PR is probably not the best way to get a discussion going! :) Can you please open an issue?

archy-bold:
Quick follow up to @bkzl's suggestion. You should ensure you get the days of the week in order relevant for the locale by passing `true` to `weekdaysMin`. Otherwise you always get them with Sunday as the first.

```
const weekdays = moment.weekdaysMin(true);
```

whydna:
@majapw I'm a bit confused - is the property to customize the control `renderCaption` or `renderMonthElement`? What's the difference?

majapw:
@whydna it was `renderCaption` in the original PR, but it was renaming to `renderMonthElement` to be a little bit more true to what it is actually doing. I would look at https://github.com/airbnb/react-dates/blob/master/stories/DayPickerRangeController.js#L151-L182 for the source code.

goranefbl:
in case someone stumbles upon here, the source code file has been updated few times, so just look for renderMonthElement property, currently @ https://github.com/airbnb/react-dates/blob/master/stories/DayPickerRangeController.js#L210-L233

gravitymedianet:
> Hi 👋. I have already been using code from this PR as a patch in my app. There is one issue that you should be aware of. The caption is rendered "below" weekdays so in some cases (f.e. when custom components are used for select inputs) the weekdays overlap the content of the caption:
> 
> screen shot 2018-06-07 at 17 57 22
> 
> The way how the DOM tree looks like make it impossible to just use z-index. The current hacky workaround for me is to hide original weekdays with CSS and render them from `renderCaption` function:
> 
> ```js-jsx
> const captionRenderer = ({ month, onMonthSelect, onYearSelect }) => {
>   const weekdays = moment.weekdaysMin();
>   return (
>     <div>
>       <div className="DayPicker_weekHeader">
>         <ul className="DayPicker_weekHeader_ul">
>           {weekdays.map(day => (
>             <li key={day} className="DayPicker_weekHeader_li">
>               {day}
>             </li>
>           ))}
>         </ul>
>       </div>
>     // ...
>     </div>
>   );
> }
> ```

Can you post a link to sample code to implement this? Year, month picker?

PR #1107, opened 2018-04-06
Fixed unexpected blur call

marlonpp:
I'm using react-dates inside a Drop. This drop is triggered by an input. As of now, when I transition months the drop is getting closed unexpectedly. My drop closes on input blur, and react-dates calls `activeElement.blur()` after the month transition. I'm not entirely sure why this is there, but I believe it should only be done if the activeElement is a child of your container.

Hopefully this PR looks good to you, otherwise let me know what needs to be changed.

FYI. I added .vscode to .gitignore so that editor files are not accidentally added.

### DIFF
--- a/src/components/DayPicker.jsx
+++ b/src/components/DayPicker.jsx
@@ -594,7 +594,11 @@ class DayPicker extends React.Component {
       // if the user is navigating around using a mouse
       if (withMouseInteractions) {
         const activeElement = getActiveElement();
-        if (activeElement && activeElement !== document.body) {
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -48,3 +48,5 @@ package-lock.json
 
 css/styles.css
 
+.vscode
```
ljharb: Please revert this and add it it your personal global gitignore; individual editors’ files shouldn’t force a change in every repo that individual uses.

PR #1109, opened 2018-04-08
Fix browser tests (Karma + Webpack setup).

joaovieira:
**TD;LR**
This has the same problem as https://github.com/airbnb/enzyme/issues/1318. Solved using https://github.com/airbnb/enzyme/issues/1318#issuecomment-342590767.

This also:
* fixes several PropTypes and Sinon warnings
* organizes the mocha setup in a single place - `mocha.opts`
* removes browser tests from "allowed failures"

**Explanation**
When using a files array, each file is fed into the webpack config creating an isolated bundle. When a bundle executes it is independent from other bundles (i.e. even if we load `tests/enzymeSetup.js` it only sets up the adapter for that bundle) - or so it seems.

Having a single bundle is also more efficient as webpack only needs to build once and there's no duplication.

The new entrypoint file is loading the tests the same way as mocha:
* load all files in `test/_helpers`
* load all `*_spec` files in `tests`


test/components/DateRangePicker_spec.jsx
```diff
@@ -18,6 +18,8 @@ import {
 const requiredProps = {
   onDatesChange: () => {},
   onFocusChange: () => {},
+  startDateId: 'startDate',
+  endDateId: 'endDate',
```
joaovieira: required

test/components/DateRangePicker_spec.jsx
```diff
@@ -316,7 +318,6 @@ describe('DateRangePicker', () => {
         const wrapper = shallow((
           <DateRangePicker
             {...requiredProps}
-            onDateChange={sinon.stub()}
```
joaovieira: not used; also `onDateChange` does not exist for `DateRangePicker` anymore

test/components/DayPickerKeyboardShortcuts_spec.jsx
```diff
@@ -117,7 +117,7 @@ describe('DayPickerKeyboardShortcuts', () => {
         });
 
         afterEach(() => {
-          openKeyboardShortcutsPanelStub.reset();
+          openKeyboardShortcutsPanelStub.resetHistory();
```
joaovieira: `sinon.reset()` has been deprecated

test/mocha.opts
```diff
@@ -2,3 +2,4 @@
 --compilers js:babel-register,jsx:babel-register
 --require airbnb-js-shims
 --recursive
+test/_helpers test/**/*_spec*
```
ljharb: This file needs a trailing newline; separately, the spec dir shouldn’t go here - i should be able to run `mocha path/to/file` and *not* run every test (and I’d prefer ever file inside “test” get treated as a spec, not just ones following a naming convention)

test/_helpers/ignoreSVGStrings.jsx
```diff
@@ -1,6 +0,0 @@
-// Don't load svg strings into tests
-require.extensions['.svg'] = (obj) => {
```
joaovieira: `require.extensions` is node.js specific (does not work in webpack/browser). Though, it also seems it's not necessary anymore as enzyme fixed support for SVGs (https://github.com/airbnb/enzyme/issues/375)?

test/browser-main.js
```diff
@@ -0,0 +1,6 @@
+const requireAll = requireContext => requireContext.keys().forEach(requireContext);
+
+if (typeof window !== 'undefined') {
```
joaovieira: I'd rather use Mocha's `--exclude` option (https://github.com/mochajs/mocha/pull/3210) once it lands instead of this.

ljharb: Mocha has dropped older node versions so we shouldn’t be upgrading too far on it; this approach is fine.

test/_helpers/ignoreSVGStrings.jsx
ljharb: The enzyme bug was on svg elements being present; this file makes svg files requirable in node, which is a different thing.

joaovieira: Thanks for clarifying. Then isn't what this is doing as well https://github.com/airbnb/react-dates/blob/master/.babelrc#L6? the `.svg` imports won't get to node.js - they'll be transformed before that in babel.

test/browser-main.js
joaovieira: Fair enough 👍

ljharb: that is true; but that is something i'm still working on fixing :-)

PR #1115, opened 2018-04-13
Patch onOutsideClick default value in the DayPickerSingleDateController

### LINKED ISSUE #1114: bug(SingleDatePicker): Uncaught TypeError "onOutsideClick is not a function"
I believe that this error only happens on the SingleDatePicker control. Furthermore, it seems to me that  this bug was introduced in [this PR](https://github.com/airbnb/react-dates/pull/643).

In order to reproduce just check [this example page](http://airbnb.io/react-dates/?selectedKind=SingleDatePicker%20%28SDP%29&selectedStory=default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel), open the developer tools and click on the calendar.

majapw:
Fixes #1114 

Whoops, I changed the way the `DayPickerSingleDateController` `onOutsideClick` prop looked to better match the `DayPickerRangeController`, but did not update the default prop value. 

to: @ljharb @ricardobrandao

### DIFF
--- a/src/components/DayPickerSingleDateController.jsx
+++ b/src/components/DayPickerSingleDateController.jsx
@@ -117,7 +117,7 @@ const defaultProps = {
 
   onPrevMonthClick() {},
   onNextMonthClick() {},
-  onOutsideClick: null,
+  onOutsideClick() {},
 
   renderCalendarDay: undefined,
   renderDayContents: null,


majapw:
@ljharb I think this requires both a full mount and actually simulating an outside click on the component to test... :/ I will think about how best to approach this problem.

PR #1116, opened 2018-04-13
Fix selective disabling of DateRangePicker

majapw:
From @ajwild: 
> @majapw some refactoring of pull request #606 caused a problem here. If disabled is used directly then it causes an error because the individual date pickers still expect a boolean but they can end up with the string values. If you use !!disabled then the endDate string can disable the start date and vice versa. I've reverted this part back to checking equality with true. If you have a suggestion for a better way to handle this let me know.

Git went all weird for me, but this is a version of https://github.com/airbnb/react-dates/pull/1112 with regression tests.

Thanks @ajwild. Sorry for the weirdness. :/

@ljharb

### DIFF
--- a/src/components/DateRangePickerInput.jsx
+++ b/src/components/DateRangePickerInput.jsx
@@ -201,8 +201,8 @@ function DateRangePickerInput({
       {calendarIcon}
     </button>
   );
-  const startDateDisabled = disabled === START_DATE || disabled;
-  const endDateDisabled = disabled === END_DATE || disabled;
+  const startDateDisabled = disabled === START_DATE || disabled === true;
+  const endDateDisabled = disabled === END_DATE || disabled === true;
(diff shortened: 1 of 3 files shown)


ajwild:
@majapw I'd just done something similar. I also added checks for `onFocusChange` on `DateRangePickerInputController_spec.js`.

majapw:
@ajwild if you share a commit with the controller spec you wrote I'll tack it on on top :)

ajwild:
Try this one: https://github.com/ajwild/react-dates/commit/1bede269ee046ffc8eeea99ef382761145247c07

majapw:
Sweet, I'll merge this in once tests pass again. :)

ajwild:
Thanks. I think #593 can be closed now :smile:

PR #1128, opened 2018-04-20
Fix DateInput state update when prop displayValue is ''

kumorko:


ljharb (approved):
This looks fine, but the test doesn't illustrate an actual bug fix - how does this change impact *consumers* of the component?

PR #1131, opened 2018-04-24
Fixes DateInput_Fang gap for DateInput with noBorder

travel-idealo:
Hi,

there is a 1px gap between "DateInput_Fang" and "DateRangePicker_picker", if property noBorder is set. This pr passes noBorder down to DateInput to handle this gap.


ljharb:
Unfortunately the fork was deleted, so the PR can never land.

PR #1135, opened 2018-04-25
New component: SingleDateRangeEndPicker

singingwolfboy:
screen shot of renderMonth

`DateRangePicker` is the perfect component to use when you want the user to select a pair of start/end dates at the same time. `SingleDatePicker` is the perfect component to use when you want the user to select just one date, without any other context. However, neither of these components are a perfect fit when you want the user to select a pair of start/end dates, but want the user to select them at _two different places_ in your form. You can use two different `SingleDatePicker` components, but the end date picker will not display the previously-selected start date, and cannot enforce validity checks (end date must be after start date, for example).

To address this, I built the `SingleDateRangeEndPicker`, which is a hybrid between `DateRangePicker` and `SingleDatePicker`. It has behavior similar to the SDP, in that the user selects only a single date, rather than a pair of dates. However, it also accepts a fixed `startDate` prop, so that when the user is selecting this end date, the user will see the date range between start and end, similar to the DRP.

I'm sure that this code can be refactored to reduce code duplication -- I basically copied the `SingleDatePicker.jsx` file and made the necessary changes to make it use `DayPickerRangeController` instead of `DayPickerSingleDateController`. I wanted to submit this pull request early, so that I could get feedback from others about whether this is a good idea, and how to get it merged into the project.

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -64,6 +64,7 @@ function loadStories() {
   require('../stories/SingleDatePicker_input');
   require('../stories/SingleDatePicker_calendar');
   require('../stories/SingleDatePicker_day');
+  require('../stories/SingleDateRangeEndPicker');
   require('../stories/DayPickerRangeController');
   require('../stories/DayPickerSingleDateController');
   require('../stories/DayPicker');
--- a/examples/SingleDateRangeEndPickerWrapper.jsx
+++ b/examples/SingleDateRangeEndPickerWrapper.jsx
@@ -1,0 +1,134 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import momentPropTypes from 'react-moment-proptypes';
+import moment from 'moment';
+import omit from 'lodash/omit';
+
+import SingleDateRangeEndPicker from '../src/components/SingleDateRangeEndPicker';
+
+import { SingleDatePickerPhrases } from '../src/defaultPhrases';
+import SingleDateRangeEndPickerShape from '../src/shapes/SingleDateRangeEndPickerShape';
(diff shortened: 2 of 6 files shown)


DiagoSmith:
This would also be a great way of getting around the lack of labels 👍 
What's the status of this MR currently?

ljharb:
Not sure what an MR is - this is a PR - but the status is that the questions posed in https://github.com/airbnb/react-dates/pull/1135#pullrequestreview-115667559 still need an answer from @singingwolfboy.

singingwolfboy:
@majapw @ljharb I just rebased the code and added a screenshot. I got the screenshot simply by running Storybook -- this pull request has Storybook stories in it already.

singingwolfboy:
@majapw @ljharb What's the next step for getting this reviewed?

singingwolfboy:
@majapw @ljharb ping

Shadowman4205:
@singingwolfboy  Hello, any news regarding this? Or is it dead?

singingwolfboy:
@Shadowman4205 I assume it's dead, since the Airbnb team doesn't seem interested in reviewing it.

singingwolfboy:
To be honest, I'm no longer interested in working with `react-dates`. [`react-day-picker`](http://react-day-picker.js.org/) is a better option.

Since I don't want to continue working on this, I'm closing this pull request. If the `react-dates` developers want to build a component like this, that's fine, but I'm not going to help.

ljharb:
Is there a reason it can’t stay open? That way someone else who’s interested can pick it up more easily.

Issue #1136, opened 2018-04-25
Incorrect position when changing openDirection


streamside:
The datepicker sometimes is rendered outside the viewport. To mitigate this problem I lookup the date input field and measure in which direction there is the most space. Based on this I change the `openDirection` to `UP` / `DOWN`. See code snippet B below for an example component.

Example of the problem:

![date-ranges-incorrect-position-datepicker](https://user-images.githubusercontent.com/7684435/39246292-698d935c-4896-11e8-9467-941662d2ce7f.PNG)

Notice that the fang is correctly placed above the field but the datepicker is not visible. The date picker is correctly placed if I close and then open it again.

I did some debugging of the code and found out that the `responsivizePickerPosition()` does the magic calculation of where to place the datepicker. The method is called from `componentDidUpdate()` when the datepicker is opened / closed.

This means that the position where to place the datepicker is cached and when `openDirection` is changed the picker gets an incorrect position. See code snippet A below for the change to address the problem.

Additional information:
* `appendToBody` is set to true as my field is placed in a scrollable container

Code Snippet A - DateRangePicker.js (code in react-dates):
```javascript
componentDidUpdate(prevProps) {
    if (!prevProps.focusedInput && this.props.focusedInput && this.isOpened()) {
        // The date picker just changed from being closed to being open.
        this.responsivizePickerPosition();
        this.disableScroll();
    } else if (prevProps.focusedInput && !this.props.focusedInput && !this.isOpened()) {
        // The date picker just changed from being open to being closed.
        if (this.enableScroll) this.enableScroll();
    } else if (prevProps.openDirection !== this.props.openDirection) {
        // The direction of the date picker has changed
        this.responsivizePickerPosition(); // THIS IS THE ADDED
    }
}
```

Code Snippet B - DateField (component using react-dates)
```javascript
const OpenDirection = {
    Down: 'down',
    Up: 'up'
};

class DateField extends React.Component {
    constructor(props) {
        super(props);
        this.inViewport = this.inViewport.bind(this);
        this.state = {
            openDirection: OpenDirection.Down
        };
    }

    componentDidUpdate() {
        this.inViewport();
    }

    inViewport() {
        const rect = this.fieldElement.getBoundingClientRect();
        const openDirection =
            rect.top > window.innerHeight - rect.bottom
                ? OpenDirection.Up
                : OpenDirection.Down;
        if (this.state.openDirection !== openDirection) {
            this.setState({ openDirection });
        }
    }

    render() {
[...]

PR #1145, opened 2018-04-27
Reposition picker when openDirection changes

streamside:
Resolving problem regarding incorrect positioning when changing `openDirection` on mounted component. See #1136 for more details.


austenpayan:
Any plans to fix/merge this one? Currently effected by the same issue as #1136

majapw:
I'll close this, given that the issue seems fixed @austenpayan but i'm also fairly confused as to why 🤔

Issue #1118, opened 2018-04-17
Blurry text in Chrome when calculated `translate3d` property contains non-integer values


aghreed:
Noticed this issue after pulling the date picker into an application and using the `appendToBody` prop.

I made a gif, but the quality makes it kinda hard to tell when its blurry or not.
![react-dates-blurry-transform](https://user-images.githubusercontent.com/1790506/38872480-f56dac56-4218-11e8-8bcf-54ba094efc12.gif)

But apparently this is an issue with webkit that has been for a while... http://martinkool.com/post/27618832225/beware-of-half-pixels-in-css

I believe this should be able to be solved by simply rounding the values with `Math.round()` at https://github.com/airbnb/react-dates/blob/master/src/utils/getDetachedContainerStyles.js#L37

More than happy to raise a PR when I get a little free time.

streamside:
Nice finding @aghreed! I've come across this problem as well.

@majapw Any chance that I or aghreed can submit a PR to resolve this problem?

PR #1155, opened 2018-05-03
Rounding values for translate3d to fix font blur in date picker 

RobiFerentz:
Rounding values for translate3d to fix font blur in date picker  on Chrome
Should solve issue #1118.


octaharon:
is it reproduced in other browsers than chrome?

RobiFerentz:
Seems to be an issue on chrome, but the pr doesnt affect other browser by any discernable difference.

PR #1158, opened 2018-05-04
Tweak default react-dates styling

majapw:
This is to align the react-dates defaults a bit better with Airbnb's brand. They're pretty subtle tweaks.

I decided to hide the arrow on small inputs because the text was overflowing the area. 

to: @garrettberg @ljharb


ljharb:
Can you show a before/after screenshot, particularly on the small breakpoint?

majapw:
Sure @ljharb !

Before (small is second):
screen shot 2018-05-06 at 8 31 21 pm
screen shot 2018-05-06 at 8 31 33 pm

After (small is second):
screen shot 2018-05-06 at 8 31 09 pm
screen shot 2018-05-06 at 8 31 00 pm

ljharb:
Hmm - could we use a shorter arrow instead of removing it entirely?

majapw:
@ljharb we use a dash (-) sometimes on the site, which is something I considered. Does that seem reasonable?

majapw:
The important thing for me is that the date doesn't get cut off.

ljharb:
Oh sure, i think any visual separator (altho better if it had a direction) is better than nothing. I agree the dates are the priority.

ljharb:
there’s a Unicode for a triangle pointing right that might fit

majapw:
I like the simplest option here:
screen shot 2018-05-07 at 8 40 44 pm

PTAL @ljharb

PR #1160, opened 2018-05-07
make nav buttons optional

amhunt:
## Summary
Beyond would like to be able to render our own nav buttons completely (rather than just be able to render inside of the existing nav button boxes). I believe that we can do this completely outside of the component by just adding a button that changes the number of months passed to `DayPickerRangeController`. However, to do this, we do _not_ want to render the existing `react-dates` nav buttons. This adds a boolean prop that makes the nav buttons optional.

This should also be helpful if people want to display only a couple of months and not allow users to move infinitely into the future (or any other custom navigation features).

## Reviewers
@majapw 
cc: @ljharb 

### Storybook example of `vertical_scrollable` orientation (didn't exist before)
screen shot 2018-05-07 at 2 08 40 pm
### Storybook example of new `noNavButtons` prop
screen shot 2018-05-07 at 2 53 12 pm

### DIFF
--- a/examples/DayPickerRangeControllerWrapper.jsx
+++ b/examples/DayPickerRangeControllerWrapper.jsx
@@ -130,7 +130,7 @@ class DayPickerRangeControllerWrapper extends React.Component {
     const endDateString = endDate && endDate.format('YYYY-MM-DD');
 
     return (
-      <div>
+      <div style={{ height: '100%' }}>
         {showInputs &&
           <div style={{ marginBottom: 16 }}>
             <input type="text" name="start date" value={startDateString} readOnly />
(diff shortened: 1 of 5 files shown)


ljharb:
> This should also be helpful if people want to display only a couple of months and not allow users to move infinitely into the future (or any other custom navigation features).

I think `isOutsideRange` already covers that by disabling navigation to those months.

majapw:
> I think isOutsideRange already covers that by disabling navigation to those months.
I'm working on this! Today hopefully. :)

amhunt:
@majapw do you mean this new prop will be useless once that feature is added? Or just that that one use case will be solved by your work? I think that this prop could still be useful, so that users don't have to create their own `isOutsideOfRange` function and update it every time they manually update the number of months or the starting month.

amhunt:
Also, @majapw, any idea what the travis failures are? Seems to be related to Node 6... but I have no idea why that would be the case or how the error message relates to this PR

ljharb:
@amhunt it's an npm 6 bug in node 6; it should be fixed now. i'll rerun your errored builds.

amhunt:
@majapw ptal :)

PR #1171, opened 2018-05-16
Improve documentation on internationalisation

HugoGiraudel:


README.md
```diff
@@ -309,11 +309,14 @@ The following is a list of other *OPTIONAL* props you may provide to the `DayPic
...
-```
+[Moment.js](http://momentjs.com) is a peer dependency of `react-dates`. The latter then uses a single instance of `moment` which is imported in one’s project. Loading a locale is done by calling `moment.locale(..)` in the component where `moment` is imported, with the [locale key](http://momentjs.com/docs/#/i18n/) of choice. For instance:
+
+```js
 moment.locale('pl'); // Polish
 ```
 
+However, this only solves date localization. For complete internationalisation of the components, `react-dates` defines a certain amount of [user interface strings](https://github.com/airbnb/react-dates/blob/master/src/defaultPhrases.js) in English which can be changed through the `phrases` prop (explore the [storybook](http://airbnb.io/react-dates/?selectedKind=DateRangePicker%20%28DRP%29&selectedStory=non-english%20locale&full=0&addons=1&stories=1&panelRight=0&addonPanel=kadirahq%2Fstorybook-addon-actions%2Factions-panel) for examples). For accessibility and usability concerns, **all these UI elements should be translated**.
```
ljharb: internationalization - the readme should be in US English

PR #1190, opened 2018-05-29
Add focusable="false" to dls-web svgs

majapw:
Because SVGs are focusable by default in IE, we need to add this attribute to prevent this behavior
and specifically to prevent a double-focus by screen-readers on our interactive buttons (e.g. month navigation).

to: @backwardok @erin-doyle


majapw:
@ljharb Yes! Sorry I meant to say that in the description. I just ran them through svgo. :)

PR #1191, opened 2018-05-30
Deprecate internal OutsideClickHandler in favor of react-outside-click-handler package

majapw:
This is a patch because I don't actually delete the OutsideClickHandler in this change but rather ship it with a deprecation message.

to: @ljharb


src/components/OutsideClickHandler.jsx
```diff
@@ -1,63 +1,8 @@
-import React from 'react';
-import PropTypes from 'prop-types';
+import OutsideClickHandler from 'react-outside-click-handler';
 
-// import { forbidExtraProps } from 'airbnb-prop-types'; // TODO: add to propTypes; semver-major
-import { addEventListener } from 'consolidated-events';
+console.log(`
```
ljharb: this should use `console.warn` instead, i think

majapw:
So for some reason the status check is stuck even though the job ran successfully on Travis - https://travis-ci.org/airbnb/react-dates/builds/385911685

@ljharb is this safe to merge or is something broken?

ljharb:
Refresh the page? Everything looks fine except that the coverage threshold still needs tweaking

majapw:
@ljharb when I click the travis link in coveralls (which what causes the pass or fail i thought), the travis job is green. Previously it told me that the branch coverage needed tweaking. Is it because of the missed lines in the diff?

screen shot 2018-05-31 at 8 36 05 am

Incidentally this missed line is the console.warn in the OutsideClickHandler... so if that's the case, I'm inclined to just merge through.

ljharb:
I think it's fine to merge through as long as it's not going to start failing on successive PRs.

PR #1192, opened 2018-06-01
Change month height calculation to be based on number of weeks instead of DOM calculations

majapw:
Hello friends! 

This is a precursor to getting https://github.com/airbnb/react-dates/pull/1106 merged in. I separated it out into its own change because it was fairly involved and I didn't want to pollute the code-change in #1106 with something not *entirely* related.

Basically, when playing around with the month/year selection I found an issue where if you jumped to a month of a different height (5 week month => 6 week month), the height wouldn't automatically adjust as expected. When I started trying to fix the issue, it became apparent that it wasn't possible to do without jank using the existing paradigms.

*Before:*
1. On render, each `CalendarMonth` component (including the two invisible ones for animation) would measure its own height in the DOM.
2. Each `CalendarMonth` would then call a callback prop, `setMonthHeight`, with this measured value.
3. The `CalendarMonthGrid` would add these values to an ordered array of `CalendarMonth` heights which it would then pass back up to the `DayPicker` via the `setCalendarMonthHeights` prop
4. Once the `DayPicker` would have all of the heights, it would calculate the expected height based on the max of the visible months and grow the calendar appropriately.
5. On month transitions, it would use this array of heights to grow or shrink the calendar appropriately while transitioning.

The important thing to note is that the only way it could change the height *while* transitioning was to use the self-calculations made by the invisible months that were already rendered to the DOM. When we want to transition to an arbitrary month or year, there's guaranteed to be jank in this system because we have to render the month before we know its height. The end result is that if you transition from July 2018 to July 2017, the 6-week month will be cut off before it grows. The other option is to hide the calendar in this interim, but that flash is also very apparent.

SO, this change is designed to amend that.

*Now:*
1. On the first render (and any time `renderMonth` changes), the `CalendarMonth` measures itself, subtracts out the height of the `CalendarDay` table (which is deterministic based on number of weeks and the `daySize` prop), and call the `setMonthTitleHeight` prop with the value. 
2. The `DayPicker` receives this value and saves it in `this.monthTitleHeight`
3. The `DayPicker` also maintains an array of the number of weeks in each visible month.
4. On every month transition, the `DayPicker` uses the number of weeks in the month coming into view along with the `daySize` prop and the saved `monthTitleHeight` to calculate the expected height of the incoming month and adjusts the height appropriately.

I've tested this in both the vertical and the horizontal pickers and it seems pretty effective! Because we can pretty readily calculate the number of weeks in any incoming month, this should open the door to slick month transitions to any month. We also, tbh, could probably remove the invisible months from render until it's time to slide them in. That can wait for another time though.

Note, this does assume a pretty constant month title height, but tbh I think that was an implicit assumption before. 

Would love some eyes! @ljharb @gabergg @GoGoCarl

These props that I'm changing are all what I would consider internal props (it would be weird if an outside consumer was relying on them for anything)... but are technically part of the API, so this change could be considered breaking. Idk. 

TODO: 
-[X] Write tests for `getNumberOfCalendarMonthWeeks`
-[X] reset month height when renderMonth changes

### DIFF
--- a/src/components/DayPicker.jsx
+++ b/src/components/DayPicker.jsx
@@ -412,14 +412,10 @@ class DayPicker extends React.Component {
   }
 
   onPrevMonthClick(nextFocusedDate, e) {
-    const { daySize, isRTL } = this.props;
-    const { calendarMonthWidth, currentMonth, monthTitleHeight } = this.state;
+    const { daySize, isRTL, numberOfMonths } = this.props;
+    const { calendarMonthWidth, monthTitleHeight } = this.state;
 
     if (e) e.preventDefault();
-
-    const newMonth = currentMonth.clone().subtract(1, 'month');
-    const numberOfWeeks = getNumberOfCalendarMonthWeeks(newMonth, this.getFirstDayOfWeek());
-    this.calendarMonthWeeks = [numberOfWeeks, ...this.calendarMonthWeeks.slice(0, -1)];
 
     let translationValue;
     if (this.isVertical()) {
@@ -431,7 +427,8 @@ class DayPicker extends React.Component {
         translationValue = -2 * calendarMonthWidth;
       }
 
-      const calendarMonthWeeksHeight = Math.max(0, ...this.calendarMonthWeeks) * (daySize - 1);
+      const visibleCalendarWeeks = this.calendarMonthWeeks.slice(0, numberOfMonths);
+      const calendarMonthWeeksHeight = Math.max(0, ...visibleCalendarWeeks) * (daySize - 1);
       const newMonthHeight = monthTitleHeight + calendarMonthWeeksHeight + 1;
       this.adjustDayPickerHeight(newMonthHeight);
     }
@@ -446,21 +443,17 @@ class DayPicker extends React.Component {
 
   onNextMonthClick(nextFocusedDate, e) {
     const { isRTL, numberOfMonths, daySize } = this.props;
-    const { calendarMonthWidth, currentMonth, monthTitleHeight } = this.state;
+    const { calendarMonthWidth, monthTitleHeight } = this.state;
 
     if (e) e.preventDefault();
-
-    const newMonth = currentMonth.clone().add(numberOfMonths, 'month');
-    const numberOfWeeks = getNumberOfCalendarMonthWeeks(newMonth, this.getFirstDayOfWeek());
 
     let translationValue;
 
     if (this.isVertical()) {
-      const calendarMonthWeeksHeight = this.calendarMonthWeeks[0] * (daySize - 1);
+      const firstVisibleMonthWeeks = this.calendarMonthWeeks[1];
+      const calendarMonthWeeksHeight = firstVisibleMonthWeeks * (daySize - 1);
       translationValue = -(monthTitleHeight + calendarMonthWeeksHeight + 1);
     }
-
-    this.calendarMonthWeeks = [...this.calendarMonthWeeks.slice(1), numberOfWeeks];
 
     if (this.isHorizontal()) {
       translationValue = -calendarMonthWidth;
@@ -468,7 +461,8 @@ class DayPicker extends React.Component {
         translationValue = 0;
       }
 
-      const calendarMonthWeeksHeight = Math.max(0, ...this.calendarMonthWeeks) * (daySize - 1);
+      const visibleCalendarWeeks = this.calendarMonthWeeks.slice(2, numberOfMonths + 2);
+      const calendarMonthWeeksHeight = Math.max(0, ...visibleCalendarWeeks) * (daySize - 1);
       const newMonthHeight = monthTitleHeight + calendarMonthWeeksHeight + 1;
       this.adjustDayPickerHeight(newMonthHeight);
     }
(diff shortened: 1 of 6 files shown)


src/components/DayPicker.jsx
```diff
@@ -399,19 +405,29 @@ class DayPicker extends React.Component {
...
+    const numberOfWeeks = getNumberOfCalendarMonthWeeks(newMonth, this.getFirstDayOfWeek());
+    this.calendarMonthWeeks = [numberOfWeeks].concat(this.calendarMonthWeeks.slice(0, -1));
+
+    let translationValue;
+    if (this.isVertical()) {
+      const calendarMonthWeeksHeight = this.calendarMonthWeeks[0] * (daySize - 1);
+      translationValue = monthTitleHeight + calendarMonthWeeksHeight + 1;
+    }
 
     if (this.isHorizontal()) {
```
ljharb: this seems like it should be an else if

src/components/DayPicker.jsx
```diff
@@ -399,19 +405,29 @@ class DayPicker extends React.Component {
...
-    const { calendarMonthWidth } = this.state;
+    const { daySize, isRTL } = this.props;
+    const { calendarMonthWidth, currentMonth, monthTitleHeight } = this.state;
 
     if (e) e.preventDefault();
 
-    let translationValue = this.isVertical() ? this.calendarMonthHeights[0] : calendarMonthWidth;
+    const newMonth = currentMonth.clone().subtract(1, 'month');
+    const numberOfWeeks = getNumberOfCalendarMonthWeeks(newMonth, this.getFirstDayOfWeek());
+    this.calendarMonthWeeks = [numberOfWeeks].concat(this.calendarMonthWeeks.slice(0, -1));
```
ljharb: `[numberOfWeeks, ...this.calendarMonthWeeks.slice(0, -1)]` ?

src/components/DayPicker.jsx
```diff
@@ -399,19 +405,29 @@ class DayPicker extends React.Component {
...
+
+    let translationValue;
+    if (this.isVertical()) {
+      const calendarMonthWeeksHeight = this.calendarMonthWeeks[0] * (daySize - 1);
```
ljharb: why does this only need the first week?
also, could it do `const [firstWeek] = this.calendarMonthWeeks;`?

src/components/DayPicker.jsx
```diff
@@ -424,18 +440,31 @@ class DayPicker extends React.Component {
...
     if (e) e.preventDefault();
 
-    let translationValue = this.isVertical() ? -this.calendarMonthHeights[1] : -calendarMonthWidth;
+    const newMonth = currentMonth.clone().add(numberOfMonths, 'month');
+    const numberOfWeeks = getNumberOfCalendarMonthWeeks(newMonth, this.getFirstDayOfWeek());
+
+    let translationValue;
+
+    if (this.isVertical()) {
+      const calendarMonthWeeksHeight = this.calendarMonthWeeks[0] * (daySize - 1);
```

src/components/DayPicker.jsx
```diff
@@ -424,18 +440,31 @@ class DayPicker extends React.Component {
...
+      translationValue = -(monthTitleHeight + calendarMonthWeeksHeight + 1);
```
ljharb: this "is vertical" logic is repeated in at least 2 places; can it be shared?

src/components/DayPicker.jsx
```diff
@@ -424,18 +440,31 @@ class DayPicker extends React.Component {
...
+    }
+
+    this.calendarMonthWeeks = this.calendarMonthWeeks.slice(1).concat([numberOfWeeks]);
```
ljharb: `const [, ...rest] = this.calendarMonthWeeks; this.calendarMonthWeeks = [...rest, numberOfWeeks];`?

src/components/DayPicker.jsx
majapw: `calendarMonthWeeks` is an array of the number of weeks in each of the calendarmonths. The vertical picker scrolls forward by the amount of weeks in the first (currently invisible) month.

src/components/DayPicker.jsx
majapw: SO there is a difference between the two. in the one the weeks array has already been updated to the new state; in the other it hasnt'.

src/components/DayPicker.jsx
majapw: it's not quite the same

src/components/DayPicker.jsx
majapw: okay

src/components/DayPicker.jsx
```diff
@@ -478,16 +521,31 @@ class DayPicker extends React.Component {
...
-    const lastVisibleMonthIndex = firstVisibleMonthIndex + numberOfMonths;
+    const { currentMonth } = this.state;
 
-    this.calendarMonthHeights = calendarMonthHeights;
-    const visibleCalendarMonthHeights = calendarMonthHeights
-      .filter((_, i) => ((i >= firstVisibleMonthIndex) && (i < lastVisibleMonthIndex)));
-    this.calendarMonthGridHeight = Math.max(0, ...visibleCalendarMonthHeights) + MONTH_PADDING;
-    this.setState({ hasSetHeight: true });
+    this.calendarMonthWeeks = [];
+    for (let i = 0; i < numberOfMonths; i += 1) {
```
gabergg: I wonder if we could tweak this to in the same way to capture months before and after the current month? And then on prev/next click, we could update the neighbor months asynchronously, the way that we might preload images in a carousel? I could be misunderstanding the logic too.

src/components/DayPicker.jsx
```diff
@@ -399,19 +412,27 @@ class DayPicker extends React.Component {
...
-    const { numberOfMonths, isRTL } = this.props;
-    const { calendarMonthWidth } = this.state;
+    const { daySize, isRTL } = this.props;
+    const { calendarMonthWidth, currentMonth, monthTitleHeight } = this.state;
 
     if (e) e.preventDefault();
 
-    let translationValue = this.isVertical() ? this.calendarMonthHeights[0] : calendarMonthWidth;
+    const newMonth = currentMonth.clone().subtract(1, 'month');
+    const numberOfWeeks = getNumberOfCalendarMonthWeeks(newMonth, this.getFirstDayOfWeek());
```
gabergg: I mostly think it would be cool to reduce the work done on prev/next click. Wondering if we can pre-calculate and key the months somehow, or even memoize `getNumberOfCalendarMonthWeeks` or something so that we don't have to recalculate week count and shift `calendarMonthWeeks` on each transition.

majapw:
Sweet! Okay, I deferred some of the heavier cloning logic from `onPrevMonthClick`/`onNextMonthClick` to `updateStateAfterTransition`. @GoGoCarl I think you're right that we'll likely want to abstract out that logic a bit in the month/year change. We can address that in your PR though, I think.

I'm gonna merge this in once everything passes.

PR #1201, opened 2018-06-07
Allow DateRangePickerInput border theme to be overriden

mmahalwy:


src/components/SingleDatePickerInput.jsx
```diff
@@ -217,7 +217,10 @@ export default withStyles(({ reactDates: { color } }) => ({
   },
 
   SingleDatePickerInput__withBorder: {
-    border: `1px solid ${color.core.border}`,
+    borderColor: color.border,
+    borderWidth: border.pickerInput.borderWidth,
+    borderStyle: border.pickerInput.borderStyle,
+    borderRadius: border.pickerInput.borderRadius,
```
mmahalwy: Maybe this is not needed? Not sure

src/components/DateRangePickerInput.jsx
```diff
@@ -296,8 +296,10 @@ export default withStyles(({ reactDates: { color, sizing } }) => ({
   },
 
   DateRangePickerInput__withBorder: {
-    border: `1px solid ${color.border}`,
```
ljharb: minor; but changing `border` might be a breaking change for those overriding it with CSS

mmahalwy: I thought in the most recent version of react-dates the classNames are scrambled, no?

ljharb: with the aphrodite interface sure, but i'm not sure about with the css interface

majapw: I think anyone overriding the border will see the same effect they did before, so I'm not concerned.

src/components/SingleDatePickerInput.jsx
majapw: def is. thank you for doing it in both pickers!

mmahalwy:
awesome, thanks @majapw

PR #1204, opened 2018-06-08
Remove default styling (border, etc.) from custom month nav buttons

majapw:
This change removes all default styling from custom navigation buttons (including position). This is a breaking change. 

I'm wondering if I should take this opportunity to change navNext/navPrev to render props a la the rest of the repo. Thoughts?

### Screenshots:

*Default*
screen shot 2018-06-19 at 2 02 54 pm
screen shot 2018-06-19 at 2 03 02 pm
screen shot 2018-06-19 at 2 03 19 pm

*Custom*
screen shot 2018-06-19 at 2 05 55 pm
screen shot 2018-06-19 at 2 06 04 pm
screen shot 2018-06-19 at 2 03 35 pm

to: @ljharb @amhunt

### DIFF
--- a/src/components/DayPickerNavigation.jsx
+++ b/src/components/DayPickerNavigation.jsx
@@ -95,13 +95,22 @@ function DayPickerNavigation({
     );
   }
 
+  const isDefaultNav =
+    isVerticalScrollable ? isDefaultNavNext : (isDefaultNavNext || isDefaultNavPrev);
+
   return (
     <div
       {...css(
-        styles.DayPickerNavigation_container,
-        isHorizontal && styles.DayPickerNavigation_container__horizontal,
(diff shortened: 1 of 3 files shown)


src/components/DayPickerNavigation.jsx
```diff
@@ -95,13 +95,21 @@ function DayPickerNavigation({
     );
   }
 
+  const isDefaultNav = isDefaultNavNext || (!isVerticalScrollable && isDefaultNavPrev);
```
ljharb: this might be more clear as
```js
const isDefaultNav = isVerticalScrollable ? isDefaultNavNext : (isDefaultNavNext || isDefaultNavPrev);
```
(i found the negation confusing)

Issue #937, opened 2018-01-03
[a11y] Tabbing away from date picker results in keyboard trap in Safari, Edge/IE


backwardok:
For both the [DateRangePicker](http://airbnb.io/react-dates/?selectedKind=DateRangePicker%20%28DRP%29&selectedStory=default&full=0&down=1&left=1&panelRight=0&downPanel=storybook%2Factions%2Factions-panel) and the [SingleDatePicker](http://airbnb.io/react-dates/?selectedKind=SingleDatePicker%20%28SDP%29&selectedStory=default&full=0&down=1&left=1&panelRight=0&downPanel=storybook%2Factions%2Factions-panel), if you tab to the date picker and then try to tab to the next thing after it, tab focus is reset to the top of the page and you're stuck in a keyboard trap.

Steps to reproduce:
1. Open one of the examples ([DateRangePicker](http://airbnb.io/react-dates/?selectedKind=DateRangePicker%20%28DRP%29&selectedStory=default&full=0&down=1&left=1&panelRight=0&downPanel=storybook%2Factions%2Factions-panel) or [SingleDatePicker](http://airbnb.io/react-dates/?selectedKind=SingleDatePicker%20%28SDP%29&selectedStory=default&full=0&down=1&left=1&panelRight=0&downPanel=storybook%2Factions%2Factions-panel)) using Safari or Edge
2. Tab to the date picker (note you'll go past the README link and the wrapper source link)
3. Tab out of the date picker

What's expected:
Next tab goes to Show Info button

What happens:
Next tab goes back to README link

Note that this works fine in Chrome, FF

Tested on Safari, Edge, IE11, Chrome (in Windows and Mac), FF (in Windows and Mac), Opera (in Windows)

rokborf:
FYI

This bug is caused by different browsers approaches to managing blur DOM method:

* Edge, Safari and IE after blur() reset tab order to start
* Chrome and Firefox - keep tab order, just unfocus an element

Demo: https://jsfiddle.net/Lebcjfeb/

Here blur method is used: https://github.com/airbnb/react-dates/blob/7dfe618bb5a6b6947799686114a00e70fe42a779/src/components/DateInput.jsx#L114

backwardok:
This issue was fixed a while back

PR #1214, opened 2018-06-15
Remove inputRef.blur() call that was causing a focus trap in Safari/IE

majapw:
Fix for https://github.com/airbnb/react-dates/issues/937

This code was originally added in https://github.com/airbnb/react-dates/pull/212, but weirdly I cannot reproduce the underlying issue. There have been a lot of changes to the internal structure of react-dates... but it is still surprising. :/ 

to: @backwardok @ljharb @wonnage


ljharb:
my bad; babel-preset-airbnb v2.5.0 is broken; v2.5.1 fixes it.

PR #1220, opened 2018-06-19
Rename renderMonth to renderMonthTitle and renderCaption to renderMonthElement

majapw:
Does this make sense? 

@GoGoCarl @ljharb @backwardok 

@backwardok suggested `renderMonthText` and `renderMonthHTML` but I don't know that that's very inline with other props in react-dates. `title` matches the current nomenclature? `renderTitleElement` might be clearer? Idk. 

I wanted to get this out with the latest breaking version.


GoGoCarl:
@majapw FWIW I actually like the combination of `renderMonthText` and `renderMonthElement`, or even `renderTitleText`/`renderTitleElement`, as the method names associate the two together as related, with the suffix giving sufficient context of what's expected to be returned (I'd prefer "Element" over "HTML").

Issue #1198, opened 2018-06-05
Uncaught TypeError: modifiers.has is not a function 


thinhdd88:
I get this error when click next month on DateRangePicker, 

Uncaught TypeError: modifiers.has is not a function
    at getCalendarDaySettings (getCalendarDaySettings.js?01ee:25)

philippefutureboy:
Hello!

I get the same issue here.
Here is my stack:

electron@2.0.2 (Chromium 61.0.3163.100)
webpack@4.8.3

moment@2.22.2
react-dom@16.4.1
react@16.4.1
react-dates@17.0.0

No polyfill used, and I'd rather not have to include extra dependencies if I can rely on babel already. .babelrc: 

```
{
  "presets": [
    ["env", {
      "targets": { "node": 7 },
      "useBuiltIns": true
    }],
    "stage-0",
    "react"
  ]
  ...
}
```

Error stack trace:

```
[2018-06-24T13:15:39.964Z] : renderer::components::decorators::ErrorBoundary : error : ErrorBoundary has catched an error

 TypeError: r.has is not a function
    at t.default (getCalendarDaySettings.js:25)
    at ProxyComponent.o.value (CalendarDay.js:225)
    at ProxyComponent.w (createClassProxy.js:93)
    at finishClassComponent (react-dom.development.js:7873)
    at updateClassComponent (react-dom.development.js:7850)
    at beginWork (react-dom.development.js:8225)
    at performUnitOfWork (react-dom.development.js:10224)
    at workLoop (react-dom.development.js:10288)
    at HTMLUnknownElement.callCallback (react-dom.development.js:542)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:581)
    at invokeGuardedCallback (react-dom.development.js:438)
    at renderRoot (react-dom.development.js:10366)
    at performWorkOnRoot (react-dom.development.js:11014)
    at performWork (react-dom.development.js:10967)
    at requestWork (react-dom.development.js:10878)
    at scheduleWorkImpl (react-dom.development.js:10732)
```

Highlighted code in the sourcemaps:

getCalendarDaySettings.js

```js
function getCalendarDaySettings(day, ariaLabelFormat, daySize, modifiers, phrases) {
  var chooseAvailableDate = phrases.chooseAvailableDate,
      dateIsUnavailable = phrases.dateIsUnavailable,
      dateIsSelected = phrases.dateIsSelected;


  var daySizeStyles = {
    width: daySize,
[...]

PR #1231, opened 2018-06-24
Pass through modifiers propType

ljharb:
This is related to #1198.

An alternative approach (which I'm happy to change to) would put the `modifiers` Set shape in a separate tiny file, and import it where needed.


ljharb:
@majapw yes, exactly right. Separate file it is!

ljharb:
k updated, PTAL - i also made the propType stricter.

PR #1240, opened 2018-06-29
[test] only run build on most recent node version

philipfweiss:
Build script fails if run on Node4. The build really only needs to be run once, and on newest version of node, so we make that change here.


philipfweiss:
Looks like this is probably going to fail as well, since `react-with-styles-interface-css` is not node4 compatible. We should still be running a build during our testing, though.

PR #640, opened 2017-07-25
Update conditions when adjustDayPickerHeight is called (new, with tests)

brosewarne:
I've attempted to complete #446 - I figured it was easier to make a new fork, rather than continue on the old one in #446 - I'm also still getting used to Github, so apologies if this isn't the right approach.

* Added checks for daySize and orientation in DayPicker.componentDidUpdate() so that DayPicker height is recalculated if either of these change between renders
* Added new tests for DayPicker.componentDidUpdate

I'm happy to keep this up to date and do whatever is needed to get it merged. I need this functionality as I need to switch the orientation from horizontal to vertical on XS breakpoint changes and back again when leaving XS. Currently this results in a 0 height DayPicker when going back to horizontal.

PR #1241, opened 2018-07-01
Update conditions when adjustDayPickerHeight is called

### LINKED ISSUE #443: Setting props responsively doesn't always trigger layout calculations
When changing properties such as `orientation` and `daySize` based on screen width, the DayPicker method `adjustDayPickerHeight` does not seem to fire consistently, so the calendar may have 0 height. It seems to work the first time and then not again. Maybe the condition that it is wrapped in here is blocking it?
https://github.com/airbnb/react-dates/blob/master/src/components/DayPicker.jsx#L224

Video of the issue: https://cl.ly/0c0j301T151L

As a related note, it would be awesome if there was a way to set the `daySize` as a % / fluid width. Doing so also does not allow the dimensions to update properly though.

nnishimura:
Fixes #443 

Replaces #446, #640

* Added checks for daySize and orientation in DayPicker.componentDidUpdate() so that DayPicker height is recalculated if either of these change between renders
* Added new tests for DayPicker.componentDidUpdate

### DIFF
--- a/src/components/DayPicker.jsx
+++ b/src/components/DayPicker.jsx
@@ -316,8 +316,20 @@ class DayPicker extends React.Component {
   }
 
   componentDidUpdate(prevProps) {
-    const { isFocused } = this.props;
-    const { focusedDate } = this.state;
+    const {
+      orientation, daySize, isFocused, numberOfMonths,
+    } = this.props;
+    const { focusedDate, monthTitleHeight } = this.state;
+
+    if (
+      this.isHorizontal()
+      && (orientation !== prevProps.orientation || daySize !== prevProps.daySize)
+    ) {
+      const visibleCalendarWeeks = this.calendarMonthWeeks.slice(1, numberOfMonths + 1);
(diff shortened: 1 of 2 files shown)


src/components/DayPicker.jsx
```diff
@@ -316,8 +316,20 @@ class DayPicker extends React.Component {
...
+    const {
+      orientation, daySize, isFocused, numberOfMonths,
+    } = this.props;
+    const { focusedDate, monthTitleHeight } = this.state;
+
+    if (
+      this.isHorizontal()
+      && (orientation !== prevProps.orientation || daySize !== prevProps.daySize)
+    ) {
+      const visibleCalendarWeeks = this.calendarMonthWeeks.slice(0, numberOfMonths);
```
majapw: ah shit, this should be `this.calendarMonthWeeks.slice(1, numberOfMonths + 1);` (we have to slice off the first and last months because they are hidden). I will update this.

majapw:
All right! This looks great! I'll merge in when tests pass. :)

Issue #1227, opened 2018-06-22
Uncaught TypeError: onOutsideClick is not a function


huchenme:
This error happens if I have 2 DatePicker on the app, one of them have `withPortal` equal to true.

demo: https://codesandbox.io/s/725j44217x

hrmoller:
@huchenme it even happens if you have just one DatePicker with `withPortal` set to `true` but then first when a date is selected: https://codesandbox.io/s/pkoj9x5mym

huchenme:
If I have a <SingleDatePicker/> with some other <SingleDatePicker withPortal />, and open <SingleDatePicker/> date selector, sometimes it takes me 3 outside clicks to close the opened date selector.

majapw:
I think this is related to the `react-outside-click-handler` upgrade (previously `onOutsideClick` wasn't required). Let me patch that real quick.

huchenme:
thanks @majapw

fuse-mars:
@majapw and @huchenme looks like the cause is from these two lines
* https://github.com/airbnb/react-dates/blob/master/src/components/DateRangePicker.jsx#L396
* https://github.com/airbnb/react-dates/blob/master/src/components/DateRangePicker.jsx#L529

Using `() => {}` instead of `undefined` should fix the issue

ljharb:
#1239 seems like it might address it.

PR #1256, opened 2018-07-09
Do not pass in undefined onOutsideClick to the OutsideClickHandler

majapw:
Fixes https://github.com/airbnb/react-dates/issues/1227

So I know we have https://github.com/airbnb/react-dates/pull/1239 open, but as an alternative, I wanted to propose this option instead as it wouldn't add the document event listeners at all (which seems preferable?). 

As for regression tests, @ljharb, we render `withPortal` datepickers in tests, shouldn't the proptype error show up there? 

to: @ljharb

### DIFF
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -529,9 +529,53 @@ class DateRangePicker extends React.Component {
 
     const { isDateRangePickerInputFocused } = this.state;
 
-    const onOutsideClick = (!withPortal && !withFullScreenPortal) ? this.onOutsideClick : undefined;
+    const enableOutsideClick = (!withPortal && !withFullScreenPortal);
 
     const hideFang = verticalSpacing < FANG_HEIGHT_PX;
+
+    const input = (
+      <DateRangePickerInputController
+        startDate={startDate}
+        startDateId={startDateId}
+        startDatePlaceholderText={startDatePlaceholderText}
+        isStartDateFocused={focusedInput === START_DATE}
+        endDate={endDate}
+        endDateId={endDateId}
+        endDatePlaceholderText={endDatePlaceholderText}
+        isEndDateFocused={focusedInput === END_DATE}
+        displayFormat={displayFormat}
+        showClearDates={showClearDates}
+        showCaret={!withPortal && !withFullScreenPortal && !hideFang}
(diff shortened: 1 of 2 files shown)


majapw:
@ljharb in fact, they do show up, but as warnings and not as errors? Is there any way to change that?

src/components/DateRangePicker.jsx
```diff
@@ -529,7 +529,9 @@ class DateRangePicker extends React.Component {
 
     const { isDateRangePickerInputFocused } = this.state;
 
-    const onOutsideClick = (!withPortal && !withFullScreenPortal) ? this.onOutsideClick : undefined;
+    const enableOutsideClick = (!withPortal && !withFullScreenPortal);
+    const Wrapper = enableOutsideClick ? OutsideClickHandler : 'div';
```
ljharb: it seems like we should be able to avoid adding the wrapping div entirely by only rendering the wrapper when `enableOutsideClick` is true?

majapw: I was thinking that I didn't want to change the DOM structure in this PR, but this is true. Do you think it could potentially cause problems if we removed this div?

ljharb: Hmm, I can't see why

src/components/SingleDatePicker.jsx
```diff
@@ -563,42 +602,19 @@ class SingleDatePicker extends React.Component {
...
-          {this.maybeRenderDayPickerWithPortal()}
-        </OutsideClickHandler>
+        <Wrapper {...wrapperProps}>
+          {enableOutsideClick
+            ? (
+              <OutsideClickHandler onOutsideClick={this.onOutsideClick}>
+                {input}
+                {this.maybeRenderDayPickerWithPortal()}
+              </OutsideClickHandler>
+            ) : ([
```
ljharb: it seems like we could avoid the array, and the keys, if we did:
```jsx
{enableOutsideClick && (
  <OutsideClickHandler…
)}
{!enableOutsideClick && input}
{!enableOutsideClick && this.maybeRenderDayPickerWithPortal()}
```

majapw: i was thinking about that but wasn't sure if that was preferable.

I guess making an array every time is p. gross. I'll update it

PR #1259, opened 2018-07-11
Fix minimum nights availability issue

majapw:
Some strange behavior that we noticed on airbnb.com is that after the first render (e.g. when you change focus or change the selected date) on a calendar with minimum nights enabled, while the actual interactivity of the minimum nights days still updates correctly, the aria-label would be the opposite of what was expected. This meant that when the days were grayed out due to not satisfying the minimum nights requirement, the aria label would read them as available and when they were clickable, like when the start date was focused, the aria label would read them as unavailable. 

Digging into this issue, it appeared to be related to how the `'blocked'` modifier was being updated when the state changed. Looking at `componentWillReceiveProps` in the `DayPickerRangeController`, where most of the logic for modifiers lives. 

Specifically, the inconsistency happened when focus changed and would remain thereafter. I think that the minimum nights logic would sometimes remove a 'blocked' modifier or forget to add a 'blocked' modifier unintentionally, which would flip the aria label from available to unavailable.

This change in ordering seems to address the issue.

@ljharb @backwardok @sdjidjev @airbnb/webinfra PTAL

### DIFF
--- a/src/components/DayPickerRangeController.jsx
+++ b/src/components/DayPickerRangeController.jsx
@@ -370,29 +370,16 @@ export default class DayPickerRangeController extends React.Component {
       }
     }
 
-    if (minimumNights > 0 && startDate && focusedInput === END_DATE) {
-      modifiers = this.addModifierToRange(
-        modifiers,
-        startDate,
-        startDate.clone().add(minimumNights, 'days'),
-        'blocked-minimum-nights',
-      );
-    }
-
     if (didFocusChange || recomputePropModifiers) {
       values(visibleDays).forEach((days) => {
         Object.keys(days).forEach((day) => {
           const momentObj = moment(day);
-
-          if (this.isBlocked(momentObj)) {
-            modifiers = this.addModifier(modifiers, momentObj, 'blocked');
-          } else {
-            modifiers = this.deleteModifier(modifiers, momentObj, 'blocked');
-          }
+          let isBlocked = false;
 
           if (didFocusChange || recomputeOutsideRange) {
             if (isOutsideRange(momentObj)) {
               modifiers = this.addModifier(modifiers, momentObj, 'blocked-out-of-range');
+              isBlocked = true;
             } else {
               modifiers = this.deleteModifier(modifiers, momentObj, 'blocked-out-of-range');
             }
@@ -401,9 +388,16 @@ export default class DayPickerRangeController extends React.Component {
           if (didFocusChange || recomputeDayBlocked) {
             if (isDayBlocked(momentObj)) {
               modifiers = this.addModifier(modifiers, momentObj, 'blocked-calendar');
+              isBlocked = true;
             } else {
               modifiers = this.deleteModifier(modifiers, momentObj, 'blocked-calendar');
             }
+          }
+
+          if (isBlocked) {
+            modifiers = this.addModifier(modifiers, momentObj, 'blocked');
+          } else {
+            modifiers = this.deleteModifier(modifiers, momentObj, 'blocked');
           }
 
           if (didFocusChange || recomputeDayHighlighted) {
@@ -415,6 +409,22 @@ export default class DayPickerRangeController extends React.Component {
           }
         });
       });
+    }
+
+    if (minimumNights > 0 && startDate && focusedInput === END_DATE) {
+      modifiers = this.addModifierToRange(
+        modifiers,
+        startDate,
+        startDate.clone().add(minimumNights, 'days'),
+        'blocked-minimum-nights',
+      );
+
+      modifiers = this.addModifierToRange(
+        modifiers,
+        startDate,
+        startDate.clone().add(minimumNights, 'days'),
+        'blocked',
+      );
     }
 
     const today = moment();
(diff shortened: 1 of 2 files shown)


majapw:
@lencioni that's a good idea. I may try to think of a way to make it less order dependent as well 🤔 

I was also walking through the function step-by-step and I still don't really understand... why this is happening. :|  so maybe I will revisit in addition to writing a test.

src/components/DayPickerRangeController.jsx
```diff
@@ -401,11 +388,18 @@ export default class DayPickerRangeController extends React.Component {
           if (didFocusChange || recomputeDayBlocked) {
             if (isDayBlocked(momentObj)) {
               modifiers = this.addModifier(modifiers, momentObj, 'blocked-calendar');
+              isBlocked = true;
             } else {
               modifiers = this.deleteModifier(modifiers, momentObj, 'blocked-calendar');
             }
           }
 
+          if (isBlocked) {
```
ljharb: this is covering isDayBlocked and isOutsideRange, but `this.isBlocked` also covers `doesNotMeetMinimumNights`. should that get a test case also?

test/components/DayPickerRangeController_spec.jsx
```diff
@@ -95,7 +95,7 @@ describe('DayPickerRangeController', () => {
     });
   });
 
-  describe('#componentWillReceiveProps', () => {
+  describe.only('#componentWillReceiveProps', () => {
```
ljharb: this needs to be reverted

test/components/DayPickerRangeController_spec.jsx
```diff
@@ -673,7 +673,7 @@ describe('DayPickerRangeController', () => {
             });
           });
 
-          describe('focusedInput changed', () => {
+          describe.only('focusedInput changed', () => {
```
ljharb: also this

test/components/DayPickerRangeController_spec.jsx
majapw: whoops! :) good catch.

src/components/DayPickerRangeController.jsx
majapw: That gets covered in a minimum nights specific code path because the triggers for re-calculating min nights are slightly different. It does get handled though!

src/components/DayPickerRangeController.jsx
```diff
@@ -417,6 +411,22 @@ export default class DayPickerRangeController extends React.Component {
...
 
+    if (minimumNights > 0 && startDate && focusedInput === END_DATE) {
+      modifiers = this.addModifierToRange(
+        modifiers,
+        startDate,
+        startDate.clone().add(minimumNights, 'days'),
+        'blocked-minimum-nights',
+      );
+
+      modifiers = this.addModifierToRange(
```
majapw: Here is where the blocked modifier will get added back if it has been removed by line 400 unnecessarily. @ljharb

majapw:
@ljharb PTAL! I think I have addressed your comments. :)

test/components/DayPickerRangeController_spec.jsx
```diff
@@ -741,6 +744,43 @@ describe('DayPickerRangeController', () => {
...
+              let day = moment(today);
+              for (let i = 0; i < minimumNights; i += 1) {
+                const monthString = toISOMonthString(day);
+                const dateString = toISODateString(day);
+                expect(visibleDays[monthString][dateString]).to.include('blocked-minimum-nights');
+                expect(visibleDays[monthString][dateString]).to.include('blocked');
+                day.add(1, 'day');
+              }
+
+              wrapper.instance().componentWillReceiveProps({
```
ljharb: btw i think this might be covered by `wrapper.setProps`? not a blocker tho

test/components/DayPickerRangeController_spec.jsx
```diff
@@ -765,79 +805,54 @@ describe('DayPickerRangeController', () => {
...
-              focusedInput: START_DATE,
+            it('updates state to include `blocked-minimum-nights` on the appropriate days', () => {
+              const startDate = today;
+              const minimumNights = 5;
+              const wrapper = shallow(<DayPickerRangeController
+                {...props}
+                startDate={startDate}
+                minimumNights={minimumNights}
+              />);
+              wrapper.instance().componentWillReceiveProps({
```
ljharb: ditto

test/components/DayPickerRangeController_spec.jsx
```diff
@@ -765,79 +805,54 @@ describe('DayPickerRangeController', () => {
...
-            ));
-            wrapper.setState({ visibleDays });
-            wrapper.instance().componentWillReceiveProps({
-              ...props,
-              minimumNights: 0,
-              focusedInput: END_DATE,
+                startDate={startDate}
+                minimumNights={minimumNights}
+              />);
+              wrapper.instance().componentWillReceiveProps({
```
ljharb: ditto

test/components/DayPickerRangeController_spec.jsx
majapw: This is what I've been using in all the tests in this describe block, so if I wanted to update I'd probably do everything all at once at a later date. :)

Issue #705, opened 2017-09-07
Selected dates not painted when load more months


cagiraudo:
Hi, i couldn't find the issue so i don't know if it's a known issue.

When you open the datepicker in mobile vertically and you load more months it doesn't paint when you select the date.

The same error is present in airbnb.com when you visit with mobile:

In this image you can see how it paint when i select the first date in the months loaded at begining:
![seleccion_031](https://user-images.githubusercontent.com/19225821/30166168-db5bfafc-93b8-11e7-853e-0a7171e88878.png)

And here you can see how it doesn't paint when i select a date in a month that was loaded after
![seleccion_032](https://user-images.githubusercontent.com/19225821/30166205-fc572ea2-93b8-11e7-82f1-ddc90ba631a4.png)

majapw:
Yup! Need to fix the vertical scrollable datepicker. X_x

crash7:
Hi @majapw 
,
I did some digging here, and found that the problem is related to [this condition](https://github.com/airbnb/react-dates/blob/master/src/components/DayPickerRangeController.jsx#L641-L643):
```
addModifier(updatedDays, day, modifier) {
    ....
    if (!day || !isDayVisible(day, currentMonth, numberOfMonths, enableOutsideDays)) {
      return updatedDays;
    }
```
`isDayVisible` is not considering the number of visible months in the calculation to validate the date.

Adding something like `numberOfMonths = Object.keys(visibleDays).length;` a couple of lines above fixes this particular problem but problably there is a better aproach to avoiding overriding a prop (seems hacky to me).
Example:
```
    if (orientation !== VERTICAL_SCROLLABLE) {
      currentMonth = currentMonth.clone().subtract(1, 'month');
      numberOfMonths += 2;
    } else {
      numberOfMonths = Object.keys(visibleDays).length;
    }
    if (!day || !isDayVisible(day, currentMonth, numberOfMonths, enableOutsideDays)) {
      return updatedDays;
    }
```

I checked the tests and adding this doesn't break anything but didn't submit a PR yet.

Daniellt1:
Hi @majapw, 

Any updates on this one?

The-Code-Monkey:
@majapw has this been fixed yet?

Daniellt1:
Is it fixed for `DayPickerSingleDateController` as well? I am still getting the same issue on `react-dates@18.1.0`.

PR #1261, opened 2018-07-14
Selected dates not painted when load more months issue

notjiam:
This PR fix about this issue https://github.com/airbnb/react-dates/issues/705


notjiam:
Hi, @ljharb I add some regression tests, I'm not sure is it correct could you review it.
Thank!

PR #1265, opened 2018-07-17
Set DayPickerNavigation__horizontal height to 0

majapw:
I ran into a specific edge-case where it turns out that the navigation buttons have an implicit expectation that this value is set to 0. Thus, in an application where there is a top-level line height is set, the navigation and week headers would get shoved down especially when using custom navigation. This change addresses that issue while not affecting existing datepickers. 

screen shot 2018-07-17 at 4 06 39 pm

to: @ljharb


majapw:
@ljharb the variation would be... that there's a global style on all dom elements :/

PR #1287, opened 2018-07-25
Giving internal state 'dateString' first priority for input value.

### LINKED ISSUE #1013: Cursor jumps to the end of the input filed when typing or deleting a character
This can easily be reproduced in the live playground:

i.e. in DateRangePicker - default:
- select a date for the start-date and the end-date with the mouse
- now set the cursor on the day or month of start- or end-date
- type one character or backspace

Observed behavior:
The cursor jumps to the end of the date field. This is very annoying if you want to set the date with the keyboard.

Desired behavior:
The cursor should stay where it is when typing in the start- or end-date.

nbakken8:
Resolve #1013 

It's an issue with [react](https://github.com/facebook/react/issues/955) when replacing the entire string value vs internal managed state.

This change will use dateString when the user is typing within the input field and switch over to displayValue when a date is selected via the calendar.

### DIFF
--- a/src/components/DateInput.jsx
+++ b/src/components/DateInput.jsx
@@ -187,7 +187,7 @@ class DateInput extends React.Component {
       theme: { reactDates },
     } = this.props;
 
-    const value = displayValue || dateString || '';
+    const value = dateString || displayValue || '';
     const screenReaderMessageId = `DateInput__screen-reader-message-${id}`;
 
     const withFang = showCaret && focused;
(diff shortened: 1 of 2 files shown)


nbakken8:
Test has been added that ensures the prop will take precedence over the state when passed in. Within the same file, tests on lines 19, 25, and 102 do a great job of covering scenarios as well.

imrank1:
This fix is awseome. Any idea when is a new version of react-dates going to be pushed to npm with this fix?

majapw:
@imrank1 will get it out as soon as https://github.com/airbnb/react-dates/pull/1305 is merged in! :)

PR #1293, opened 2018-08-01
fix for vertical scrollable modifiers

amhunt:
The number of months was incorrectly assumed to be constant in `addModifier` for the `VERTICAL_SCROLLABLE` orientation. I switched it to use the number of months in `visibleDays`. I was going to switch it for _all_ orientations, but then decided this would be less error prone (smaller change).

@majapw


ljharb:
@amhunt is this a duplicate of #1261?

ljharb:
I've rebased this on top of #1261; so merging this one can close both.

PR #1298, opened 2018-08-04
[vertical scrollable] add spacing and scrollbar theme

amhunt:
This adds three new theme variables, primarily aimed at the vertical scrollable configuration.

1. Adds the option to remove the scrollbar from the vertical scrollable orientation. The scroll bar looks bad on the lux calendar and also pushes things over so that elements are no longer aligned.
2. Adds a horizontal spacing prop for months (i.e. the padding around each individual month)
3. Adds a horizontal spacing prop for the daypicker container (padding around _all_ of the months).

I also rewrote the `getCalendarMonthWidth` function because it was using the wrong value (9 instead of 13 in the default case).

@majapw @ljharb

### DIFF
--- a/src/components/CalendarMonth.jsx
+++ b/src/components/CalendarMonth.jsx
@@ -249,7 +249,7 @@ export default withStyles(({ reactDates: { color, font, spacing } }) => ({
   CalendarMonth: {
     background: color.background,
     textAlign: 'center',
-    padding: '0 13px',
+    padding: `0 ${spacing.calendarMonthHorizontalPadding}px`,
     verticalAlign: 'top',
     userSelect: 'none',
   },
--- a/src/components/CalendarMonth.jsx
+++ b/src/components/CalendarMonth.jsx
@@ -25,7 +25,6 @@ import DayOfWeekShape from '../shapes/DayOfWeekShape';
 
 import {
   HORIZONTAL_ORIENTATION,
-  VERTICAL_ORIENTATION,
   VERTICAL_SCROLLABLE,
   DAY_SIZE,
 } from '../constants';
@@ -33,6 +32,7 @@ import {
 const propTypes = forbidExtraProps({
   ...withStylesPropTypes,
   month: momentPropTypes.momentObj,
+  horizontalMonthPadding: nonNegativeInteger,
   isVisible: PropTypes.bool,
   enableOutsideDays: PropTypes.bool,
   modifiers: PropTypes.objectOf(ModifiersShape),
@@ -62,6 +62,7 @@ const propTypes = forbidExtraProps({
 
 const defaultProps = {
   month: moment(),
+  horizontalMonthPadding: 13,
   isVisible: true,
   enableOutsideDays: false,
   modifiers: {},
@@ -155,26 +156,27 @@ class CalendarMonth extends React.Component {
 
   render() {
     const {
-      month,
-      monthFormat,
-      orientation,
+      dayAriaLabelFormat,
+      daySize,
+      focusedDate,
+      horizontalMonthPadding,
+      isFocused,
       isVisible,
       modifiers,
+      month,
+      monthFormat,
       onDayClick,
       onDayMouseEnter,
       onDayMouseLeave,
       onMonthSelect,
       onYearSelect,
-      renderMonthText,
+      orientation,
+      phrases,
       renderCalendarDay,
       renderDayContents,
       renderMonthElement,
-      daySize,
-      focusedDate,
-      isFocused,
+      renderMonthText,
       styles,
-      phrases,
-      dayAriaLabelFormat,
       verticalBorderSpacing,
     } = this.props;
 
@@ -187,9 +189,7 @@ class CalendarMonth extends React.Component {
       <div
         {...css(
           styles.CalendarMonth,
-          orientation === HORIZONTAL_ORIENTATION && styles.CalendarMonth__horizontal,
-          orientation === VERTICAL_ORIENTATION && styles.CalendarMonth__vertical,
-          verticalScrollable && styles.CalendarMonth__verticalScrollable,
+          { padding: `0 ${horizontalMonthPadding}px` },
         )}
         data-visible={isVisible}
       >
@@ -203,7 +203,9 @@ class CalendarMonth extends React.Component {
           {renderMonthElement ? (
             renderMonthElement({ month, onMonthSelect, onYearSelect })
           ) : (
-            <strong>{monthTitle}</strong>
+            <strong>
+              {monthTitle}
+            </strong>
           )}
         </div>
 
@@ -249,7 +251,6 @@ export default withStyles(({ reactDates: { color, font, spacing } }) => ({
   CalendarMonth: {
     background: color.background,
     textAlign: 'center',
-    padding: `0 ${spacing.calendarMonthHorizontalPadding}px`,
     verticalAlign: 'top',
     userSelect: 'none',
   },
--- a/src/components/CalendarMonthGrid.jsx
+++ b/src/components/CalendarMonthGrid.jsx
@@ -260,6 +260,7 @@ class CalendarMonthGrid extends React.Component {
       isFocused,
       isRTL,
       styles,
+      theme: { reactDates: theme },
       phrases,
       dayAriaLabelFormat,
       transitionDuration,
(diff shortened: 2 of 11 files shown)


src/components/CalendarMonthGrid.jsx
```diff
@@ -260,6 +260,7 @@ class CalendarMonthGrid extends React.Component {
       isFocused,
       isRTL,
       styles,
+      theme,
```
ljharb: `theme: { reactDates: theme }`?

src/components/CalendarMonthGrid.jsx
```diff
@@ -363,7 +367,11 @@ class CalendarMonthGrid extends React.Component {
 CalendarMonthGrid.propTypes = propTypes;
 CalendarMonthGrid.defaultProps = defaultProps;
 
-export default withStyles(({ reactDates: { color, zIndex } }) => ({
+export default withStyles(({
+  reactDates: {
+    color, noScrollBarOnVerticalScrollable, spacing, zIndex,
```
ljharb: each of these needs to be on its own line, if the entire object literal can't fit on one line

majapw:
Taking a look now!

src/components/CalendarMonthGrid.jsx
```diff
@@ -363,7 +367,14 @@ class CalendarMonthGrid extends React.Component {
 CalendarMonthGrid.propTypes = propTypes;
 CalendarMonthGrid.defaultProps = defaultProps;
 
-export default withStyles(({ reactDates: { color, zIndex } }) => ({
+export default withStyles(({
+  reactDates: {
+    color,
+    noScrollBarOnVerticalScrollable,
```
majapw: QQ: can you investigate as to whether the scrollbar is an issue on the `VERTICAL_ORIENTATION` as well? It may make sense to have `noScrollBarOnVertical` be the option

src/components/CalendarMonthGrid.jsx
```diff
@@ -272,7 +273,10 @@ class CalendarMonthGrid extends React.Component {
     const isVerticalScrollable = orientation === VERTICAL_SCROLLABLE;
     const isHorizontal = orientation === HORIZONTAL_ORIENTATION;
 
-    const calendarMonthWidth = getCalendarMonthWidth(daySize);
+    const calendarMonthWidth = getCalendarMonthWidth(
+      daySize,
+      theme.spacing.calendarMonthHorizontalPadding,
```
majapw: This is the reason why `calendarMonthHorizontalPadding` seems like it could be a prop instead of a theme variable. I get that it's only related to CSS---but because it also influences the animations and open source consumers have asked for this as well, I would maybe make this one a prop. We could call it `horizontalMonthPadding` to align with `verticalSpacing` and `verticalHeight` props.

src/components/CalendarMonthGrid.jsx
```diff
@@ -376,7 +387,7 @@ export default withStyles(({ reactDates: { color, zIndex } }) => ({
 
   CalendarMonthGrid__horizontal: {
     position: 'absolute',
-    left: 9,
+    left: spacing.dayPickerHorizontalPadding,
```
majapw: I think this makes sense in the theme.

src/components/DayPicker.jsx
```diff
@@ -956,7 +962,8 @@ class DayPicker extends React.Component {
       : 0;
 
     const firstVisibleMonthIndex = this.getFirstVisibleIndex();
-    const wrapperHorizontalWidth = (calendarMonthWidth * numberOfMonths) + (2 * DAY_PICKER_PADDING);
+    const wrapperHorizontalWidth = (calendarMonthWidth * numberOfMonths)
+      + (2 * dayPickerHorizontalPadding);
```
majapw: What do we use this for?

src/components/DayPicker.jsx
```diff
@@ -1093,7 +1100,12 @@ DayPicker.propTypes = propTypes;
 DayPicker.defaultProps = defaultProps;
 
 export { DayPicker as PureDayPicker };
-export default withStyles(({ reactDates: { color, font, zIndex } }) => ({
+
+export default withStyles(({
+  reactDates: {
+    color, font, noScrollBarOnVerticalScrollable, spacing, zIndex,
```
majapw: nit: I'd prefer these to be on their own lines (for ease of rebasing in the future)

src/utils/getCalendarMonthWidth.js
```diff
@@ -1,5 +1,3 @@
-const CALENDAR_MONTH_PADDING = 9;
-
-export default function getCalendarMonthWidth(daySize) {
-  return (7 * (daySize + 1)) + (2 * (CALENDAR_MONTH_PADDING + 1));
+export default function getCalendarMonthWidth(daySize, calendarMonthPadding) {
+  return (7 * daySize) + (2 * calendarMonthPadding) + 1;
```
majapw: This makes way more sense :P

src/components/DayPicker.jsx
ljharb: +1, see https://github.com/airbnb/react-dates/pull/1298#discussion_r207699220

src/components/CalendarMonthGrid.jsx
amhunt: It isn't/if there were an issue, it could be managed outside of the vertical calendar (because the largest div for the vertical calendar is the outermost div, whereas the smaller surrounding div in VERTICAL_SCROLLABLE is the one that's causing the scrollbar to appear).

src/components/DayPicker.jsx
amhunt: This is the width of the calendar for horizontal orientation. So if the outside padding of the dayPicker is 9 and the calendarMonth padding is 13, then the width of the horizontal calendar would be: `2*9 + 2*((2*13) + width_of_calendar_without_padding)`. Does that answer your question?

amhunt:
@majapw updated :) I made the horizontalMonthPadding a prop. I also removed three styles that didn't have corresponding style objects in `CalendarMonth`... idk why they were there.

My one remaining question - how should I handle the default case? I just set it as default 13 at all levels - or should it be default 13 at the top level and then a required prop below?

src/components/CalendarMonthGrid.jsx
```diff
@@ -386,6 +400,13 @@ export default withStyles(({ reactDates: { color, zIndex } }) => ({
   CalendarMonthGrid__vertical_scrollable: {
     margin: '0 auto',
     overflowY: 'scroll',
+    ...noScrollBarOnVerticalScrollable && {
```
ljharb: could we add parens here to make it clear that the entire thing is spread, not just the first conditional?

src/components/DayPicker.jsx
```diff
@@ -260,7 +264,8 @@ class DayPicker extends React.Component {
 
     if (nextProps.daySize !== daySize) {
       this.setState({
-        calendarMonthWidth: getCalendarMonthWidth(nextProps.daySize),
+        calendarMonthWidth:
+          getCalendarMonthWidth(nextProps.daySize, horizontalMonthPadding),
```
ljharb: this should be instead formatted like this:
```js
calendarMonthWidth: getCalendarMonthWidth(
  nextProps.daySize,
  horizontalMonthPadding,
),
```

src/components/DayPicker.jsx
```diff
@@ -1210,5 +1231,12 @@ export default withStyles(({ reactDates: { color, font, zIndex } }) => ({
     right: 0,
     left: 0,
     overflowY: 'scroll',
+    ...noScrollBarOnVerticalScrollable && {
```
ljharb: also parens here?

src/components/CalendarMonth.jsx
```diff
@@ -187,9 +189,7 @@ class CalendarMonth extends React.Component {
       <div
         {...css(
           styles.CalendarMonth,
-          orientation === HORIZONTAL_ORIENTATION && styles.CalendarMonth__horizontal,
-          orientation === VERTICAL_ORIENTATION && styles.CalendarMonth__vertical,
-          verticalScrollable && styles.CalendarMonth__verticalScrollable,
```
majapw: Hey @amhunt Did these get intentionally removed?

majapw: Ah I see below that they don't actually exist :x

Issue #339, opened 2017-02-20
Out of range months


danielzlotnik:
Would be great if we could disable navigation to "out of range" months (defined by a function, similar to out of range days).

Months where all days "out of range" should be out of range by default (and therefore unnavigatable) unless explicitly allowed.

Thanks!

ljharb:
If you're in April, and all of March is out of range, but February is in range, what happens when you try to go back a month?

danielzlotnik:
@ljharb rephrased the issue.

ljharb:
min/max date props might indeed make sense, thanks,

epifab:
min and max is for some cases a must have.. any progress on this?

ljharb:
Nobody’s submitted a PR yet.

erin-doyle:
This is on my team's list of things we'd like to work on at some point but it's pretty low in priority so if someone else needs it faster I'd suggest submitting a PR.  Otherwise I can get it it eventually. :woman_shrugging:

majapw:
Same here --- it is on the list, but it is not yet a high priority. The fastest way to get to it would be to submit a PR! :)

danielzlotnik:
@fernandofleury you'd still be able to navigate through all those unavailable months.

fernandofleury:
@danielzlotnik that is true. but personally, I don't see that as a big UI concern. but it might be different for your case

Vgurnani:
> Would be great to have the ability to limit months navigation to a specific range.
> 
> Could be achieved by min / max date props.
> 
> Thanks!

I tried with adding minDate in my wrapper but its now working as expected.

PR #1311, opened 2018-08-14
minDate and maxDate to block navigation

pedroabreu:
Adds feature to block month navigation depending on a minDate/maxDate prop.

Resolves https://github.com/airbnb/react-dates/issues/339


src/components/DayPickerNavigation.jsx
```diff
@@ -33,6 +35,8 @@ const propTypes = forbidExtraProps({
 });
 
 const defaultProps = {
+  hasPrev: true,
```
ljharb: boolean props should default to false, so that absence and falsiness are the same. Can the naming of these be inverted?

src/components/DayPickerRangeController.jsx
```diff
@@ -747,6 +752,23 @@ export default class DayPickerRangeController extends React.Component {
...
 
+  getNavigationStatus(visibleMonth) {
+    const {
+      minDate,
+      maxDate,
+      numberOfMonths,
+      enableOutsideDays,
+    } = this.props;
+
+    const isMinDateVisible = isDayVisible(minDate, visibleMonth, numberOfMonths, enableOutsideDays);
```
ljharb: If minDate is falsy, it seems like there’s no point in calling isDayVisible?

src/components/DayPickerRangeController.jsx
```diff
@@ -747,6 +752,23 @@ export default class DayPickerRangeController extends React.Component {
...
+    const isMaxDateVisible = isDayVisible(maxDate, visibleMonth, numberOfMonths, enableOutsideDays);
```
ljharb: Same here

src/utils/isDayVisible.js
```diff
@@ -1,7 +1,10 @@
+import moment from 'moment';
+
 import isBeforeDay from './isBeforeDay';
 import isAfterDay from './isAfterDay';
 
 export default function isDayVisible(day, month, numberOfMonths, enableOutsideDays) {
+  if (!moment.isMoment(day)) return false;
```
ljharb: This is a good change, even if my above comments are addressed (either this function should throw or return false on a non-moment object)

src/components/DayPicker.jsx
```diff
@@ -73,6 +73,8 @@ const propTypes = forbidExtraProps({
   horizontalMonthPadding: nonNegativeInteger,
 
   // navigation props
+  hasPrev: PropTypes.bool,
```
majapw: Let's call this `disablePrev` and `disableNext` to address Jordan's comments.

src/components/DayPickerNavigation.jsx
majapw: Agreed! Let's call this `disablePrev` and `disableNext` instead

src/components/DayPickerNavigation.jsx
```diff
@@ -138,8 +147,9 @@ function DayPickerNavigation({
             ]),
           )}
           aria-label={phrases.jumpToPrevMonth}
-          onClick={onPrevMonthClick}
+          onClick={hasPrev ? onPrevMonthClick : () => {}}
```
majapw: let's have this be undefined if it's disabled, so something like `hasPrev ? onPrevMonthClick : undefined` instead

src/components/DayPickerNavigation.jsx
```diff
@@ -138,8 +147,9 @@ function DayPickerNavigation({
...
           onKeyUp={(e) => {
+            if (!hasPrev) return;
```
majapw: Same here. let's just not define this function if hasPrev is false.

src/components/DayPickerNavigation.jsx
```diff
@@ -177,8 +188,9 @@ function DayPickerNavigation({
           ]),
         )}
         aria-label={phrases.jumpToNextMonth}
-        onClick={onNextMonthClick}
+        onClick={hasNext ? onNextMonthClick : () => {}}
```
majapw: Same comments apply here about using undefined instead of a thunk

src/components/DayPickerNavigation.jsx
```diff
@@ -177,8 +188,9 @@ function DayPickerNavigation({
...
         onKeyUp={(e) => {
+          if (!hasNext) return;
```
majapw: Let's not define this when we're disabling the button.

src/components/DayPickerNavigation.jsx
pedroabreu: Fair point. I've come up with `isPrevMonthBlocked`/`isNextMonthBlocked`, to try and keep it consistent with `isDayBlocked` prop.

src/components/DayPickerNavigation.jsx
```diff
@@ -121,6 +129,7 @@ function DayPickerNavigation({
           {...css(
```
majapw: @backwardok how do we mark a fake button as disabled? I'd like to do that in this scenario so that there's no confusion. `aria-disabled` maybe?

src/components/DayPickerRangeController.jsx
pedroabreu: Yep, it will return false in that case.

Happy to change it to make it more readable. I guess it's a good thing not to rely on the method accounting for that scenario (non-moment date)

src/components/DayPickerRangeController.jsx
```diff
@@ -747,6 +752,23 @@ export default class DayPickerRangeController extends React.Component {
     });
   }
 
+  getNavigationStatus(visibleMonth) {
```
majapw: Hmm, I'm not sure that this name is clear. I'd almost rather have two methods, one that asks if we should disable prev month navigation and one that asks about next month navigation.

src/utils/isDayVisible.js
pedroabreu: Followed the same pattern as other utils that do an early return if non-moment object, not sure if it was intended not to have a check but the tests don't break

src/components/DayPickerNavigation.jsx
backwardok: Yep, `aria-disabled` would be the right prop (https://www.w3.org/TR/wai-aria/#aria-disabled)

src/components/DayPickerRangeController.jsx
pedroabreu: I think it should be kept as one method, as the only difference between them would be the arguments `(date, visibleMonth)`. Maybe `shouldDisableMonthNavigation` ?

src/components/DayPickerNavigation.jsx
pedroabreu: Curious: I guess there's an advantage by using undefined instead of a thunk ?

backwardok: Setting it to undefined is akin to not having that prop set at all, which means no unnecessary click handler. Otherwise you're adding a click handlers to things that don't need them, only to call an empty function upon invocation.

pedroabreu:
@majapw any other thoughts on this ?

elinskytap:
any updates on this PR? 😃

src/components/DayPickerRangeController.jsx
```diff
@@ -666,6 +676,8 @@ export default class DayPickerRangeController extends React.Component {
     const newCurrentMonth = currentMonth.clone().subtract(1, 'month');
     this.setState({
       currentMonth: newCurrentMonth,
+      disablePrev: this.shouldDisableMonthNavigation(minDate, newCurrentMonth),
```
travel-idealo: should also be set as initial state inside constructor

Remownz: should also be set as initial state inside constructor

ljharb:
Please do rebase it, instead of clicking "update branch" which produces a merge commit.

pedroabreu:
Anything missing ? cc @majapw @ljharb @backwardok

src/components/DayPickerNavigation.jsx
```diff
@@ -137,9 +146,10 @@ function DayPickerNavigation({
               ] : []),
             ] : []),
           )}
+          aria-disabled={disablePrev}
```
backwardok:
```suggestion
          aria-disabled={disablePrev ? true : undefined}
```

src/components/DayPickerNavigation.jsx
```diff
@@ -137,9 +146,10 @@ function DayPickerNavigation({
...
           aria-label={phrases.jumpToPrevMonth}
-          onClick={onPrevMonthClick}
-          onKeyUp={(e) => {
+          onClick={disablePrev ? undefined : onPrevMonthClick}
+          onKeyUp={disablePrev ? undefined : (e) => {
             const { key } = e;
             if (key === 'Enter' || key === ' ') onPrevMonthClick(e);
           }}
```
backwardok: should the `onMouseUp` similarly be disabled when `disablePrev` is set?

src/components/DayPickerNavigation.jsx
```diff
@@ -176,9 +187,10 @@ function DayPickerNavigation({
             ] : []),
           ] : []),
         )}
+        aria-disabled={disableNext}
```
backwardok:
```suggestion
        aria-disabled={disableNext ? true : undefined}
```

src/components/DayPickerRangeController.jsx
```diff
@@ -217,6 +221,8 @@ export default class DayPickerRangeController extends React.PureComponent {
         chooseAvailableDate,
       },
       visibleDays,
+      disablePrev: false,
```
backwardok: can we prepopulate this given the initial props so that it's accurate on initial render?

src/components/DayPickerNavigation.jsx
pedroabreu: The `onClick` covers onMouse down/up on the same element. Is there a scenario where it would only get the `onMouseUp` event ?

elinskytap:
@majapw do you have time to review this? Really looking forward to seeing this to be released. Thanks!!

avivash:
I'm running v`20.1.0` of `react-dates`, but the props that were added here don't seem to be supported? 
image

pedroabreu:
Seems to be as easy as adding `minDate` and `maxDate` to the `DateRangePicker` component. I can have a look with more time tomorrow

avivash:
@pedroabreu thanks for the quick reply. When I try adding them to `DateRangePicker`, I get an error, as it doesn't recognize the props 
image

pedroabreu:
Sorry, I mean add them as props to the component and pass it down to the
controller.

avivash:
>Sorry, I mean add them as props to the component and pass it down to the
controller.

@pedroabreu sorry, I think I'm misunderstanding you. Are you suggesting a hierarchy like this?:
```
<DateRangePicker>
  <DayPickerRangeController
    startDate={this.state.startDate} // momentPropTypes.momentObj or null,
    endDate={this.state.endDate} // momentPropTypes.momentObj or null,
    onDatesChange={({ startDate, endDate }) => this.setState({ startDate, endDate })} // PropTypes.func.isRequired,
    focusedInput={this.state.focusedInput} // PropTypes.oneOf([START_DATE, END_DATE]) or null,
    onFocusChange={focusedInput => this.setState({ focusedInput })} // PropTypes.func.isRequired,
  />
</DateRangePicker>
```

ljharb:
I believe they are suggesting making a change in this package itself to pass those props down.

pedroabreu:
Sorry, it's what @ljharb mentioned. These props are not part of the `DateRangePicker`, only the `DateRangePickerController`.

I'll open a PR for this, should be pretty straightforward (famous last words)

avivash:
Thanks @ljharb / @pedroabreu. That's great! I'll be sure to look out for a PR

pedroabreu:
PR opened: #1582

avivash:
Thanks for getting this addressed so quickly!

ianduvall:
PR to improve upon the stories created by this PR: #1598

DBjelovuk:
Does this not work for the SingleDatePicker?

PR #1312, opened 2018-08-15
Tweak default border color

cshaver:
## Summary
This is a subtle tweak to border color to better align the `react-dates` input style with Airbnb's brand.

@majapw

## Screenshots

| Before | After | Changes highlighted in red |
| --- | --- | --- |
| drp-new | drp-old | drp-red |
| hover-new | hover-old | hover-red |
| scroll-new | scroll-old | scroll-red |

### DIFF
--- a/src/theme/DefaultTheme.js
+++ b/src/theme/DefaultTheme.js
@@ -6,7 +6,7 @@ const core = {
   grayLightest: '#f2f2f2',
 
   borderMedium: '#c4c4c4',
-  border: '#dbdbdb',
+  border: '#ebebeb',
   borderLight: '#e4e7e7',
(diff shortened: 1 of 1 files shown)


src/theme/DefaultTheme.js
```diff
@@ -6,7 +6,7 @@ const core = {
   grayLightest: '#f2f2f2',
 
   borderMedium: '#c4c4c4',
-  border: '#dbdbdb',
+  border: '#ebebeb',
```
ljharb: this seems like the sort of thing that airbnb could provide by a theme - and not require any changes in react-dates that some users might consider semver-major.

PR #1322, opened 2018-08-22
Remove propTypes in production

lencioni:
This will help reduce the bundle size impact. Note that any consumers
who are depending on any react-dates components having a `.propTypes`
property will no longer work as expected after this change.

I enabled this using wrap mode, which adds `process.env.NODE_ENV`
checks, which is a pretty standard method for React apps. I believe this
will allow this library to have propTypes in development, but for them
to be minified out in production builds.

To make this safe to enable in this repo, I enabled the
react/forbid-foreign-prop-types ESLint rule, which was written
specifically for the benefit of this plugin.

Here's a diff showing what effect this has on the esm build:

  https://gist.github.com/lencioni/1b436d365570394ccde659c829ba02c3


.babelrc
```diff
@@ -8,32 +8,51 @@
...
         ["transform-replace-object-assign", { "moduleSpecifier": "object.assign" }],
       ],
     },
+
     "production": {
       "presets": ["airbnb"],
       "plugins": [
         "inline-react-svg",
         ["transform-replace-object-assign", { "moduleSpecifier": "object.assign" }],
+        ["transform-react-remove-prop-types", {
```
ljharb: is this something we could add in babel-preset-airbnb, possibly behind an option, instead of in individual repos?

lencioni: Sure, I think so!

I think we'd still need to make it configurable though, since in some places we want the mode to be remove instead of wrap for instance, and other options might be important to manage on a case by case basis, but we could just have a passthrough option.

ljharb: sure, a passthrough option sounds great

lencioni:
@ljharb I just published the new preset with the removePropTypes option and updated this PR to use that instead.

lencioni:
Agreed! @majapw feel free to merge this when you think it makes sense for the next major.

Issue #345, opened 2017-02-23
Validate input using multiple formats


cssetian:
Hello @majapw, @ljharb, I am using react-dates in a project, and have a use case where I would like to be able to validate user input using multiple formats (e.g. YYYY-MM-DD, YYYY-M-D), but always display in a single specific format (YYYY-MM-DD). If I use YYYY-MM-DD as the displayFormat prop, it will not successfully validate a date entered in the YYYY-M-D format. 

Is there any support for allowing multiple moment formats to be passed for validating manual entry? What are your thoughts on having a separate prop for `InputFormats`, where an array of moment string formats can be passed and then spread into the `toMomentObject` fn?

evonsdesigns:
:+1: Or allow the consumer to pass in a function to validate inputs with.

majapw:
I think that allowing a consumer to pass in a function is actually a pretty good idea. That would give a lot more control over validation (which people have been asking for) and allow for whatever customization your app requires.

We'd probably want to couple this (perhaps with two PRs) with something that would report back that an error was received.

MaxiSantos:
I'd love to use this functionality. Right now I want a live validation and I don't know how to do it.

timweightman:
Would love it if this could get through to a release.

Currently our best option is to implement the <input /> ourselves, parse and format things with moment, handle all the focus, blur and several keyboard interactions ourself and use <DayPickerSingleDateController /> to show the calendar...

Floriferous:
Maybe simply passing an array could work, and then you use the first format in the array to actually display the date:

```jsx
<DatePicker
  dateFormat={["D MMMM YYYY", "D/M/YYYY", "D.M.YYYY"]}
/>
```

PR #1327, opened 2018-08-28
Allow different date input formats

damjan-gjurovski:
Allows specifying multiple acceptable input formats so users can manually add dates. For example, both dates in the DD.MM.YYYY, D.M.YYYY and D.M.YY format could be accepted.  Provides a simple implementation for the feature requested in #345

### DIFF
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -112,6 +112,7 @@ const defaultProps = {
 
   // internationalization
   displayFormat: () => moment.localeData().longDateFormat('L'),
+  inputFormats: [moment.localeData().longDateFormat('L')],
   monthFormat: 'MMMM YYYY',
   weekDayFormat: 'dd',
   phrases: DateRangePickerPhrases,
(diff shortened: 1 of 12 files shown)


src/components/DateRangePicker.jsx
```diff
@@ -112,6 +112,7 @@ const defaultProps = {
 
   // internationalization
   displayFormat: () => moment.localeData().longDateFormat('L'),
+  inputFormats: [moment.localeData().longDateFormat('L')],
```
majapw: Instead of this new prop, could we overload `displayFormat` so that it returned either an array or a string? 

I believe that not much else in the code would have to change.

majapw: And if it's an array, we could default to the first format. I wonder if that's too brittle. 🤔

timweightman:
Would love it if this could get through to a release.

Currently our only option is to implement the `<input />` ourselves, parse and format things with moment, handle all the focus, blur and several keyboard interactions ourself and use `<DayPickerSingleDateController />` to show the calendar...

src/components/DateRangePicker.jsx
damjan-gjurovski: @majapw Any news?

PR #1335, opened 2018-08-30
Conditionally use `PureComponent` instead of `Component`

### LINKED ISSUE #1297: Performance Issues
I am debugging the app I am working on for performance issues and have identified that react-dates might potentially be the biggest culprit of my slow down. Has anyone else run [why-did-you-update](https://github.com/maicki/why-did-you-update) on react dates and noticed the tons of potentially avoidable re-renders? On hovers, key press events all cause massive amounts of re-renders and none of the prop values have changed.

screen shot 2018-08-03 at 12 50 22 pm

AntiFish03:
Working to resolve #1297.  Conditionally using `React.PureComponent` instead of `React.Component`

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -1,6 +1,6 @@
 import React from 'react';
 if (process.env.NODE_ENV !== 'production') {
-  const {whyDidYouUpdate} = require('why-did-you-update');
+  const { whyDidYouUpdate } = require('why-did-you-update');
   whyDidYouUpdate(React);
 }
 
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -1,4 +1,9 @@
 import React from 'react';
+if (process.env.NODE_ENV !== 'production') {
+  const {whyDidYouUpdate} = require('why-did-you-update');
+  whyDidYouUpdate(React);
+}
+
 import moment from 'moment';
 import aphroditeInterface from 'react-with-styles-interface-aphrodite';
 
--- a/src/components/CalendarDay.jsx
+++ b/src/components/CalendarDay.jsx
@@ -1,6 +1,5 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import shallowCompare from 'react-addons-shallow-compare';
 import momentPropTypes from 'react-moment-proptypes';
 import { forbidExtraProps, nonNegativeInteger } from 'airbnb-prop-types';
 import { css, withStyles, withStylesPropTypes } from 'react-with-styles';
@@ -10,6 +9,7 @@ import { CalendarDayPhrases } from '../defaultPhrases';
 import getPhrasePropTypes from '../utils/getPhrasePropTypes';
 import getCalendarDaySettings from '../utils/getCalendarDaySettings';
 import ModifiersShape from '../shapes/ModifiersShape';
+import baseClass, { pureComponentAvailable } from '../utils/baseClass';
 
 import { DAY_SIZE } from '../constants';
 
@@ -48,15 +48,13 @@ const defaultProps = {
   phrases: CalendarDayPhrases,
 };
 
-class CalendarDay extends React.Component {
+const BaseClass = baseClass();
+
+class CalendarDay extends BaseClass {
   constructor(...args) {
     super(...args);
 
     this.setButtonRef = this.setButtonRef.bind(this);
-  }
-
-  shouldComponentUpdate(nextProps, nextState) {
-    return shallowCompare(this, nextProps, nextState);
   }
 
   componentDidUpdate(prevProps) {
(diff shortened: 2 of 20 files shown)


src/components/CalendarDay.jsx
```diff
@@ -48,17 +48,15 @@ const defaultProps = {
   phrases: CalendarDayPhrases,
 };
 
-class CalendarDay extends React.Component {
+const BaseClass = baseClass();
+
+class CalendarDay extends BaseClass {
```
ljharb: this is going to break react linting; we'll need a jsdoc comment to indicate that it extends React.Component

src/components/CalendarDay.jsx
```diff
@@ -48,17 +48,15 @@ const defaultProps = {
   phrases: CalendarDayPhrases,
 };
 
-class CalendarDay extends React.Component {
+const BaseClass = baseClass();
```
ljharb: `extends baseClass()` works too

src/utils/baseClass.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import shallowCompare from 'react-addons-shallow-compare';
+
+export const pureComponentAvailable = typeof React.PureComponent !== 'undefined';
+
+export default function baseClass(pureComponent = pureComponentAvailable) {
```
ljharb: why is `baseClass` a function, but `pureComponentAvailable` is not?

I'd expect either both to be a function (to support polyfilling of PureComponent, i assume), or neither.

AntiFish03:
@ljharb what does the jsdoc comment need to be? Your linting system is different than I am used to working with.

AntiFish03:
Also, @majapw this PR only resolves most of the avoidable re-rendering issue.  As I mentioned in #1297, I was not able to resolve some re-rendering on `DayPickerNavigation` or on `DayPicker`

ljharb:
(what are you used to working with that’s not eslint?) i believe it’s `/** @extends React.Component */` but I’ll have to check the docs and get back to you. You can test by introducing a lint error, verifying that it fails when it extends React.Component, verify that it passes when extending baseClass, and then verify that it fails with the JSdoc comment.

AntiFish03:
@ljharb not in that way.  I guess I never knew about being able to do that is a better way of describing it.

ljharb:
@AntiFish03 here's an example: https://github.com/yannickcr/eslint-plugin-react/pull/513#issue-64025078

ljharb:
@AntiFish03 what's the reasoning that the baseClass stuff needs to be in a function, instead of caching it once at module load time?

AntiFish03:
It doesn't have to be. Just trying to keep it similar to what is used in react-with-styles as was requested.

AntiFish03:
travis-ci decided to have some timeout issues there for no apparent reason... @ljharb care to restart that and everything should be good to go

ljharb:
After a few discussions, i wonder if it'd be worth it to try something like this?
```jsx
/** @extends React.Component */
class Foo extends React.PureComponent || React.Component {
  [React.PureComponent || 'shouldComponentUpdate'](nextProps, nextState) {
  }
}
```

Otherwise, your change LGTM.

AntiFish03:
@ljharb, Change made as requested anything else?

AntiFish03:
@ljharb, Readme updated.

.storybook/config.js
```diff
@@ -1,4 +1,9 @@
 import React from 'react';
+if (process.env.NODE_ENV !== 'production') {
+  const {whyDidYouUpdate} = require('why-did-you-update');
```
majapw: nit: spaces around the `whyDidYouUpdate`
```
const { whyDidYouUpdate } = require('why-did-you-update');
```

src/utils/baseClass.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import shallowCompare from 'react-addons-shallow-compare';
+
+class BaseClass extends (React.PureComponent || React.Component) {
+  [React.PureComponent || 'shouldComponentUpdate'](nextProps, nextState) {
```
majapw: I'm not sure I fully understand what this will do in the case that `React.PureComponent` is defined

AntiFish03: shouldn't it actually be `!React.PureComponent &&` so that it injects the `shouldComponentUpdate` when PureComponent isn't available.

ljharb: aha, yes, that's better. coding is hard.

AntiFish03:
@majapw, @ljharb There you go both of those pieces should be fixed now

majapw:
So fun fact, I believe that this breaks react-dates in IE10. *sigh* It seems to be related to how this is transformed by babel (current working theory is something about the double-extends). 
I'm seeing `Cannot call a class as a function` coming from https://unpkg.com/react-dates@18.0.2/lib/utils/baseClass.js.

AntiFish03:
@majapw all of a sudden I am getting
```
Warning: DateRangePickerInputController has a method called shouldComponentUpdate(). shouldComponentUpdate should not be used when extending React.PureComponent. Please extend React.Component if shouldComponentUpdate is used.
```
Is this something you are seeing also?

majapw:
@AntiFish03 locally or in storybook? on the latest version?

AntiFish03:
@majapw, Locally from master.

Issue #898, opened 2017-12-12
VerticalScrollable dateRangePicker would be nice


Bubbi:
As far as I can see, it is not possible to set orientation to verticalScrollable on a dateRangePicker. Only on dayPickerRange. it would be nice to include this, as you can then have a scrollable fullscreen portal with for example 6 months. Not possible atm.

nolastan:
+1

Adding the property `orientation="verticalScrollable"` seems to accomplish what you're looking for, but it throws a warning:

```
Warning: Failed prop type: Invalid prop `orientation` of value `verticalScrollable` 
supplied to `withStyles(DateRangePicker)`, expected one of ["horizontal","vertical"]
```

robhuzzey:
Hey, is there any movement on this as I'd like to use `verticalScrollable` on a project :)

nshimiye:
Any update on this request?

mrseanbaines:
Any way to prevent that warning being thrown? I'm using `VERTICAL_SCROLLABLE ` and seems to work fine but would be nice to have a clean console.

dougmcmillan:
+1 here. Possibly that it is not fully supported?

Kyngo:
> +1 here. Possibly that it is not fully supported?

It is possible, but it seems the only thing left to do is to add "verticalScrollable" as a valid word for this parameter.

dani9417:
Any movement on this? Would be really nice to have this feature!

PR #1353, opened 2018-09-12
Allows the use of VERTICAL_SCROLLABLE constant in DateRangePicker com…

theolampert:
I'm unsure if this is actually wanted or not but there seems to be a few [others](https://github.com/airbnb/react-dates/issues/898) wanting to use `DateRangePicker` component with vertical scrolling. 

Currently it works fine but results in an ugly proptype warning. This PR attempts to address that by importing `ScrollableOrientationShape` rather than `OrientationShape` within `DateRangePickerShape`


ljharb:
Sorry for the delay. This is a great bug fix, but it needs a regression test, can you add one?

theolampert:
@ljharb sorry closed this thinking it was no longer needed. Yeah I can add a test.

theolampert:
@ljharb Could you take another look?

theolampert:
@ljharb done 🙂

ljharb:
@theolampert I've rebased this; could we add some storybook stories that cover this new orientation?

theolampert:
@ljharb I wasn't able to get storybook to build on an M1 mac, complains about node-sass missing so I assume it's some native module thing I didn't really want to debug. I'll try and take a look from an intel machine either tonight or tomorrow.

ljharb:
@theolampert node-sass requires very specific node versions - in this case, i think node 14? and on an M1 mac, node < 16 requires installing it via rosetta.

Issue #1352, opened 2018-09-10
Date Picker width not being calculated


codylawson:
I'm getting the following error: ```Warning: `NaN` is an invalid value for the `width` css style property.```

Which results in the date picker displaying like this:
<img width="661" alt="screen shot 2018-09-10 at 9 31 05 am" src="https://user-images.githubusercontent.com/5149922/45309949-346d7f80-b4e2-11e8-96c6-2e77c29eb5b1.png">

Seems to happen in all browsers, and we're currently using the latest `18.0.2`. `react-with-styles-interface-aphrodite` is also on the latest `5.0.1`.

When I manually add `width: 619px;` to the popover it displays correctly:
<img width="682" alt="screen shot 2018-09-10 at 9 31 40 am" src="https://user-images.githubusercontent.com/5149922/45310079-90380880-b4e2-11e8-8ef6-194914ae4b58.png">

Have you seen any similar issues? Thanks!

majapw:
Hmm, that seems unusual (and does not seem to be happening in the storybook environment). Can you share the props you are passing in and any display/visibility related CSS overrides you may be applying?

codylawson:
@majapw Thanks for the quick reply. Here's a code sandbox with my implementation. https://codesandbox.io/s/53yvnmm2k

majapw:
Huh, so it is! I will take a look at what may be going on.

codylawson:
@majapw are there any documented breaking changes between `v17.1.1` and `v18.0.2`? It's a major version so it would make sense. It's also completely possible another dependency we have is conflicting, I just thought I'd check to see if you've seen this error in the past.

majapw:
@codylawson welp! it's probably this change - https://github.com/airbnb/react-dates/pull/1298

Ohhhh okay I understand. You need to add `dayPickerHorizontalPadding: 9` to your spacing theme key. Otherwise, it's undefined so when the calendarmonthwidth is calculated... it is also undefined. I would recommend spreading on the datepicker theme BUT with the hierarchy that's difficult. Maybe we should add a `buildTheme` method that would do the deep merge for you.

majapw:
To follow up, https://github.com/airbnb/react-dates/blob/master/src/theme/DefaultTheme.js#L156 got added and is now turning your width into NaN. I could also probably do a quick patch that would not cause this bad of a breakage when that's undefined. I will do that!

majapw:
[...]

PR #1355, opened 2018-09-13
Address missing daypicker width when calendarMonthPadding theme variable is undefined

majapw:
Fixes https://github.com/airbnb/react-dates/issues/1352

We added a new theme variable (`horizontalMonthPadding`) in v17.2.0. Unfortunately, when it was undefined, the effect was a trashfire (the daypicker width would be set to `NaN`). This PR addresses that edgecase and offers a fix!

to: @ljharb @noratarano @codylawson


ljharb:
tbh I'm pleased with myself that I didn't know that off the top of my head

PR #1360, opened 2018-09-14
Add SingleDatePickerInputController component

majapw:
This is the first step to creating a component structure that will allow consumers to asynchronously load the picker.

Basically, this creates an input controller component similar to the DateRangePickerInputController that takes care of everything related to input and input validation. There should be no functional change to any of the main components. 

to: @ljharb @lencioni @noratarano

### DIFF
--- a/src/components/SingleDatePicker.jsx
+++ b/src/components/SingleDatePicker.jsx
@@ -10,8 +10,6 @@ import OutsideClickHandler from 'react-outside-click-handler';
 import SingleDatePickerShape from '../shapes/SingleDatePickerShape';
 import { SingleDatePickerPhrases } from '../defaultPhrases';
 
-import toMomentObject from '../utils/toMomentObject';
-import toLocalizedDateString from '../utils/toLocalizedDateString';
 import getResponsiveContainerStyles from '../utils/getResponsiveContainerStyles';
 import getDetachedContainerStyles from '../utils/getDetachedContainerStyles';
 import getInputHeight from '../utils/getInputHeight';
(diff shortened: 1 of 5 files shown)


src/components/SingleDatePicker.jsx
```diff
@@ -513,7 +466,7 @@ class SingleDatePicker extends BaseClass {
             {...css(styles.SingleDatePicker_closeButton)}
             aria-label={phrases.closeDatePicker}
             type="button"
-            onClick={this.onClearFocus}
+            onClick={this.onOutsideClick}
```
ljharb: do you mean to use `onOutsideClick` here?

src/components/SingleDatePicker.jsx
```diff
@@ -604,7 +559,7 @@ class SingleDatePicker extends BaseClass {
         )}
       >
         {enableOutsideClick && (
-          <OutsideClickHandler onOutsideClick={this.onClearFocus}>
+          <OutsideClickHandler onOutsideClick={this.onOutsideClick}>
```
ljharb: also here?

src/components/SingleDatePickerInput.jsx
```diff
@@ -88,8 +88,6 @@ const defaultProps = {
   phrases: SingleDatePickerInputPhrases,
 };
 
-/* eslint react/no-this-in-sfc: 1 */
```
ljharb: lol thank you

src/components/SingleDatePicker.jsx
majapw: Yup! I changed the naming of the `onClearFocus` method to `onOutsideClick` so it would match the nomenclature of the `DateRangePicker` component, but the method itself does the same thing.

src/components/SingleDatePicker.jsx
majapw: Yup! I changed the naming of the `onClearFocus` method to `onOutsideClick` so it would match the nomenclature of the `DateRangePicker` component, but the method itself does the same thing.

src/components/SingleDatePickerInput.jsx
majapw: Lol, I think this was just... left over from a refactor. :O

src/components/SingleDatePicker.jsx
ljharb: I meant the variable instead of the instance method

majapw: Ah, no I def meant the instance method in both places

PR #1363, opened 2018-09-18
Address DayPickerNavigation rerender

majapw:
Fix for the DayPickerNavigation rerender discussed in https://github.com/airbnb/react-dates/issues/1297

Pretty simple fix involving moving around some arguments.

to: @ljharb @lencioni @AntiFish03

### DIFF
--- a/src/components/DayPicker.jsx
+++ b/src/components/DayPicker.jsx
@@ -207,7 +207,9 @@ class DayPicker extends BaseClass {
     this.onKeyDown = this.onKeyDown.bind(this);
     this.throttledKeyDown = throttle(this.onFinalKeyDown, 200, { trailing: false });
     this.onPrevMonthClick = this.onPrevMonthClick.bind(this);
+    this.onPrevMonthTransition = this.onPrevMonthTransition.bind(this);
     this.onNextMonthClick = this.onNextMonthClick.bind(this);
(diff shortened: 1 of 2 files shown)


src/components/DayPicker.jsx
```diff
@@ -446,7 +446,7 @@ class DayPicker extends BaseClass {
     }
   }
 
-  onPrevMonthClick(nextFocusedDate, e) {
+  onPrevMonthClick(e, nextFocusedDate) {
```
ljharb: it's a pretty strong convention to have `e` be last; why is this breaking change needed?

majapw: The motivation for this was because we were creating a new function in every render and causing unnecessary updates as a result. Flipping the order of arguments was a pretty simple solution, and was motivated by the fact that both arguments are occasionally omitted. 

We could alternatively cache the function with the event as the second arg as well if you think it makes more sense @ljharb. It may make sense to have `onNextMonthClick` take the event as an argument and call a function like `transitionToNextMonth` that can also be called directly. 🤔 Actually, that may be less confusing in the long run

majapw:
@ljharb I've split out the two cases that call the month transition into two functions. I think this approach makes more sense! Can you take another look?

PR #1365, opened 2018-09-18
[WIP] Add `DropDownController` component

majapw:
Part 1: https://github.com/airbnb/react-dates/pull/1360

This is part of some continued efforts to rearchitect the library allow for asynchronous loading of the `DayPicker` tree. 

The diff is quite big, but the general gist of the PR is to create an in-between layer of the component tree called the `DropDownController`. In the future, both the `DateRangePicker` and the `SingleDatePicker` (and any other pickers we can imagine) will leverage this component. 

The current architecture transforms the `SingleDatePicker` into a much simpler component that basically does:
```
function SingleDatePicker(props) {
  return (
    <DropDownController
      {...props}
      input={SingleDatePickerInputController}
      inputProps={SingleDatePickerInputControllerProps}
      dropdown={DayPickerSingleDateController}
      dropdownProps={DayPickerSingleDateControllerProps}
    />
  );
}
```

The `DropDownController` handles all the shared input/picker logic like focus state between the input and the dropdown, rendering portals appropriately, responsively styling the picker, styling the input and picker containers, showing/hiding the keyboard shortcuts panel, etc. 

The thing I'm doing now that I'd love some guidance on is with the props. There are some challenges in this area because:
- there are props that are unique to the SDP or the DRP in addition to those that are shared
- there are props that are controlled by the `DropDownController` that need to be spread onto the input and picker directly

You can see in the first commit that my initial approach was to have the `DropDownController` only take the props that it uses directly and to have the `SingleDatePicker` pass down nodes for the input and dropdown. That was much simpler as far as the code (and propTypes went), *BUT* because the `DropDownController` had to spread on props like `isDayPicker/InputFocused` and some handlers, it required cloning the element which was such a performance disaster that it made a difference on my high-powered macbook pro. :| 

Anyways, any advice on a good approach would be appreciated! 

@lencioni @ljharb @noratarano @amhunt


amhunt:
Hmmm - could we use objects to contain the props for each of these different types of props (e.g. shared props for SDP/DRP, additional props for SDP, additional props for DRP, props for the dropdown component itself).

I see that you're currently creating objects like this within the component, but could we just have the user pass in those objects instead?

majapw:
The only challenge with that approach @amhunt is that if we did

src/components/DropDownController.jsx
```diff
@@ -0,0 +1,501 @@
...
+        {!enableOutsideClick && this.maybeRenderDayPickerWithPortal()}
+      </div>
+    );
+  }
+}
+
+DropDownController.propTypes = propTypes;
+DropDownController.defaultProps = defaultProps;
+
+export { DropDownController as PureDropDownController };
```
ljharb: this is an antipattern we should avoid; can `.dive()` be used instead?

src/components/DropDownController.jsx
```diff
@@ -0,0 +1,501 @@
...
+  customCloseIcon: PropTypes.node,
+  disableScroll: PropTypes.bool,
+  focused: PropTypes.bool,
+  horizontalMargin: nonNegativeInteger,
+  isRTL: PropTypes.bool,
+  keepFocusOnInput: PropTypes.bool,
+  onFocusChange: PropTypes.func,
+  openDirection: openDirectionShape,
+  orientation: orientationShape,
+  phrases: PropTypes.object, // only the closeDatePicker phrase is used
```
ljharb: `PropTypes.shape({ closeDatePicker: PropTypes.string.isRequired })`?

package.json
```diff
@@ -23,7 +23,7 @@
     "pretests-karma": "npm run react",
     "tests-karma": "karma start",
     "test": "npm run build && npm run tests-only",
-    "storybook": "start-storybook -p 6006",
+    "storybook": "NODE_ENV=development start-storybook -p 6006",
```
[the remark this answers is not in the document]
majapw: I wasn't getting proptype errors in storybook anymore (I think due to Joe's PR that removed them in production... so it seemed that the answer is no!

src/components/DropDownController.jsx
majapw: A lot of this file was simply pulled out from the `SingleDatePicker`. The reason for this pattern is the same as it was previously there---which is that we're spying on and stubbing instance methods. Unfortunately, dive does not have a good story for that case.

PR #1382, opened 2018-09-29
Integrate with Happo.io

trotzig:
Integrate with Happo.io

Happo.io is a screenshot testing service. It is used to find bugs,
visualize change and provide a visual log of your application's
components. It can be used to take screenshots in different browsers.

I've been using react-dates for a while when stress-testing happo. One
of the things that have been most tricky is that react-dates relies on
measuring elements in the DOM as part of rendering the UI. This is an
issue for default Happo, because it uses JSDOM to prerender components,
then collects the HTML and CSS and sends it to the happo.io to take
screenshots. JSDOM can't measure things. The pre-rendering phase is
important for speed and security, and is something that would be less
than trivial to bypass.  So, instead of bypassing pre-rendering I've
made it pre-render in a real browser. The `happo-plugin-puppeteer` npm
module injects a puppeteer-powered (uses Chrome) pre-renderer to Happo.
And by using that module, we're good to go for react-dates.

Another important module here is `happo-plugin-storybook`. This plugin
converts all storybook stories to happo examples. This means we can keep
using Storybook and we'll get screenshot testing "for free". I picked
the .storybook-css project and disabled a few decorators/addons in the
storybook config.

For now, we'll be taking screenshots in Chrome in two viewports:

- 900x600 ("desktop")
- 320x640 ("mobile")

Going forward, we may choose to add other targets/browsers as well. But
to begin with, I want to make sure things are working smoothly before we
expand.


trotzig:
Here's what the current set of screenshots look like: https://happo.io/a/32/report/dev-267bf0c808a57df525b5 . There are a few known issues, some of which I'll work on fixing upstream and some that I think we'll just have to live with: 
- Examples in non-english locales don't look right. This is a happo.io bug, I'll look into that soon.
- Some examples require a user action (like a click) to reveal more. I thought about auto-clicking and such but decided it wasn't worth it.  There are plenty of examples of the date picker being shown anyway. 

In order for this to get integrated with CI, I'm going to need help with a few things: 
- Configure happo.io API keys in Travis
- Add happo to the Travis run
- Install the [Happo Github app](https://github.com/apps/happo) for the repo

trotzig:
It looks like I also need help preventing failures in Node 4: https://travis-ci.org/airbnb/react-dates/builds/435084746?utm_source=github_status&utm_medium=notification. Happo doesn't support Node 4, and a dependency we use (puppeteer) is failing to install.

ljharb:
You might have to make puppeteer an optional dependency then :-/ binary deps are a big hassle, i'm afraid.

trotzig:
@ljharb I made `happo-plugin-puppeteer` an optional dependency here and it seems to do the right thing (ignore puppeteer install failures). Let me know if you think that's the wrong approach. And yeah, binary dependencies are a hassle.

package.json
```diff
@@ -136,6 +135,9 @@
     "react-dom": "^0.14 || ^15.5.4 || ^16.1.1",
     "react-with-direction": "^1.3.0"
   },
+  "optionalDependencies": {
```
ljharb: ah, i misunderstood :-/ this would make it an optional runtime dep of react-dates - there’s no optional dev deps. The only way to make this work is to either a) have it be an optional dep of another package that is a dev dep here, or b) don’t put it in package.json at all and manually install it in travis.yml, or c) find an alternative package

trotzig: Looks like that didn't work...I'll probably remove it from package.json then and just install it when it's needed.

trotzig:
Okay, I think I'm at a state where all I need is help
- Install the [Happo Github app](https://github.com/apps/happo) for the repo

.travis.yml
```diff
@@ -18,9 +18,12 @@ script:
   - 'if [ "${TEST-}" = true ]; then npm run tests-only ; fi'
   - 'if [ -n "${KARMA-}" ]; then npm run tests-karma ; fi'
   - 'if [ -n "${COVERAGE-}" ] && [ "${TRAVIS_BRANCH-}" = "master" ]; then npm run cover; cat ./coverage/lcov.info | ./node_modules/.bin/coveralls ; fi'
+  - 'if [ -n "${HAPPO-}" ]; then npm install --no-save happo-plugin-puppeteer && INSTALL_CMD="npm install && npm install --no-save happo-plugin-puppeteer" npm run happo-ci ; fi'
```
ljharb: slightly more concise? `INSTALL_CMD="npm install && npm install --no-save happo-plugin-puppeteer" && $INSTALL_CMD && npm run happo-ci`?

.travis.yml
```diff
@@ -18,9 +18,12 @@ script:
...
 env:
   global:
     - TEST=true
+    - secure: "lv5bnKRHd3A0tUequn7CrpKLkaZjVc1YmY/MzvX53wjqIwP4llpIil9XqeVYoG/VDGcf810XktZe2QeAuFUTXIZhPGQRXcWsbKbHVNmKSFQFDtXqEpB9NGw4NBCytK2ZLVXhRfrIcHm6Ltr0CSa+mO/KHtk9bE/on/Itjc0aB5b7Y1gVVHTfyZ3QOjIsPUenLRsY1tn1rUJ0meFlogPFIBbcnK+GPDL2oHsiVIOjCfBauIcRlNP7XZggFhZbHARZG7bTDr+YBuybAIBCVnex5Av7ho5EHu8FTjNrGXZyKF5IOMjqJla8UvOAoEnNe5k98O+Wx9ehlObUYGcr9szV9zGHvp0Wo+sflHdWYy5IOQAp+AQqjIz8J18ia6q0Zm9TBL5h2OEx/XwKzp4Q+F0XeTL7USnsfLuVdRBfEi4fOq/ucXnmOOJNJnBl8Q1IW5NYwF0nL1sW2jNVnQsp6a0sjLs0QWHEs1J0OMyru0vsWCJfkFQn1ozla62EzKClNW0dAvRJcHsOCD2JWCzToM3HysghPkT1jAZvfrW1ndBDYCipetgSx9AoYb2jYxrxFx3BoZBd23zIy5WDa0V/HkwijV04GEt0/V/rXF3pQzEaKE07sY8rjeTHNew/bmmNcid3/MsPv/8596j6ir4QmuxaHi/kP5ZTQWjtjytUADB7HqU="
```
ljharb: can you elaborate on what these secure env vars are called, and what their purpose is? if we need to update them later, how can that be done?

.travis.yml
```diff
@@ -37,6 +40,8 @@ matrix:
       env: COVERAGE=true TEST=false
     - node_js: "lts/*"
       env: KARMA=true TEST=false
+    - node_js: "lts/*"
+      env: HAPPO=true TEST=false
```
ljharb: do we want this to be an allowed failure, or will it always robustly complete?

package.json
```diff
@@ -25,6 +25,8 @@
     "test": "npm run build && npm run tests-only",
     "storybook": "start-storybook -p 6006",
     "storybook:css": "npm run build:css && start-storybook -p 6006 -c .storybook-css",
+    "happo": "npm i --no-save happo-plugin-puppeteer && happo",
```
ljharb:
```
"prehappo": "npm i --no-save happo-plugin-puppeteer",
"happo": "happo"
```
?

package.json
```diff
@@ -25,6 +25,8 @@
...
+    "happo-ci": "happo-ci-travis",
```
ljharb: why is this a separate binary? what does it do differently?

ljharb:
The github app's been activated for this repo.

package.json
trotzig: The first command will create a single report (set of screenshots) and can be run locally as well as in CI. The second command is designed to run in CI. It will (if needed) check out the commit that the PR is based on (usually a commit on master), run the first command, then check out the tip of the PR, run the first command again. As a final step, it will compare the two reports which will create a diff between the two commits.

.travis.yml
trotzig: Ah, good callout. I meant to do this but forgot. I'll comment inline.

.travis.yml
trotzig: Done!

.travis.yml
trotzig: I added the name of the variable in .travis.yml, then used the commit message to explain things. Anyone ending up git-blaming these should land in this commit: 7c4e388.

package.json
trotzig: TIL! I'll make this change.

.travis.yml
trotzig: I had to roll that back. I tried multiple versions of this but I couldn't get it right.

trotzig:
Alright, I'm at a place now where I'm ready for a review. Sorry for all the back-and-forth and force-pushes, I had to do some trial-and-error to get a few things to work. While doing so, I found a bug in the happo.io client (first-time run didn't work as expected). I've also noted down a few things I could improve on: 
- actually use `happo` script in package.json command (basically revert https://github.com/happo/happo.io/commit/84aff08dce30633009e5656cb62bf8d41451f8d4#diff-afc9d820011be36622385f5147e4c8c4)
- Provide better error messages when puppeteer fails to launch (right now you get a "can't call `close` of undefined").

.travis.yml
trotzig: I added the name of the variable in .travis.yml, then used the commit message to explain things. Anyone ending up git-blaming these should land in this commit: ecb728a.

trotzig:
@lencioni working on it!

trotzig:
It looks like the regular `.storybook` setup can't be built statically. Running this on the master branch currently fails:

```bash
$(npm bin)/build-storybook
```

I can't make much of the error message: 
```
ERR! ./examples/DateRangePickerWrapper.jsx
ERR! Module build failed (from ./node_modules/babel-loader/lib/index.js):
ERR! TypeError: Property declarations[0] of VariableDeclaration expected node to be of a type ["VariableDeclarator"] but instead got "ExpressionStatement"
ERR!     at validate (/home/travis/build/airbnb/react-dates/node_modules/@babel/types/lib/definitions/utils.js:128:13)
ERR!     at validator (/home/travis/build/airbnb/react-dates/node_modules/@babel/types/lib/definitions/utils.js:97:7)
ERR!     at Object.validate (/home/travis/build/airbnb/react-dates/node_modules/@babel/types/lib/definitions/utils.js:172:7)
ERR!     at validate (/home/travis/build/airbnb/react-dates/node_modules/@babel/types/lib/validators/validate.js:17:9)
ERR!     at builder (/home/travis/build/airbnb/react-dates/node_modules/@babel/types/lib/builders/builder.js:46:27)
ERR!     at Object.VariableDeclaration (/home/travis/build/airbnb/react-
```
(full log here: https://travis-ci.org/airbnb/react-dates/jobs/541444935)

It looks like others are running into this as well: https://github.com/airbnb/react-dates/pull/1595#issuecomment-492701463

@lencioni any chance you can look into that?

trotzig:
With a bit of git bisecting, the commit introducing the above mentioned error is ae3b1a68639d1d1df600b8eb469d2fb6f7f9e319. This commit makes a number of package updates, among others `@babel/core`. I'm going to see if bumping that dependency can help.

trotzig:
Just to be clear, running storybook via `npm run storybook` does not cause the error to happen. 

The failure happens right around when something like this is in the logs:
```
92% Chunk asset (something something) TerserPlugin
```

(The console is overwritten when the error appears, so this is from memory). 

This would explain why the regular dev command would work but the static build doesn't (TerserPlugin is doing minification which is most likely ignore in dev mode).

ljharb:
I'd expect happo's webpack build to run in NODE_ENV development, not production, which i would expect not to include minification.

trotzig:
@ljharb It does, but the happo-plugin-storybook plugin avoids the happo webpack build and uses a static storybook build instead. 

I've got things working with Storybook version 5. Is there anything you see stopping us from getting on to the latest storybook version? If not I'll push a separate PR with that change.

ljharb:
As long as it works, and doesn't force any webpack or babel or react versions (like if it makes `npm ls` error out on react < 16, for example, then we couldn't test on that react version properly), then a PR sounds great!

trotzig:
Sorry for all the force-pushing. The last one was rebasing on master again after the Storybook v5 upgrade was merged. 

I also had to resolve an issue with Storybook v5 and its usage of core-js. For anyone reviewing, it might be worth taking a look at 15611832b01a2d298789af4c004 in isolation (lengthy commit message) and the [Storybook issue](https://github.com/storybookjs/storybook/issues/6204) that's related to the fix.

trotzig:
@lencioni did you try accepting the happo diff as well. You should (theoretically) have access, as long as you have write-access to the github repo (`react-dates`).

lencioni:
@trotzig I did, but it didn't seem to change the status on github. It looks like when I select Accept, I get a 401 response from the API request, but the UI doesn't indicate any problems.

package.json
```diff
@@ -70,6 +72,7 @@
     "babel-preset-airbnb": "^3.2.1",
     "chai": "^4.2.0",
     "clean-css": "^4.2.1",
+    "core-js": "^3.1.3",
```
ljharb: why is this needed?

trotzig: The commit message has the details: https://github.com/airbnb/react-dates/commit/15611832b01a2d298789af4c0049c96d2b794cfa

ljharb: The implication is that by using storybooks webpack config we have created this problem for ourselves. should that be reverted?

trotzig: I think there’s no real way out of using storybook’s webpack config. And just to be clear, this isn’t a happo issue - it just happened to trigger the issue.

Perhaps I’m misunderstanding something though, can you elaborate on what you mean should be reverted?

ljharb: why must we use their config, versus manually replicating the parts we need (since the linked issue is due to weird aliasing in it)?

trotzig: The alias isn’t there anymore, tried removing it. A theory I have is that they’re generating some code that ends up being bundled with webpack. At least that would explain why a `require('core-js')` wouldn’t resolve to the local copy of core-js (storybook/core/node_modules/core-js). If I have some time I’ll explore this theory and try to push a fix upstream.

PR #1384, opened 2018-10-01
Make sure modifiers recomputation uses noon instead of midnight.

### LINKED ISSUE #1358: isOutsideRange / isDayBlocked styling has a 1 day gap on refresh while the picker is open
Hello there,

I am refreshing my blocked days (using `isOutsideRange`, but `isDayBlocked` has the same issue) on `focusedInput` change (because I'm using daily availabilities and changeover days), and on refresh, the click handlers are correctly set, but the styles have a 1-day offset.
Closing the picker and reopening it focusing immediately the end date input makes it show up correctly.

react-dates_blocked_day_offset_issue

On this GIF, my change over days forbid departure on Mondays and Tuesdays.
You can notice that Monday the 12th show up free (wrong behavior) but cannot be selected (right behavior).
Also, Wednesday the 14th show up blocked (wrong behavior) but can be selected (right behavior).
I also demonstrate the closing/reopening whigh correctly updates everything.

Here's how I implemented my isOutsideRange method:
```js
  checkPickerOutsideRange = day => {
    const pricingIndex = this.listing.prices.findIndex(p => moment.range(p.startDate, p.endDate).contains(day))

    if (pricingIndex === -1) {
      return true
    }

    const { startDate } = this.listing.prices[pricingIndex]
    const allFutureAvailabilities = this.listing.prices.slice(pricingIndex).map(p => p.availability).join('')
    const allFutureChangeOvers = this.listing.prices.slice(pricingIndex).map(p => p.changeOverDays).join('')
    const currentDayIndex = moment.range(startDate, day).diff('days')

    if (this.state.focusedInput === START_DATE) {
      return allFutureAvailabilities.charAt(currentDayIndex) === 'U' ||
        allFutureChangeOvers.charAt(currentDayIndex) === 'O' ||
        allFutureChangeOvers.charAt(currentDayIndex) === 'X'
    } else if (this.state.focusedInput === END_DATE) {
      const nextBlockedDayOffset = allFutureAvailabilities.indexOf('U', moment.range(startDate, this.props.applicationStore.travelDates.startDate).diff('days'))
      const nextBlockedDay = moment(startDate).add(nextBlockedDayOffset, 'days')
      return day.isBefore(this.props.applicationStore.travelDates.startDate, 'day') ||
        day.isAfter(nextBlockedDay, 'day') ||
        allFutureChangeOvers.charAt(currentDayIndex) === 'I' ||
        allFutureChangeOvers.charAt(currentDayIndex) === 'X'
    }

    // Default is blocked, shouldn't be reached
    return true
  }
```

Have I done anything wrong or is that a bug ?
Thanks !

fmauNeko:
All other moments are set at noon in the components.

Closes #1358.

### DIFF
--- a/src/components/DayPickerRangeController.jsx
+++ b/src/components/DayPickerRangeController.jsx
@@ -381,7 +381,7 @@ export default class DayPickerRangeController extends BaseClass {
     if (didFocusChange || recomputePropModifiers) {
       values(visibleDays).forEach((days) => {
         Object.keys(days).forEach((day) => {
-          const momentObj = moment(day);
+          const momentObj = moment(day).hour(12);
           let isBlocked = false;
 
           if (didFocusChange || recomputeOutsideRange) {
--- a/test/components/DayPickerRangeController_spec.jsx
+++ b/test/components/DayPickerRangeController_spec.jsx
@@ -882,6 +882,17 @@ describe('DayPickerRangeController', () => {
             });
             expect(isOutsideRangeStub.callCount).to.not.equal(0);
           });
+
+          it('calls isOutsideRange with moments set at 12pm', () => {
+            const isNoon = sinon.match(day => day.hour() === 12);
+            const isOutsideRangeStub = sinon.stub();
+            const wrapper = shallow(<DayPickerRangeController {...props} />);
+            wrapper.instance().componentWillReceiveProps({
+              ...props,
+              isOutsideRange: isOutsideRangeStub,
+            });
+            expect(isOutsideRangeStub.alwaysCalledWithMatch(isNoon)).to.equal(true);
+          });
         });
 
         describe('focusedInput changed', () => {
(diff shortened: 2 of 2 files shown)


fmauNeko:
Hi! Here are some test, I hope I did it the right way.

ljharb:
@fmauNeko thanks - i think it'd be better, instead of testing "sets the time to noon", if the tests tested the actual real-world case that breaks when the time is set wrong.

fmauNeko:
I'm not sure I see how to test that, as the moments are created from inside the component themselves, from the component's state.

ljharb:
I was envisioning using enzyme to construct a component while mocking time to recreate the conditions under which the bug appears.

ljharb:
@fmauNeko looks like you deleted your fork, which is unfortunate; it means this PR can't ever land :-(

Issue #1394, opened 2018-10-11
adjustDayPickerHeight not calculating predictably when setting numberOfMonths prop dynamically


LMulvey:
On our project, we're using a grid system that optionally passes us a `screenClass` prop to tell us what breakpoint we're currently in. We're using that breakpoint to pass a `numberOfMonths` prop to the `DayPickerRangeController` to render either two months for desktop or one month for mobile. 

When we're doing this, our function initializes as `sm`, which initially passes `numberOfMonths={1}`. It's changed to `numberOfMonths={2}` shortly after once `window.innerWidth` is captured and the method is run. 

Setting it to 1 on initial load breaks `adjustDayPickerHeight` for a month either 2 or 3 greater than the current one. As it's calculated based off of `calendarMonthWeeks`, which is only set on `componentDidMount` right now, that value becomes stale until it's triggered again.

![adjustmonthheightissues](https://user-images.githubusercontent.com/16218249/46836493-ecef4300-cd6e-11e8-8b77-1f8563a6dd29.gif)
## Possible Fix

Adding a check to `componentDidUpdate` for a variation in `numberOfMonths` and then triggering `setCalendarMonthWeeks` again fixes the issue.

haocow:
I am also running into this issue currently on v21.8.0

LMulvey:
@haocow I put in a PR three years ago that fixed the problem but unfortunately the reviewer wanted me to fix their entire test suite in addition to the problem I identified. I never had time to do that, nor do I agree that I should have had to do that. We ended up going in a different direction because of it as it was a blocker for us.

Feel free to review #1395 and see if that would still apply to this library–you may be able to fix it on a local fork or attempt another PR.

ljharb:
@LMulvey the repo is still maintained; the latest commit was *authored* in 2019, but it was merged 29 days ago (if you click on the commit, and to the linked PR, you'll see that).

As for #1395, I don't think that's a fair characterization of what I asked you, and there were no updates after https://github.com/airbnb/react-dates/pull/1415#issuecomment-430089543 to respond to.

LMulvey:
@ljharb You're right on the commit part for sure, apologies for that. As for #1395, perhaps I misunderstood your comment but my PR involved adding updated tests to reflect the fix I had made. This, however, revealed that an entire test block had been skipped for some time and was not functional. Your comment sounded like you were asking me to fix that test suite before my PR could be merged–something that I did not have the capacity to do at the time. 

[...]

PR #1395, opened 2018-10-11
Add prop.numberOfMonths comparison into componentDidUpdate and run this.setCalendarMonthWeeks if prop has changed.

LMulvey:
This change addresses #1394 to ensure that `adjustDayPickerHeight` is always being called with fresh data from `this.calendarMonthWeeks`.

In our project, we're using a function to return a `screenClass` depending on the current value of `window.innerWidth`. We use `screenClass` to dynamically set the `numberOfMonths` prop. Before `window.innerWidth` is set, `screenClass` returns `sm`, which causes the _initial_ value of `numberOfMonths` to be set to `1` and then immediately changed to `2`.

```
<ScreenClassRender
   render={screenClass => (
      <DayPickerRangeController
        {...props}
        numberOfMonths={
          ['md', 'lg', 'xl'].includes(screenClass) ? 2 : 1
        }
        onDatesChange={this.onDatesChange}
        focusedInput={focusedInput}
        startDate={startDate}
        endDate={endDate}
      />
    ) }
/>
```

Since `this.setCalendarMonthWeeks` was called in `componentDidMount` and not again when the prop changes, the values in `this.calendarMonthWeeks` were no longer fresh.

**Before fix:**
adjustmonthheightissues

**After fix:**
issuesfixed

I totally get that this may be a complete edge case and as a first-time contributor, I'm very open to any changes/criticism/etc with this.

Thanks!

### DIFF
--- a/test/components/DayPicker_spec.jsx
+++ b/test/components/DayPicker_spec.jsx
@@ -12,6 +12,7 @@ import CalendarMonthGrid from '../../src/components/CalendarMonthGrid';
 import DayPickerNavigation from '../../src/components/DayPickerNavigation';
 import DayPickerKeyboardShortcuts from '../../src/components/DayPickerKeyboardShortcuts';
 import {
+  DAY_SIZE,
   HORIZONTAL_ORIENTATION,
   VERTICAL_ORIENTATION,
   VERTICAL_SCROLLABLE,
@@ -22,8 +23,9 @@ const today = moment().locale('en');
 const event = { preventDefault() {}, stopPropagation() {} };
 
 describe('DayPicker', () => {
+  let adjustDayPickerHeightSpy;
   beforeEach(() => {
-    sinon.stub(PureDayPicker.prototype, 'adjustDayPickerHeight');
+    adjustDayPickerHeightSpy = sinon.stub(PureDayPicker.prototype, 'adjustDayPickerHeight');
   });
 
   afterEach(() => {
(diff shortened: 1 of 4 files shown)


test/components/DayPicker_spec.jsx
```diff
@@ -790,11 +791,14 @@ describe('DayPicker', () => {
...
     describe('#componentDidUpdate', () => {
       let updateStateAfterMonthTransitionSpy;
+      let setCalendarMonthWeeksSpy;
+
       beforeEach(() => {
         updateStateAfterMonthTransitionSpy = sinon.stub(
           DayPicker.prototype,
           'updateStateAfterMonthTransition',
         );
+        setCalendarMonthWeeksSpy = sinon.stub(PureDayPicker.prototype, 'setCalendarMonthWeeks');
```
ljharb: If you unskip the containing describe, do your new tests pass?

LMulvey: Ah! They don't 👎 In fact, none of the tests in that section are passing (I didn't even notice the skip). It looks like it's been skipped for about two years now.

I can take it on and go back and fix all those tests. Should I fix mine for this PR and make a separate PR for fixing the other tests? Or lump it all into this one?

ljharb: I don't mind if your new test happens to be skipped; but it's important that the new test *would* be passing, so that when we unskip this section, we don't have to go fix it :-)

LMulvey: The issue here is that _all_ the lifecycle tests are being skipped because  `componentDidUpdate` isn't being called without fully mounting the component. Full mount isn't working right now and is throwing: `Error: It looks like you called `mount()` without a global document being loaded.`

I'm happy to implement something like `jsdom-global` to allow mounting to work with the setup but I'm wondering if that should be a separate PR from this one? Or if I should take another approach to testing these methods? I could obviously stub it out but that seems intuitive to me as that wouldn't _really_ be mimicking production.

ljharb: If you could enable these tests in another PR, we could merge that prior to this one - that’d be great!

LMulvey: #1415 addresses this - @ljharb

haojonat:
Hi, I have created a new pr here: https://github.com/airbnb/react-dates/pull/2095

PR #1417, opened 2018-10-14
Update to babel 7

ljharb:
Updates react-dates to babel 7.

Closes #1389. Closes #1346. Closes #1416. Closes #1390. Closes #1377. Closes #1378.


src/components/DayPickerNavigation.jsx
```diff
@@ -122,21 +122,21 @@ function DayPickerNavigation({
           {...css(
             styles.DayPickerNavigation_button,
             isDefaultNavPrev && styles.DayPickerNavigation_button__default,
-            ...(isHorizontal && [
+            ...(isHorizontal ? [
               styles.DayPickerNavigation_button__horizontal,
-              ...isDefaultNavPrev && [
+              ...(isDefaultNavPrev ? [
```
majapw: well this is annoying. can I no longer spread on falsey things?

ljharb: apparently you never could for non-objects, but Babel mistakenly allowed it.

https://github.com/babel/babel/pull/7491

majapw: weird okay!

pedroabreu:
This update seems to break storybook locally, as it can't find `babel-core` (replaced with `@babel/core`)

Opened issue here https://github.com/airbnb/react-dates/issues/1429

PR #1451, opened 2018-11-05
Fix incorrect VO for selected check-in date

nkinser:
This PR fixes an accessibility bug where VO incorrectly defines a selected check-in date in calendar as unavailable. This is fixed by checking if the day is both blocked and selected, and if this is true the phrase should be `dateIsSelected`.

### DIFF
--- a/src/utils/getCalendarDaySettings.js
+++ b/src/utils/getCalendarDaySettings.js
@@ -35,7 +35,9 @@ export default function getCalendarDaySettings(day, ariaLabelFormat, daySize, mo
   const formattedDate = { date: day.format(ariaLabelFormat) };
 
   let ariaLabel = getPhrase(chooseAvailableDate, formattedDate);
-  if (modifiers.has(BLOCKED_MODIFIER)) {
+  if (modifiers.has(BLOCKED_MODIFIER) && selected) {
+    ariaLabel = getPhrase(dateIsSelected, formattedDate);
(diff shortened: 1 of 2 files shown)


nkinser:
@majapw Can you take a look at this? Thanks!

src/utils/getCalendarDaySettings.js
```diff
@@ -35,7 +35,9 @@ export default function getCalendarDaySettings(day, ariaLabelFormat, daySize, mo
   const formattedDate = { date: day.format(ariaLabelFormat) };
 
   let ariaLabel = getPhrase(chooseAvailableDate, formattedDate);
-  if (modifiers.has(BLOCKED_MODIFIER)) {
+  if (modifiers.has(BLOCKED_MODIFIER) && selected) {
+    ariaLabel = getPhrase(dateIsSelected, formattedDate);
+  } else if (modifiers.has(BLOCKED_MODIFIER)) {
     ariaLabel = getPhrase(dateIsUnavailable, formattedDate);
   } else if (selected) {
```
majapw: would it make sense to just move the `if(selected)` condition to be fire? e.g.
```
if (selected) {
    ariaLabel = getPhrase(dateIsSelected, formattedDate);
  } else if (modifiers.has(BLOCKED_MODIFIER)) {
    ariaLabel = getPhrase(dateIsUnavailable, formattedDate);
  } else ...
```

nkinser:
@majapw PTAL! Thanks!

Issue #1354, opened 2018-09-12
[IE10] `Cannot call a class as a function` console error


majapw:
I've made a simple demo app here: https://github.com/majapw/react-dates-demo/pull/3

It appears that our `baseClass` (https://github.com/airbnb/react-dates/blob/master/src/utils/baseClass.js) errors out in IE10 when an HOC-wrapped component extends it. It does not cause any problems when a non-wrapped component extends it. 

to: @ljharb @noratarano @lencioni @AntiFish03

majapw:
The second commit (which inlines the check instead of double-extending) does not error out:
https://github.com/majapw/react-dates-demo/pull/3/commits/d79813cdf3a8576e954a9f0dee4ad8312bea9305

ljharb:
I believe this is because react-with-styles (along with almost every HOC) hoists statics, but does not preserve inheritance (https://github.com/airbnb/react-with-styles/blob/master/src/withStyles.jsx#L23-L33) - I don't think that's something that an HOC could or should reliably solve.

I think the solution is to find an ergonomic way to use the inlined implementation, rather than creating an inheritance hierarchy.

majapw:
Is the only real answer to this a babel plugin? I've gone my whole life without writing a babel plugin and I was really hoping to maintain that streak.

I guess this could be a `PureComponent` polyfill of sorts? Does that make sense?

ljharb:
There's really these options:
 1. revert the change, and keep the manual sCU until we're ready to drop support for React < 15.3.
 1. drop support for React <15.3, and use PureComponent, and direct people to use a PureComponent polyfill (which would require us to locate/vet or create one)
 1. manually inline the gross code everywhere
 1. write a babel transform that converts PureComponents to the gross version, and then we can drop that transform when we want to drop support for React < 15.3 (and we could still recommend the polyfill for those people now)

majapw:
ugggghhhh. okay i'm going to revert the change and publish an 18.0.3 and then republish in 18.0.4 so that I can update our internals and we can WRITE A BABEL TRANSFORM I GUESS. :|

PR #1452, opened 2018-11-06
Add PureComponent fallback babel transform

majapw:
Fix for https://github.com/airbnb/react-dates/issues/1354

Never got around to doing this before and now it is blocking.

Thanks @skevy for writing the transform!

to: @ljharb @skevy @Sumeet-Jain


.babelrc
```diff
@@ -7,7 +7,8 @@
           "svgo": false
         }],
         ["transform-replace-object-assign", { "moduleSpecifier": "object.assign" }],
-        "istanbul"
+        "istanbul",
```
ljharb: i think the istanbul transform needs to come last, so it can properly instrument the code for coverage

majapw:
@ljharb @skevy i think this makes sense now!

scripts/pure-component-fallback.js
```diff
@@ -0,0 +1,67 @@
+module.exports = function pureComponentFallback(babel) {
+  const { types: t, template } = babel;
```
ljharb: nbd but this can be destructured in the previous line

scripts/pure-component-fallback.js
```diff
@@ -0,0 +1,67 @@
...
+      t.blockStatement([template('return shallowCompare(this, nextProps, nextState);')()]),
+    );
+    return method;
+  };
+
+  function superclassIsPureComponent(path) {
+    const superclass = path.get('superClass');
+    // check for PureComponent
+    if (t.isIdentifier(superclass)) {
+      return superclass.node.name === 'PureComponent';
```
ljharb: probably fine for "just this repo", but if we'll want to use this in other projects, we should make sure that `PureComponent` is actually a named import from `'react'` or a destructure from React imported/required from `'react'`

scripts/pure-component-fallback.js
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    if (t.isMemberExpression(superclass)) {
+      // Check for React.PureComponent
+      return superclass.get('object').node.name === 'React'
```
ljharb: same here

src/components/CalendarDay.jsx
```diff
@@ -340,4 +338,4 @@ export default withStyles(({ reactDates: { color, font } }) => ({
   CalendarDay__today: {},
   CalendarDay__firstDayOfWeek: {},
   CalendarDay__lastDayOfWeek: {},
-}), { pureComponent: pureComponentAvailable })(CalendarDay);
+}), { pureComponent: typeof React.PureComponent !== 'undefined' })(CalendarDay);
```
ljharb: not a blocker, but it'd be nicer if the transform could add this instead

scripts/pure-component-fallback.js
```diff
@@ -0,0 +1,67 @@
...
+  const buildPureOrNormalSuperclass = () => t.LogicalExpression('||',
+    t.memberExpression(t.identifier('React'), t.identifier('PureComponent')),
+    t.memberExpression(t.identifier('React'), t.identifier('Component')));
+
+  const buildShouldComponentUpdate = () => {
+    const method = t.classMethod(
+      'method',
+      t.identifier('shouldComponentUpdate'),
+      [t.identifier('nextProps'), t.identifier('nextState')],
+      t.blockStatement([template('return shallowCompare(this, nextProps, nextState);')()]),
```
ljharb: the trailing function comma here is breaking node 4.

PR #1430, opened 2018-10-26
Fix storybook for babel 7 by upgrading to v4

ljharb:
Fixes #1429.

PR #1453, opened 2018-11-06
Update the import of  the action addon in Controller storybook stories

philkeys:
These two stories were throwing an error on the action function. Updated the `action` import from `@storybook/react` to `@storybook/addon-actions`.


ljharb:
(Linking to #1430)

PR #1457, opened 2018-11-07
Conditionally apply the shouldComponentUpdate method

### LINKED ISSUE #1455: New Warning with last version 18.2.1
When I use your last DateRangePicker, I see in console this:
`DateRangePicker has a method called shouldComponentUpdate(). shouldComponentUpdate should not be used when extending React.PureComponent. Please extend React.Component if shouldComponentUpdate is used.` - `react-dom.development.js:506`

on every kind of interaction. I'm using last version of React and React-dom (16.6.1)

majapw:
Fixes #1455

This updates the PureComponent fallback babel transform to conditionally apply `shouldComponentUpdate` only in cases where `React.PureComponent` is not supported (otherwise it throws an error).

to: @skevy @ljharb

### DIFF
--- a/scripts/pure-component-fallback.js
+++ b/scripts/pure-component-fallback.js
@@ -6,9 +6,14 @@ module.exports = function pureComponentFallback({ types: t, template }) {
   const buildShouldComponentUpdate = () => {
     const method = t.classMethod(
       'method',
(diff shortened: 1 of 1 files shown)


majapw:
Verified that the console errors are gone!

PR #1468, opened 2018-11-16
Added clearTimeout for the setCalendarMonthGridHeight timeout

sheldonj:
I know that this might seem like a trivial PR, but in our application we are getting " Can't call setState (or forceUpdate) on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method." warnings because we aren't clearing the timeout.


sheldonj:
Hi @majapw is there any extra work you need to get this merged?

majapw:
Looks great! Mostly my difficulty in giving this repo the time it needs right now. :)

Issue #647, opened 2017-07-26
Option to control total rendered months.


ArtyomResh:
Hi, 
There is prop called "numberOfMonths", which is responsible for rendering passed amount of months at the same time, but I want to render the total amount of months and keep users in it. Right now you can manage to do something like this by hiding controls, but it is still possible to break out by [keyboard ](http://s.csssr.ru/U31J879TR/calendar.gif) and do calendar, not keyboard accessible is not an option. Props "onPrevMonthClick" and "onNextMonthClick" take an only event argument, so I can't achieve my goal through them either. 

Is there something I'm missing? 
Can I achieve what I want in the current realization of your library?

majapw:
Hey @ArtyomResh, I think this is somewhat a duplicate of https://github.com/airbnb/react-dates/issues/631

If we implemented that solution, we could use `isOutsideRange` to selectively disable the month navigation and that would solve your usecase.

So short answer, this isn't possible rn, but... we want to implement it.

ArtyomResh:
Good to know, thank you!

ArtyomResh:
Hi, @majapw! :) 
Is there any news about this issue? I've checked #631 and #657 but didn't found a solution for this.

majapw:
This has not yet been implemented, but I think we would accept a PR. Unfortunately, I do not currently have the bandwidth myself.

ljharb:
Does #1311 solve this, @ArtyomResh?

Issue #631, opened 2017-07-20
onNextMonthClick & onPrevMonthclick not passed down to DayPickerSingleDateController


brosewarne:
Hey,

First, thanks for an awesome component :) I've been using it for a few days and it's been much easier to use than other react datepickers I've tried.

My issue is that I'm trying to use a SingleDatePicker with restricted dates, which works great. However the user can still use the next / prev month navigation to go to months where there are no days available for them to select. 

I thought I'd be able to stop this by playing around with the onNextMonthClick and onPrevMonthClick functions, however these don't seem to be passed down to the DayPickerSingleDateController, so only the empty function from defaultProps is called. Playing around with the src I was able to pass them from SingleDatePicker to the controller and they were called as expected.

I noticed that DateRangePicker doesn't have this issue. These functions are listed as optional props in the docs so I'm not sure if this is intentional and just something that shouldn't be used with the SingleDatePicker or if it's a bug? 

Any help is appreciated, thanks again.

brosewarne:
I've been able to work around this for some use cases by using DayPickerSingleDateController, however it looks like onNextMonthClick doesn't let me stop the navigation.

It'd be great of there was a way to stop navigation when requesting a month that is outside the supplied range, but it's not a huge issue / blocker at the moment.

majapw:
Hiyo! I think this has a couple of duplicate issues: https://github.com/airbnb/react-dates/issues/628 and https://github.com/airbnb/react-dates/issues/400. :) 

I will close those in favor of this issue so we can have some discussion here. 

I think with the expectation of how `isOutsideRange` works (`(- ∞, X]` or `[Y,  ∞)` or both), it would be reasonable to have a `disableOutsideRangeNavigation` boolean prop that would allow for this behavior. 

FYI @kakadiadarpan @AntiFish03

PR #1475, opened 2018-11-26
Disable navigation outside of range

petetalksweb:
Addresses issue #647 

Adds property ```disableOutsideRangeNavigation``` (as suggested by @majapw in [this thread](https://github.com/airbnb/react-dates/issues/631)) to ```SingleDatePicker``` and ```DateRangePicker``` components, which when set to ```true```, removes navigation to months outside of the range specified by the ```isOutsideRange``` function.

### DIFF
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -511,7 +511,6 @@ class DateRangePicker extends React.PureComponent {
       openDirection,
       phrases,
       isOutsideRange,
-      disableOutsideRangeNavigation,
       minimumNights,
       withPortal,
       withFullScreenPortal,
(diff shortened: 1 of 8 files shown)


leBoer:
What if the isOutsideRange function resolves to something that is not a single block of dates? For instance every July could return true. Whilst the rest is false? Would this still work?

majapw:
I'm not sure if it makes sense to add this now that #1311 has been merged in and release in v18.3.0. I appreciate the contribution but the `minDate` and `maxDate` props should address this behavior, no, @ljharb?

majapw:
In that case, I'm gonna close the pull request! :) Thank you for the contribution @petetalksweb

PR #1482, opened 2018-12-06
Update the SDP and DRP to be compatible with react-with-direction

majapw:
This change prevents automatic flipping on styles that are manually handled via the isRTL prop.

to: @airbnb/ui-frameworks @ljharb @yzimet

PR #1492, opened 2018-12-12
[RTL] Fix the SDP and DRP noflip util function

yzimet:
### Summary
In #1482 we added compatibility with react-with-direction. In the process we
created a helper utility `noflip` to make it so that react-with-direction
wouldn't interfere with the preexisting `isRTL` flag. This PR fixes the noflip
behavior for Numbers, and adds unit tests.

to: @majapw @ljharb @airbnb/ui-frameworks

### Screenshot
example: DateRangePicker: with DirectionProvider

_before_
image

_after_
image

### DIFF
--- a/src/utils/noflip.js
+++ b/src/utils/noflip.js
@@ -4,7 +4,7 @@ const NOFLIP = '/* @noflip */';
 // flipped in RTL contexts. This should be used only in situations where the style must remain
 // unflipped regardless of direction context. See: https://github.com/kentcdodds/rtl-css-js#usage
 export default function noflip(value) {
-  if (typeof value === 'number') return `${value}px`;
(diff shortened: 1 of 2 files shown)


src/utils/noflip.js
```diff
@@ -4,7 +4,7 @@ const NOFLIP = '/* @noflip */';
 // flipped in RTL contexts. This should be used only in situations where the style must remain
 // unflipped regardless of direction context. See: https://github.com/kentcdodds/rtl-css-js#usage
 export default function noflip(value) {
-  if (typeof value === 'number') return `${value}px`;
+  if (typeof value === 'number') return `${value}px ${NOFLIP}`;
```
majapw: OOPS. 😱

PR #1499, opened 2019-01-03
Improve keyboard navigation for date range selection

monokrome:
These changes represent an effort to improve the navigation of the date range selection via keyboard. This is primarily modifying the position of the `DayPicker` so that it is after the currently focused input on the date ranges, and comes with a few small changes to keep reasonable usability for keyboard navigation inside of the `DayPicker`.


.gitignore
```diff
@@ -48,3 +48,4 @@ package-lock.json
 
 css/styles.css
 
+.cache
```
ljharb: what creates this dir? (also it'd be good to ensure all files have a trailing newline)

src/components/DateRangePicker.jsx
```diff
@@ -584,14 +587,15 @@ class DateRangePicker extends React.PureComponent {
...
-        {enableOutsideClick && (
-          <OutsideClickHandler onOutsideClick={this.onOutsideClick}>
-            {input}
-            {this.maybeRenderDayPickerWithPortal()}
-          </OutsideClickHandler>
-        )}
-        {!enableOutsideClick && input}
-        {!enableOutsideClick && this.maybeRenderDayPickerWithPortal()}
+        {
+          enableOutsideClick
```
ljharb: i think the "two statements" form is a bit more readable than the ternary form, but i don't feel strongly about it

src/components/DateRangePickerInput.jsx
```diff
@@ -153,6 +158,7 @@ function DateRangePickerInput({
   customArrowIcon,
   customCloseIcon,
   isFocused,
+  datePickerIsAfterEndInput,
```
ljharb: what's the propType and defaultProp for this?

src/components/SingleDatePicker.jsx
```diff
@@ -242,6 +250,13 @@ class SingleDatePicker extends React.PureComponent {
     });
   }
 
+  onFocusOut(e) {
+    if (!this.container.contains(e.relatedTarget)) {
```
ljharb: hmm, looks like we don't document that we rely on `.contains` on HTML elements, but we use this elsewhere, so it's not a new requirement.

src/components/SingleDatePicker.jsx
```diff
@@ -559,14 +576,15 @@ class SingleDatePicker extends React.PureComponent {
...
-          <OutsideClickHandler onOutsideClick={this.onOutsideClick}>
-            {input}
-            {this.maybeRenderDayPickerWithPortal()}
-          </OutsideClickHandler>
-        )}
-        {!enableOutsideClick && input}
-        {!enableOutsideClick && this.maybeRenderDayPickerWithPortal()}
+        {
+          enableOutsideClick
+            ? (
```
ljharb: ditto here re ternary

src/components/SingleDatePickerInput.jsx
```diff
@@ -19,6 +19,7 @@ import { ICON_BEFORE_POSITION, ICON_AFTER_POSITION, OPEN_DOWN } from '../constan
 const propTypes = forbidExtraProps({
   ...withStylesPropTypes,
   id: PropTypes.string.isRequired,
+  children: PropTypes.node,
```
ljharb: do we need to allow any kind of children here?

src/components/DateRangePickerInputController.jsx
```diff
@@ -28,6 +28,8 @@ import {
 } from '../constants';
 
 const propTypes = forbidExtraProps({
+  children: PropTypes.node,
```
ljharb: do we need to allow any kind of children here?

src/components/DateRangePickerInput.jsx
```diff
@@ -27,6 +27,9 @@ import {
 
 const propTypes = forbidExtraProps({
   ...withStylesPropTypes,
+
+  children: PropTypes.node,
```
ljharb: do we need to allow any kind of children here?

majapw: Yeah, let's restrict this to the DayPicker, DayPickerRangeController, and DayPickerSingleDateController. We can open it up if people ask.

src/components/DateRangePickerInput.jsx
majapw: Also can we not calculate this based on the isStartDateFocused/isEndDateFocused props? Do we need this new one?

src/components/DateRangePickerInput.jsx
```diff
@@ -233,11 +241,14 @@ function DateRangePickerInput({
           {...css(styles.DateRangePickerInput_arrow)}
           aria-hidden="true"
           role="presentation"
+          onKeyDownTab={onEndDateTab}
         >
           {arrowIcon}
         </div>
       }
 
+      {datePickerIsAfterEndInput || children}
```
majapw: I think this can be `{isStartDateFocused && children}`

src/components/DateRangePickerInput.jsx
```diff
@@ -252,14 +263,15 @@ function DateRangePickerInput({
...
         onFocus={onEndDateFocus}
-        onKeyDownTab={onEndDateTab}
         onKeyDownArrowDown={onKeyDownArrowDown}
         onKeyDownQuestionMark={onKeyDownQuestionMark}
         verticalSpacing={verticalSpacing}
         small={small}
         regular={regular}
       />
 
+      {datePickerIsAfterEndInput && children}
```
majapw: I think this can be `{isEndDateFocused && children}`

src/components/DateRangePickerInputController.jsx
majapw: Would be good to restrict

src/components/DateRangePickerInputController.jsx
```diff
@@ -232,6 +242,11 @@ export default class DateRangePickerInputController extends React.PureComponent
 
   onStartDateFocus() {
     const { disabled, onFocusChange } = this.props;
+
+    this.setState({
```
majapw: I'm not convinced we need this state value. we can use the `isStartDateFocused` and `isEndDateFocused` props instead

src/components/SingleDatePicker.jsx
```diff
@@ -161,6 +162,11 @@ class SingleDatePicker extends React.PureComponent {
     }
 
     this.isTouchDevice = isTouchDevice();
+
+    // We manually set event because React has not implemented onFocusIn/onFocusOut.
+    // Keep an eye on https://github.com/facebook/react/issues/6410 for updates
+    // We use "blur w/ useCapture param" vs "onfocusout" for FF browser support
+    this.container.addEventListener('blur', this.onFocusOut, true);
```
majapw: As mentioned offline, I think it would be preferable and simpler to not have this blur functionality but instead rely on the tab callbacks in the DayPicker

.gitignore
ljharb: this is a good change if it’s generated by a tool in the repo, but otherwise it should probably be in your global gitignore instead of in the repo

src/components/DateRangePickerInput.jsx
monokrome: @majapw Maybe. Does `isEndDateFocused` take into account whether the date input is focused, or just the day picker widget?

monokrome: @ljharb It's not added, but this isn't ready for review yet - either. I was just asked by @majapw to put the PR up anyway :)

.gitignore
monokrome: @TaeKimJR Do you know what generated this / why it was added? I'm guessing it was part of the repo since @majapw has seen it as well, but it'd be good to have it documented at least in a discussion somewhere. I can add a comment as well if we figure it out :)

src/components/DateRangePicker.jsx
monokrome: I agree with this. Are you cool w/ me changing it, @TaeKimJR?

src/components/DateRangePickerInput.jsx
monokrome: @majapw I'll try it! That's something I didn't notice until I added the other one, so makes sense as long as the input element is taken into account w/ it as well!

src/components/DateRangePickerInput.jsx
monokrome: 🎉

src/components/DateRangePickerInputController.jsx
monokrome: @TaeKimJR added this stuff, so pinging them to get thoughts on this approach and whether we should change it :)

src/components/DateRangePickerInputController.jsx
monokrome: Neither was I, I was going to investigate it a bit more after I figured out the tabbing issues (which I think I'll need to do some somewhat-hacky stuff to get working, unfortunately). Thanks :)

src/components/SingleDatePicker.jsx
monokrome: @majapw This is from @TaeKimJR's implementation before I took on this project, but the weird situation here is that the only context we have from the browser `onTab` is what element we were focused on *before* the `onTab`? `document.activeElement` seems to be reset to `body` during the event cycle w/ React. I think `e.relatedTarget` may be what I'm looking for, but still need to test that theory. The only solution I've found thus far is quite hacky.

src/components/SingleDatePicker.jsx
monokrome: Not sure that we need this, and I will investigate if there's another approach - but it seems that we should document that regardless if other things are using it. Is the documentation in the README, elsewhere, or both?

ljharb: It should go in the readme; it's not currently there.

src/components/DateRangePickerInputController.jsx
TaeKimJR: Restricting makes sense!

src/components/SingleDatePicker.jsx
```diff
@@ -562,11 +579,9 @@ class SingleDatePicker extends React.PureComponent {
         {enableOutsideClick && (
           <OutsideClickHandler onOutsideClick={this.onOutsideClick}>
             {input}
-            {this.maybeRenderDayPickerWithPortal()}
           </OutsideClickHandler>
         )}
-        {!enableOutsideClick && input}
-        {!enableOutsideClick && this.maybeRenderDayPickerWithPortal()}
+        { enableOutsideClick || input}
```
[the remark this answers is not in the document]
monokrome: @ljharb I'll commit this suggestion, but eslint seems to keep changing it back >.<

.gitignore
monokrome: I'm removing this. If it comes back, let's figure out why and add back w/ a comment! @TaeKimJR said it may have been storybook, but I can't reproduce it. :)

src/components/DateRangePickerInput.jsx
monokrome: I think that it needs to be in DatePickerController as well, or else we'll need to move around         the logic for `this.maybeRenderDayPickerWithPortal()`

src/components/SingleDatePickerInput.jsx
monokrome: I think so, because we parent the daypicker in the range selector (at least) so that it points the arrow up to the input itself

src/components/DateRangePickerInputController.jsx
monokrome: I did some restricting and it looks like we use this

ljharb: what things did it break?

it's mainly that, we can loosen easily later, but restricting later is a hassle, so it's always nicer to restrict as much as possible at the start

monokrome:
Oops! Need to write test! Will do so tonight :D

src/components/DateRangePicker.jsx
```diff
@@ -238,6 +241,11 @@ class DateRangePicker extends React.PureComponent {
     });
   }
 
+  onDayPickerFocusOut(e) {
+    if (this.dayPickerContainer.contains(e.relatedTarget)) return;
```
majapw: Just want to confirm that clicking on the month navigation, keyboard shortcuts, or any white space in the calendar does *not* close the picker, yeah?

src/components/DateRangePicker.jsx
```diff
@@ -247,13 +255,28 @@ class DateRangePicker extends React.PureComponent {
   }
 
   setDayPickerContainerRef(ref) {
+    if (this.dayPickerContainer) this.removeDayPickerEventListeners();
+    if (!ref) return;
     this.dayPickerContainer = ref;
+    this.addDayPickerEventListeners();
```
majapw: Is it expensive to be removing and adding this all the time? Can we add/remove it only if the ref has changed maybe?

src/components/DateRangePicker.jsx
```diff
@@ -247,13 +255,28 @@ class DateRangePicker extends React.PureComponent {
...
   }
 
   setContainerRef(ref) {
     this.container = ref;
   }
 
+  addDayPickerEventListeners() {
+    this.removeDayPickerFocusOut = addEventListener(
```
majapw: It would be great if we could add a comment here as to why we're using this approach rather than a React event listener!

src/components/DateRangePickerInput.jsx
majapw: Sorry @monokrome, can you clarify what you mean here?

src/components/DateRangePickerInputController.jsx
majapw: Can you clarify what it broke? It's a new prop so I feel like we should be able to control the type of what's passed in

src/components/SingleDatePicker.jsx
```diff
@@ -128,6 +128,7 @@ class SingleDatePicker extends React.PureComponent {
       showKeyboardShortcuts: false,
```
majapw: @monokrome I think the SingleDatePicker needs to be updated to match the DRP still. It's still using the `blur` event and it'd be good for the implementation to look as identical as possible with the class methods and so forth.

majapw:
Hey @monokrome, would be good to get a status update on this tomorrow! I'd like to do a release for @nkinser's change and I'd love to get this in too if it's close.

monokrome:
Discussed a bit with @majapw, but I want to document why we are using `focusout`instead of tab for anyone else following.

I spent a bit of time trying to find a way to do this with onblur/onfocus/onkeydown, but the event doesn't give us the information that we need. Tabbing out sets `document.activeElement` to the *body* element when the next item isn't a form element (possibly in that case as well).

In order to ensure that we are staying safe, I have implemented this using the `focusout` event on an element *ref* until https://github.com/facebook/react/issues/6410 is released, at which time we can consider using FocusOut natively. 😺

src/components/DateRangePicker.jsx
monokrome: @majapw Hmm... Apparently it does? Not sure why, I'll look into it. I'd expect this contains check to work but I may need to get a ref to a parent container?

monokrome: @majapw I've pushed a fix - good find! I also fixed another instance of the same issue.

src/components/DateRangePicker.jsx
monokrome: Sure!

src/components/DateRangePicker.jsx
monokrome: @majapw I didn't realize that the ref would be `set` unless it changed. Is React not that efficient at this? If it's not then we should definitely be more careful 😺

monokrome: @majapw Is React not efficient about only calling this when the `ref` changes? If not, then we will need to be a bit more careful for sure.

src/components/DateRangePickerInput.jsx
monokrome: @majapw Sure! I think that ` this.maybeRenderDayPickerWithPortal() ` is rendered as a child of the input component right now. We can refactor to change this, but it would require additional effort:

https://github.com/airbnb/react-dates/blob/21e4a37a0ff02167d882a71137a8e8bfd1daa22c/src/components/SingleDatePicker.jsx#L568

src/components/DateRangePickerInputController.jsx
monokrome: Seems we're having the same conversation in two places 😺 See here: https://github.com/airbnb/react-dates/pull/1499/files#r245158005

src/components/SingleDatePicker.jsx
monokrome: @majapw Sure! I didn't do that part, so was getting everything right before the cleanup part of the code I wasn't working on directly. Working on it now 😺

src/components/DateRangePicker.jsx
majapw: I guess I don't actually know. 🤔I know it becomes null sometimes (presumably during rerenders and I guess it would be a new object afterwards), but maybe that would be more of a symptom of unnecessary rerenders. This is probably fine. :)

ljharb: That happens when the ref callback is `!==` across renders, for one.

src/components/SingleDatePicker.jsx
```diff
@@ -175,7 +177,8 @@ class SingleDatePicker extends React.PureComponent {
 
   /* istanbul ignore next */
   componentWillUnmount() {
-    if (this.removeEventListener) this.removeEventListener();
+    if (this.removeResizeEventListener) this.removeResizeEventListener();
```
monokrome: Renamed this to `removeResizeEventListener` because we've now added another event listener and it could lead to confusion otherwise 😺

src/components/DateRangePicker.jsx
```diff
@@ -238,6 +241,11 @@ class DateRangePicker extends React.PureComponent {
     });
   }
 
+  onDayPickerFocusOut(event) {
+    if (this.dayPickerContainer.contains(event.relatedTarget || event.target)) return;
```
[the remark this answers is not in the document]
monokrome: @majapw I'll add a comment to explain this one better, but it fixes the issue where clicking on one of the days caused the `DayPicker` to close. In this case, it will set **relatedTarget** to `null`, while in the other cases it gives us the correct **relatedTarget** but an incorrect **target**.

src/components/SingleDatePickerInputController.jsx
```diff
@@ -251,7 +255,9 @@ export default class SingleDatePickerInputController extends React.PureComponent
         small={small}
         regular={regular}
         verticalSpacing={verticalSpacing}
-      />
+      >
+        {children && children}
```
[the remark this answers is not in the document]
monokrome: @majapw I don't think so, because there wasn't a conditioner earlier. The case we're checking for here is whether **children** is `null`. In the case that **focused** is `true`, I still think that it's still possible to have **children** set to `null`

src/components/DayPickerRangeController.jsx
```diff
@@ -564,8 +570,8 @@ export default class DayPickerRangeController extends React.PureComponent {
         };
 
         // eslint-disable-next-line react/destructuring-assignment
-        if (this.state.dateOffset && this.state.dateOffset.start && this.state.dateOffset.end) {
-          modifiers = this.deleteModifierFromRange(modifiers, this.state.dateOffset.start, this.state.dateOffset.end, 'hovered-offset');
+        if (this.state.dateOffset && currentDateOffset.start && currentDateOffset.end) {
```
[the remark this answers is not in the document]
monokrome: Seems like a good idea! 😺

majapw:
Any update on this?

monokrome:
@majapw I think this is ready for review now. All changes mentioned have been made, with the caveat that I switched the naming for the `currentDateOffset` to renaming the local variable to `nextDateOffset` to retain `dateOffset` in state. Happy to change that, but this feels like a smaller change and the name **current** in state felt a bit redundant 😺

src/components/DayPickerRangeController.jsx
```diff
@@ -558,14 +564,14 @@ export default class DayPickerRangeController extends React.PureComponent {
...
-        dateOffset = {
+        nextDateOffset = {
           start,
           end,
         };
 
         // eslint-disable-next-line react/destructuring-assignment
-        if (this.state.dateOffset && this.state.dateOffset.start && this.state.dateOffset.end) {
-          modifiers = this.deleteModifierFromRange(modifiers, this.state.dateOffset.start, this.state.dateOffset.end, 'hovered-offset');
+        if (this.state.nextDateOffset && dateOffset.start && dateOffset.end) {
```
majapw: @monokrome I don't think `this.state.nextDateOffset` exists. Did you mean just `nextDateOffset`? Also, I think that it is always truthy so this check doesn't really make sense. 

I think that this is checking for the existing state and so should just use the deconstructed `dateOffset` and this line should be:
```
if (dateOffset && dateOffset.start && dateOffset.end) {
```

monokrome: @majapw Yep! Seems right. This was using `this.state`directly instead of the desctructured value, which seems like this is only for an edge-case that wasn't represented by tests. 🙀 

Problem solved now! Thanks for finding it :)

PR #1513, opened 2019-01-21
Remove `darken` function from css

LukeAskew:
Currently, this breaks the processing chain when importing of css when importing via ` @import 'react-dates/lib/css/_datepicker';`. For example, I have a component library that imports styles via the method above. When the compiled css (via postcss) is imported as a dependency, it throws a syntax error.

Removing the lines, without replacing, because the styles don't appear to do anything since the icon is an SVG and not text.

### DIFF
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -658,12 +658,10 @@ export default withStyles(({ reactDates: { color, zIndex } }) => ({
     zIndex: zIndex + 2,
 
     ':hover': {
-      color: `darken(${color.core.grayLighter}, 10%)`,
       textDecoration: 'none',
     },
 
     ':focus': {
-      color: `darken(${color.core.grayLighter}, 10%)`,
       textDecoration: 'none',
     },
   },
--- a/src/components/SingleDatePicker.jsx
+++ b/src/components/SingleDatePicker.jsx
@@ -633,12 +633,10 @@ export default withStyles(({ reactDates: { color, zIndex } }) => ({
     zIndex: zIndex + 2,
 
     ':hover': {
-      color: `darken(${color.core.grayLighter}, 10%)`,
       textDecoration: 'none',
     },
 
     ':focus': {
-      color: `darken(${color.core.grayLighter}, 10%)`,
       textDecoration: 'none',
     },
   },


11bit:
Hi @LukeAskew. This is pretty old PR.. but maybe you remember why it is closed and not merged?  I have the similar issue with build. Maybe there is a better solution for this?

LukeAskew:
@11bit Not 100% sure but I'm guessing just due to inactivity.

ljharb:
Just removing `darken` here will "break" the styling of the library; what makes this a syntax error?

LukeAskew:
From what I can remember, the imported styles were passing through the sass function, so it would throw a syntax error if you weren't using Sass since `darken` is Sass-specific.

ljharb:
hmm, indeed i see `darken()` calls in https://unpkg.com/browse/react-dates@21.8.0/lib/css/_datepicker.css, so that suggests that the proper fix isn't to remove `darken()`, it's to fix our CSS build process to it processes them out.

11bit:
Thanks for the detailed answer!

Yes, `darken` is sass specific. And it doesn't work very well with modern dart-sass - it is not possible to `@import "~react-dates/lib/css/_datepicker.css";` because dart-sass now expects only CSS syntax in .css files. 
So probably the best solution is to tune CSS build to process them out.

ljharb:
A PR to do that would be most appreciated.

11bit:
@ljharb sure, I can look into it. 

What approach do you think is better:
1. Process out `darken` on a build step
2. Transform colors in runtime with something like https://github.com/ricokahler/color2k. It will add a bit to the bundle but it is more aligned with css-in-js way

ljharb:
`darken` shouldn't just be removed - I think transforming the colors into static css is the only way forward.

11bit:
BTW how does it work when react-dates is used in css-in-js mode? I tried to check examples in storybook and `darken` seems to be not working there because it is not transpiled to normal css. But probably I miss something.

If `darken` is not working in css-in-js mode then the second option with processing colors in runtime will better because: 
1. It will fix css-in-js mode 
2. It will have the same static css output as if we process darken out when building css.

Just to clarify, the second option I am talking about is something like this:

```diff
+ import { darken } from 'color2k';

':hover': {
-  color: `darken(${color.core.grayLighter}, 10%)`,
+  color: darken(color.core.grayLighter, 0.1),
  textDecoration: 'none',
},
```

Anyway, I am willing to help with PR with either solution. Please let me know what you think.

ljharb:
I think I prefer the runtime option for "in JS" mode, and computing the darkened color for the `.css` output.

11bit:
@ljharb I've made a [PR](https://github.com/airbnb/react-dates/pull/2058). It works both for "in JS" mode by calculating darken color in run time and for static css. In static css there will be actual darkened color.

PR #1521, opened 2019-01-30
Add aria-disabled prop to CalendarDay and CustomizableCalendarDay

nkinser:
This PR adds the prop `aria-disabled` to `CalendarDay` and `CustomizableCalendarDay` so that days that are unavailable have the "dimmed" label. The unavailable dates are still perceivable by VO. 

#### Reviewers
@majapw @ljharb @backwardok


nkinser:
@majapw can I please get a stamp? Thanks!

PR #1524, opened 2019-02-01
Fixes the focus out event in IE11

marcos0x:


src/components/DateRangePicker.jsx
```diff
@@ -248,7 +248,7 @@ class DateRangePicker extends React.PureComponent {
     //
     // We handle both situations here by using the ` || ` operator to fallback
     // to *event.target** when **relatedTarget** is not provided.
-    if (this.dayPickerContainer.contains(event.relatedTarget || event.target)) return;
+    if (this.dayPickerContainer.contains(event.relatedTarget || event.target) || this.dayPickerContainer.contains(event.target)) return;
```
ljharb: If possible, can you add a test here to cover this?

In IE 11, what element does it provide as the `relatedTarget` when you actually want the `target`?

marcos0x: In IE11 relatedTarget is <body>, so the condition is false and all clicks are closing the datepicker

ljharb: Would it be more targeted to make the current `||` instead be something like `event.relatedTarget === document.body ? event.target : event.relatedTarget || target`?

majapw:
Can we add context to the comment about this check?

ljharb:
also surfacing https://github.com/airbnb/react-dates/pull/1524#discussion_r253175998

src/components/DateRangePicker.jsx
```diff
@@ -250,7 +250,8 @@ class DateRangePicker extends React.PureComponent {
     //
     // We handle both situations here by using the ` || ` operator to fallback
     // to *event.target** when **relatedTarget** is not provided.
-    if (this.dayPickerContainer.contains(event.relatedTarget || event.target)) return;
+    if (this.dayPickerContainer.contains(event.relatedTarget || event.target)
+      || this.dayPickerContainer.contains(event.target)) return;
```
monokrome: @marcos0x This is potentially doing the same operation twice, I think. I feel like we should follow @ljharb's advice here and change this to `this.dayPickerContainer.contains(event.relatedTarget) || this.dayPickerContainer.contains(event.target))` in order to prevent us from duplicating logic unnecessarily.

Although the current implementation works, it means that the `|| event.target` has become redundant. If that call fails, it'll revert to checking `event.target` anyway.

Is this something that you would be open to changing?

monokrome:
@marcos0x Would you mind adding tests for this usecase?

src/components/DateRangePicker.jsx
ljharb: (note that we'll probably need something more complex than an inline `||` here, since we won't want to call `contains` with a falsy element)

marcos0x:
@monokrome It is tested in IE11 and works well

src/components/DateRangePicker.jsx
```diff
@@ -250,7 +250,9 @@ class DateRangePicker extends React.PureComponent {
     //
     // We handle both situations here by using the ` || ` operator to fallback
     // to *event.target** when **relatedTarget** is not provided.
-    if (this.dayPickerContainer.contains(event.relatedTarget || event.target)) return;
+    const { contains } = this.dayPickerContainer;
+    const { relatedTarget, target } = event;
+    if (contains(relatedTarget) || contains(target)) return;
```
ljharb: if relatedTarget doesn’t exist, we shouldn’t invoke contains on it.

also, i suspect `contains` requires a proper `this` value, which means it can’t be destructured off of the ref.

marcos0x: @ljharb if relatedTarget or target doesn’t exist, contains return false

ljharb: I understand, but that doesn't mean we should bother to call the function when we can know it's going to return false.

(also if `contains()` by itself doesn't throw, it probably is doing `document.contains` implicitly, so it's always returning true for every element in the DOM)

marcos0x: @ljharb Why you add more checks? Is unnecessary for the functionality

ljharb: Calling into DOM functions is slow; any chance we can avoid doing that is a performance win.

marcos0x: @ljharb
const { contains } = this.dayPickerContainer;
It is not a DOM function. Do you know how destructuring works?

ljharb: yes it is, since `this.dayPickerContainer` is an HTML element. You can test it yourself and verify that `contains === HTMLElement.prototype.contains`.

There's no need to be hostile, I can assure you that I know very well how destructuring works.

marcos0x: @ljharb I'm sorry, I will fix

src/components/DateRangePicker.jsx
```diff
@@ -250,7 +250,8 @@ class DateRangePicker extends React.PureComponent {
     //
     // We handle both situations here by using the ` || ` operator to fallback
     // to *event.target** when **relatedTarget** is not provided.
-    if (this.dayPickerContainer.contains(event.relatedTarget || event.target)) return;
+    const relatedTarget = event.relatedTarget === document.body ? event.target : event.relatedTarget || event.target;
```
ljharb:
```suggestion
    const relatedTarget = event.relatedTarget === document.body ? event.target : (event.relatedTarget || event.target);
```

src/components/DateRangePicker.jsx
```diff
@@ -250,7 +250,8 @@ class DateRangePicker extends React.PureComponent {
     //
     // We handle both situations here by using the ` || ` operator to fallback
     // to *event.target** when **relatedTarget** is not provided.
```
ljharb: can we add a comment here explaining the document.body stuff for IE 11?

marcos0x: @ljharb done

marcos0x:
@monokrome Can you approve this?

devans-k4:
Waiting on this as well. Thanks btw @marcos0x

PR #1525, opened 2019-02-05
Move the call to onDatesChange before the call to onFocusChange

nkinser:
This PR moves the call to `onDatesChange` before `onFocusChange`. The purpose of this change is to allow the logic of `onDatesChange` to potentially affect the logic of `onFocusChange`. 

For example, I would like to allow users to click on certain unavailable dates and then show them a corresponding error state. When such a date is clicked, I do not want the focus to change because the date they clicked would not actually be selected. The logic of whether or not to change focus needs to occur in `onDatesChange`, so this function needs to be called before `onFocusChange`.

### DIFF
--- a/src/components/DayPickerRangeController.jsx
+++ b/src/components/DayPickerRangeController.jsx
@@ -490,6 +490,7 @@ export default class DayPickerRangeController extends React.PureComponent {
     if (startDateOffset || endDateOffset) {
       startDate = getSelectedDateOffset(startDateOffset, day);
       endDate = getSelectedDateOffset(endDateOffset, day);
+      onDatesChange({ startDate, endDate });
 
       if (this.isBlocked(startDate) || this.isBlocked(endDate)) {
         return;
@@ -512,6 +513,8 @@ export default class DayPickerRangeController extends React.PureComponent {
         }
       }
 
+      onDatesChange({ startDate, endDate });
+
       if (isEndDateDisabled && !isStartDateAfterEndDate) {
         onFocusChange(null);
         onClose({ startDate, endDate });
@@ -523,9 +526,11 @@ export default class DayPickerRangeController extends React.PureComponent {
 
       if (!startDate) {
         endDate = day;
+        onDatesChange({ startDate, endDate });
         onFocusChange(START_DATE);
       } else if (isInclusivelyAfterDay(day, firstAllowedEndDate)) {
         endDate = day;
+        onDatesChange({ startDate, endDate });
         if (!keepOpenOnDateSelect) {
           onFocusChange(null);
           onClose({ startDate, endDate });
@@ -533,10 +538,14 @@ export default class DayPickerRangeController extends React.PureComponent {
       } else if (disabled !== START_DATE) {
         startDate = day;
         endDate = null;
+        onDatesChange({ startDate, endDate });
+      } else {
+        onDatesChange({ startDate, endDate });
       }
+    } else {
+      onDatesChange({ startDate, endDate });
     }
 
-    onDatesChange({ startDate, endDate });
     onBlur();
   }
 
--- a/src/components/DayPickerRangeController.jsx
+++ b/src/components/DayPickerRangeController.jsx
@@ -490,11 +490,12 @@ export default class DayPickerRangeController extends React.PureComponent {
     if (startDateOffset || endDateOffset) {
       startDate = getSelectedDateOffset(startDateOffset, day);
       endDate = getSelectedDateOffset(endDateOffset, day);
-      onDatesChange({ startDate, endDate });
 
       if (this.isBlocked(startDate) || this.isBlocked(endDate)) {
         return;
       }
+
+      onDatesChange({ startDate, endDate });
 
       if (!keepOpenOnDateSelect) {
         onFocusChange(null);
--- a/test/components/DayPickerRangeController_spec.jsx
+++ b/test/components/DayPickerRangeController_spec.jsx
@@ -1635,7 +1635,7 @@ describe('DayPickerRangeController', () => {
           />
         ));
         wrapper.instance().onDayClick(clickDate);
-        expect(onDatesChangeStub.callCount).to.equal(1);
+        expect(onDatesChangeStub).to.have.property('callCount', 1);
         const args = onDatesChangeStub.getCall(0).args[0];
         expect(args.startDate.format()).to.equal(clickDate.clone().format());
         expect(args.endDate).to.equal(null);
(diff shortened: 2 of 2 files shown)


nkinser:
@majapw PTAL! 
This passes all the tests and I've also checked in storybook to make sure everything is acting as expected. Is there anything I should do to test this more extensively?

ljharb:
If the ordering is important, then I think we should add a test that asserts it (also, you'll need to rebase)

src/components/DayPickerRangeController.jsx
```diff
@@ -512,6 +514,8 @@ export default class DayPickerRangeController extends React.PureComponent {
         }
       }
 
+      onDatesChange({ startDate, endDate });
```
ljharb: i'm a little worried about this line being the first of twice - can we add a test with spies, for each of these branches, to ensure it's called both once, and in the proper order?

nkinser: @ljharb Can you explain what you mean by "this line being the first of twice"? 
`onDatesChange` will only be called once in this function, and I can add a test to ensure this

ljharb: That's my exact concern; a test is great :-)

test/components/DayPickerRangeController_spec.jsx
```diff
@@ -1622,6 +1622,177 @@ describe('DayPickerRangeController', () => {
...
+        const clickDate = moment(today);
+        const onDatesChangeStub = sinon.stub();
+        const wrapper = shallow((
+          <DayPickerRangeController
+            onDatesChange={onDatesChangeStub}
+            endDateOffset={day => day.add(4, 'days')}
+          />
+        ));
+        wrapper.instance().onDayClick(clickDate);
+        expect(onDatesChangeStub.callCount).to.equal(1);
```
ljharb:
```suggestion
        expect(onDatesChangeStub).to.have.property('callCount', 1);
```

test/components/DayPickerRangeController_spec.jsx
```diff
@@ -1622,6 +1622,177 @@ describe('DayPickerRangeController', () => {
...
+        const clickDate = moment(today);
+        const onDatesChangeStub = sinon.stub();
+        const wrapper = shallow((
+          <DayPickerRangeController
+            onDatesChange={onDatesChangeStub}
+            startDateOffset={day => day.subtract(2, 'days')}
+          />
+        ));
+        wrapper.instance().onDayClick(clickDate);
+        expect(onDatesChangeStub.callCount).to.equal(1);
```
ljharb:
```suggestion
        expect(onDatesChangeStub).to.have.property('callCount', 1);
```

test/components/DayPickerRangeController_spec.jsx
```diff
@@ -1622,6 +1622,177 @@ describe('DayPickerRangeController', () => {
...
+            onDatesChange={onDatesChangeStub}
+            focusedInput={END_DATE}
+            disabled={START_DATE}
+            minimumNights={2}
+            startDate={startDate}
+            endDate={null}
+          />
+        ));
+        wrapper.instance().onDayClick(clickDate);
+        expect(onDatesChangeStub.callCount).to.equal(1);
```
ljharb:
```suggestion
        expect(onDatesChangeStub).to.have.property('callCount', 1);
```

test/components/DayPickerRangeController_spec.jsx
```diff
@@ -1622,6 +1622,177 @@ describe('DayPickerRangeController', () => {
...
+          <DayPickerRangeController
+            onDatesChange={onDatesChangeStub}
+            focusedInput={END_DATE}
+            minimumNights={2}
+            startDate={clickDate.clone().add(1, 'days')}
+            endDate={null}
+          />
+        ));
+        wrapper.instance().onDayClick(clickDate);
+        expect(onDatesChangeStub.callCount).to.equal(1);
```
ljharb:
```suggestion
        expect(onDatesChangeStub).to.have.property('callCount', 1);
```

test/components/DayPickerRangeController_spec.jsx
```diff
@@ -1622,6 +1622,177 @@ describe('DayPickerRangeController', () => {
...
+        const wrapper = shallow((
+          <DayPickerRangeController
+            onDatesChange={onDatesChangeStub}
+            focusedInput={END_DATE}
+            minimumNights={2}
+            startDate={startDate}
+          />
+        ));
+        wrapper.instance().onDayClick(clickDate);
+        expect(onDatesChangeStub.callCount).to.equal(1);
```
ljharb:
```suggestion
        expect(onDatesChangeStub).to.have.property('callCount', 1);
```

test/components/DayPickerRangeController_spec.jsx
```diff
@@ -1622,6 +1622,177 @@ describe('DayPickerRangeController', () => {
...
+        const onDatesChangeStub = sinon.stub();
+        const wrapper = shallow((
+          <DayPickerRangeController
+            onDatesChange={onDatesChangeStub}
+            focusedInput={END_DATE}
+            startDate={null}
+          />
+        ));
+        wrapper.instance().onDayClick(clickDate);
+        expect(onDatesChangeStub.callCount).to.equal(1);
```
ljharb:
```suggestion
        expect(onDatesChangeStub).to.have.property('callCount', 1);
```

test/components/DayPickerRangeController_spec.jsx
```diff
@@ -1622,6 +1622,177 @@ describe('DayPickerRangeController', () => {
...
+        const onDatesChangeStub = sinon.stub();
+        const wrapper = shallow((
+          <DayPickerRangeController
+            onDatesChange={onDatesChangeStub}
+            focusedInput={START_DATE}
+            endDate={null}
+          />
+        ));
+        wrapper.instance().onDayClick(clickDate);
+        expect(onDatesChangeStub.callCount).to.equal(1);
```
ljharb:
```suggestion
        expect(onDatesChangeStub).to.have.property('callCount', 1);
```

nkinser:
@majapw @ljharb Do you know when we will be able to release this?

test/components/DayPickerRangeController_spec.jsx
```diff
@@ -1652,10 +1779,8 @@ describe('DayPickerRangeController', () => {
             focusedInput={START_DATE}
           />
         ));
-        // The first day click sets preventFocusChange to true, but it doesn't take effect until the
-        // second day click because onFocusChange is called before onDatesChange
         wrapper.instance().onDayClick(clickDate);
-        expect(focusedInput).to.equal(END_DATE);
```
ljharb: this does seem like a possible breaking change to me. @majapw?

majapw: Yep! I'm gonna release it as v19

PR #1528, opened 2019-02-05
Fix issues with Portal not including container elements.

### LINKED ISSUE #1522: SingleDatePicker auto closes on clicking anywhere when appendToBody={true}
**react-dates version**
react-dates@18.4.1 (18.4.0 is fine)

**Describe the bug**
When `SingleDatePicker` has `appendToBody={true}`, clicking anywhere (including a date) closes the calendar (and no date is selected).

If you click any of the day names ("Mo", "Tu", etc.) the calendar works as expected from then on.

```jsx
const Datepicker = props => {
  const {
    date,
    id,
    placeholder,
    isDisabled,
    isFocused,
    onDateChange,
    onFocusChange,
  } = props
  return (
    <SingleDatePicker
      date={date}
      onDateChange={onDateChange}
      focused={isFocused}
      onFocusChange={onFocusChange}
      id={id}
      placeholder={placeholder}
      disabled={isDisabled}
      numberOfMonths={1}
      verticalSpacing={11}
      hideKeyboardShortcutsPanel
      customInputIcon={<Icon name="calendar" />}
      navPrev={<Icon name="arrow-left" />}
      navNext={<Icon name="arrow-right" />}
      inputIconPosition="after"
      isOutsideRange={() => false}
      appendToBody={true}
    />
  )
}
```

Parent component handlers:

```jsx
<Datepicker
  date={values.date}  // set from `setFieldValue`
  onFocusChange={({ focused }) => {
    if (!focused) {
      setFieldTouched('date', true, true)
    }

    setDatePickerFocused(focused)
  }}
  onDateChange={date => {
    setFieldValue('date', date, true)
  }}
  id="event_date"
  isFocused={datePickerFocused} // set from `setDatePickerFocused`
  isDisabled={false}
/>
```

**Desktop:**
 - OS: Mac OS Mojave Version 10.14 (18A391)
 - Browser: Chrome Version 71.0.3578.98 (Official Build) (64-bit)

But happens anywhere I tried.

Yes

https://airbnb.io/react-dates/?selectedKind=SDP%20-%20Calendar%20Props&selectedStory=appended%20to%20body&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

**Additional context**

Happens since v18.4.1.

monokrome:
After the 18.4.1 patch, we an issue has occurred where `focusout` always causes the day picker to close when a click occurs anywhere on the page - even within the day picker itself. This is because the `<Portal>` is being placed around the day picker instead of it's containers. This PR fixes that issue.

Thanks to @alvaro-cuesta for discovering this one and letting us know; fixes #1522

### DIFF
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -563,7 +563,6 @@ class DateRangePicker extends React.PureComponent {
       isOutsideRange,
       minimumNights,
       withPortal,
-      appendToBody,
       withFullScreenPortal,
       displayFormat,
       reopenPickerOnClearDates,
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -631,7 +631,7 @@ class DateRangePicker extends React.PureComponent {
       </DateRangePickerInputController>
     );
 
-    return this.maybeWithPortal(
+    return this.maybeWithPortal((
       <div
         ref={this.setContainerRef}
         {...css(
@@ -645,8 +645,8 @@ class DateRangePicker extends React.PureComponent {
           </OutsideClickHandler>
         )}
         {enableOutsideClick || input}
-      </div>,
-    );
+      </div>
+    ));
   }
 }
 
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -363,22 +363,29 @@ class DateRangePicker extends React.PureComponent {
     });
   }
 
-  maybeRenderDayPickerWithPortal() {
-    const { withPortal, withFullScreenPortal, appendToBody } = this.props;
-
+  maybeRenderDayPicker() {
     if (!this.isOpened()) {
       return null;
     }
 
+    return this.renderDayPicker();
+  }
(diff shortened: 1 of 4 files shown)


src/components/DateRangePicker.jsx
```diff
@@ -637,7 +645,7 @@ class DateRangePicker extends React.PureComponent {
           </OutsideClickHandler>
         )}
         {enableOutsideClick || input}
-      </div>
+      </div>,
```
ljharb: an alternative to the trailing comma is `((` and `))`

monokrome: @ljharb I think this just came w/ prettier? :D

majapw:
@monokrome can you add some information about how the DOM structure is actually changed by this? You also mentioned there was a regression with the positioning of the calendar with this change; if that's still true, can you provide a screenshot of that behavior?

monokrome:
Apparently github thinks this was closed, so opening another.

ljharb:
Replaced by #1533

Issue #1536, opened 2019-02-08
Last in range day has incorrect border


krissalvador27:
**react-dates version**
react-dates@18.4.0

**Describe the bug**
Last in selected date range has wrong border top. In the screenshot below day 21 has a gray border (looks like it's the default border). Would expect it to be green like the days in the range before it.

**Source code (including props configuration)**
```
<DateRangePicker
  startDate={this.state.startDate}
  startDateId="your_unique_start_date_id"
  endDate={this.state.endDate}
  endDateId="your_unique_end_date_id"
  onDatesChange={({ startDate, endDate }) => this.setState({ startDate, endDate })}
  focusedInput={this.state.focusedInput}
  onFocusChange={focusedInput => this.setState({ focusedInput })}
/>
```

**Screenshots/Gifs**
<img width="596" alt="screen shot 2019-02-07 at 10 57 14 pm" src="https://user-images.githubusercontent.com/5139481/52458164-bd3d2580-2b2b-11e9-804f-2d2bad56155e.png">

**Desktop (please complete the following information):**
 - OS: Mac OS 10.14
 - Browser: Chrome 
 - Version 71.0.3578.98

**Is the issue reproducible in Storybook?**
http://airbnb.io/react-dates/?selectedKind=DateRangePicker%20%28DRP%29&selectedStory=default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

**Additional context**
Inspecting the styles it looks like this rule 
```
.CalendarDay__last_in_range,
.CalendarDay__last_in_range:hover {
  border-style: solid;
}
```
conflicts with the default `border-style: double`.

krissalvador27:
@ljharb would love to fix and test it, but I don't think the CSS file is exposed in the repo (only in the published version on npm). Any suggestions?

ljharb:
The css file is built from the in-JS styles. You should be able to make a fix there.

krissalvador27:
fixed in `v20.0.0`

PR #1538, opened 2019-02-10
Remove last-in-range style modifiers to fix border issue

krissalvador27:
Fixes issue https://github.com/airbnb/react-dates/issues/1536

## Before
* Border top of last day in selected range is not selected color

screen shot 2019-02-08 at 3 37 04 pm

* Border of the hovered day overlaps border of the last day in selected range 

screen shot 2019-02-09 at 4 41 49 pm

## After
* Border top of last day in selected range is expected selected color
screen shot 2019-02-08 at 3 36 44 pm

* Border of the hovered day no longer overlaps border of the last day in selected range 

screen shot 2019-02-09 at 4 41 35 pm

* Not sure if it's an issue (I don't think it is) but the border of the last in range day overlaps the border of the last day, making it appear 1 pixel smaller than the first day. It was previously fixed with PR https://github.com/airbnb/react-dates/pull/1328 but I think that surfaced the issue this PR fixes now.

### DIFF
--- a/src/components/CustomizableCalendarDay.jsx
+++ b/src/components/CustomizableCalendarDay.jsx
@@ -63,6 +63,7 @@ const propTypes = forbidExtraProps({
   blockedOutOfRangeStyles: DayStyleShape,
   hoveredSpanStyles: DayStyleShape,
   selectedSpanStyles: DayStyleShape,
+  lastInRangeStyles: DayStyleShape,
   selectedStyles: DayStyleShape,
   selectedStartStyles: DayStyleShape,
   selectedEndStyles: DayStyleShape,
@@ -159,6 +160,8 @@ export const selectedSpanStyles = {
   },
 };
 
+export const lastInRangeStyles = {};
+
 export const selectedStyles = {
   background: color.selected.backgroundColor,
   border: `1px double ${color.selected.borderColor}`,
@@ -194,6 +197,7 @@ const defaultProps = {
   blockedOutOfRangeStyles,
   hoveredSpanStyles,
   selectedSpanStyles,
+  lastInRangeStyles,
   selectedStyles,
   selectedStartStyles: {},
   selectedEndStyles: {},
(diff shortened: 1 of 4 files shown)


test/components/DayPickerRangeController_spec.jsx
```diff
@@ -2698,18 +2698,6 @@ describe('DayPickerRangeController', () => {
       expect(modifiers.has('selected-span')).to.equal(true);
     });
 
-    it('contains `last-in-range` if this.isLastInRange returns true', () => {
```
ljharb: The removal of this test indicates it’s a breaking change; can we avoid that?

krissalvador27: I can revert the code deleted on [`this.modifier`](https://github.com/airbnb/react-dates/pull/1538/files/34ded9e0d6685c34b324ba7b51d562816100c3ee#diff-a5cfcc83315d3bea09a9b06437d7ee80L201) but keep the deleted [last in range check and styles add](https://github.com/airbnb/react-dates/pull/1538/files/34ded9e0d6685c34b324ba7b51d562816100c3ee#diff-e5503a8207c75f758aa5c4cbe229a678L330) that caused the issue

krissalvador27:
@ljharb let me know if there's anything else you'd like me to do :)

src/components/CustomizableCalendarDay.jsx
```diff
@@ -160,14 +159,6 @@ export const selectedSpanStyles = {
   },
 };
 
-export const lastInRangeStyles = {
```
ljharb: to be really careful about breaking changes, we might want to keep this as `export const lastInRangeStyles = {}`

nkinser: I think making `lastInRangeStyles = {}` would fix the bug without needing to completely remove the modifier's styles. This would make it so that there is no built-in styling for `lastInRange`, thus fixing the bug, but you could still supply your own styles if you wanted to.

krissalvador27:
@nkinser @ljharb Great catch! Removing lastInRangeStyles would have been a breaking change for sure. Made it an `{}` and rebased commits on master.

krissalvador27:
@majapw anything else needed to get this merged in?

PR #1540, opened 2019-02-10
Fix date selection issue with withPortal or appendToBody. Fixes #1535.

### LINKED ISSUE #1535: appendToBody property breaks date selection in SingleDatePicker
**react-dates version**

- react-dates@19.0.0
- react-dates@18.5.x
- react-dates@18.4.1

**Describe the bug**

When `appendToBody` is set in a `SingleDatePicker`, the picker displays correctly on click but `onDateChange` callback is never called and the picker gets closed immediately. The date doesn't update.

This actually looks like a regression from #1522 . It appears that it was introduced in 18.4.1 (18.4.0 and previous versions work as expected).

```jsx
class App extends Component {
  constructor(props) {
    super(props);
    this.state = {
      focused: props.autoFocus,
      date: props.initialDate,
    };

    this.onDateChange = this.onDateChange.bind(this);
  }

  onDateChange(date) {
    this.setState({ date });
  }

  onFocusChange({ focused }) {
    this.setState({ focused });
  }

  render() {
   const { focused, date } = this.state; 
    return (
        <SingleDatePicker
          id="mydatepicker"
          date={date}
          focused={focused}
          onDateChange={this.onDateChange}
          appendToBody
        />
    );
  }
}
```

**Screenshots/Gifs**

None.

 - OS: Archlinux
 - Browser: Chromium, Firefox
 - Version: 72, 65

I don't know how to use storybook.

**Additional context**

Let me know if I can do anything to help.

danedavid:
Fixes #1535 .

### DIFF
--- a/src/components/SingleDatePicker.jsx
+++ b/src/components/SingleDatePicker.jsx
@@ -248,7 +248,7 @@ class SingleDatePicker extends React.PureComponent {
 
   onFocusOut(e) {
     const { onFocusChange } = this.props;
-    if (this.container.contains(e.relatedTarget || e.target)) return;
+    if (this.dayPickerContainer.contains(e.relatedTarget || e.target)) return;
     onFocusChange({ focused: false });
   }
 
--- a/src/components/SingleDatePicker.jsx
+++ b/src/components/SingleDatePicker.jsx
@@ -248,7 +248,12 @@ class SingleDatePicker extends React.PureComponent {
 
   onFocusOut(e) {
     const { onFocusChange } = this.props;
-    if (this.dayPickerContainer.contains(e.relatedTarget || e.target)) return;
+    if (
+      this.dayPickerContainer
+      && this.dayPickerContainer.contains(e.relatedTarget || e.target)
+    ) {
+      return;
+    }
     onFocusChange({ focused: false });
   }
 
--- a/src/components/SingleDatePicker.jsx
+++ b/src/components/SingleDatePicker.jsx
@@ -248,12 +248,7 @@ class SingleDatePicker extends React.PureComponent {
 
   onFocusOut(e) {
     const { onFocusChange } = this.props;
-    if (
-      this.dayPickerContainer
-      && this.dayPickerContainer.contains(e.relatedTarget || e.target)
-    ) {
-      return;
-    }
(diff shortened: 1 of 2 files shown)


ljharb:
This appears to be failing tests.

danedavid:
@ljharb Ok, let me check and get back to you

danedavid:
@ljharb Updated the PR.

src/components/SingleDatePicker.jsx
```diff
@@ -248,7 +248,12 @@ class SingleDatePicker extends React.PureComponent {
 
   onFocusOut(e) {
     const { onFocusChange } = this.props;
-    if (this.container.contains(e.relatedTarget || e.target)) return;
+    if (
+      this.dayPickerContainer
+      && this.dayPickerContainer.contains(e.relatedTarget || e.target)
```
ljharb: I’m not sure this is the right change - is it possible the relevant tests were mocking this.container but need to be updated to mock this.dayPickerContainer?

danedavid: Even I thought of that. The fact is that there are always 2 refs - the outer ref `this.container`, and the inner ref `this.dayPickerContainer`. The issue was caused because the wrong container ref was being searched in case of elements rendered in portal. So I referred to `DateRangePicker.jsx`, where I saw that `this.dayPickerContainer` was the one being used in `focusOut` handler. And thus the change.
So should we update the tests, or is the change wrong?

ljharb: Your change might be perfect :-) but maybe you could add tests that would help ensure it's not broken in the future?

danedavid: I've changed the ref being used inside the test, and reverted the previous check. @ljharb

monokrome:
@majapw Not sure that my fix didn't work. Seems unrelated to me - which use case wasn't working in 19.0.2?

majapw:
@monokrome Sorry I wasn't clear! This issue seemed very similar to https://github.com/airbnb/react-dates/issues/1522 (clicking anywhere closes the SingleDatePicker) as it was related to the SDP autoclosing and the `appendToBody` prop. This seemed like a specific edgecase of that (the description of the use case is in the linked issue) and both were rooted in the v18.4.1 changes. Thanks for taking a look!

PR #1553, opened 2019-02-19
Stop calendar blinking on DateRangePickerInput focus switch (fixes #1523)

### LINKED ISSUE #1523: DateRangePicker blinking after select startDate
Hello, there.
I'm  setting up DateRangePicker, but when I select startDate the calendar blinks. 

screencast-localhost-3000-2019 02 01-11-33-13

Here's my code

```
<DateRangePicker
        startDateId="startDate" 
        endDate={this.state.endDate} 
        endDateId="endDate" 
        onDatesChange={({ startDate, endDate }) => this.setState({startDate, endDate})}
        onFocusChange={focusedInput => this.setState({focusedInput})} 
      />
```

GfxKai:
Using `isStartDateFocused` and `isEndDateFocused` to conditionally render `children` causes the calendar to disappear and reappear whenever the focus switches from the start date input to end date input.

This pr adopts same approach as dd9d6b27b300dc8e31d89fcd982d525845e9c698 by skipping the focus check and just rendering the children.

### DIFF
--- a/src/components/DateRangePickerInput.jsx
+++ b/src/components/DateRangePickerInput.jsx
@@ -190,7 +190,11 @@ function DateRangePickerInput({
     />
   );
 
-  const screenReaderText = screenReaderMessage || phrases.keyboardNavigationInstructions;
+  const screenReaderStartDateText = screenReaderMessage
+    || phrases.keyboardForwardNavigationInstructions;
+  const screenReaderEndDateText = screenReaderMessage
+    || phrases.keyboardBackwardNavigationInstructions;
+
   const inputIcon = (showDefaultInputIcon || customInputIcon !== null) && (
     <button
       {...css(styles.DateRangePickerInput_calendarIcon)}
@@ -224,7 +228,7 @@ function DateRangePickerInput({
         placeholder={startDatePlaceholderText}
         ariaLabel={startDateAriaLabel}
         displayValue={startDate}
-        screenReaderMessage={screenReaderText}
+        screenReaderMessage={screenReaderStartDateText}
         focused={isStartDateFocused}
         isFocused={isFocused}
         disabled={startDateDisabled}
@@ -242,6 +246,8 @@ function DateRangePickerInput({
         regular={regular}
       />
 
+      {children}
+
       {
         <div
           {...css(styles.DateRangePickerInput_arrow)}
@@ -252,14 +258,12 @@ function DateRangePickerInput({
         </div>
       }
 
-      {isStartDateFocused && children}
-
       <DateInput
         id={endDateId}
         placeholder={endDatePlaceholderText}
         ariaLabel={endDateAriaLabel}
         displayValue={endDate}
-        screenReaderMessage={screenReaderText}
+        screenReaderMessage={screenReaderEndDateText}
         focused={isEndDateFocused}
         isFocused={isFocused}
         disabled={endDateDisabled}
@@ -277,7 +281,6 @@ function DateRangePickerInput({
         regular={regular}
       />
 
-      {isEndDateFocused && children}
 
       {showClearDates && (
         <button
--- a/src/components/SingleDatePickerInput.jsx
+++ b/src/components/SingleDatePickerInput.jsx
@@ -139,7 +139,7 @@ function SingleDatePickerInput({
     />
   );
 
-  const screenReaderText = screenReaderMessage || phrases.keyboardNavigationInstructions;
+  const screenReaderText = screenReaderMessage || phrases.keyboardForwardNavigationInstructions;
   const inputIcon = (showDefaultInputIcon || customInputIcon !== null) && (
     <button
       {...css(styles.SingleDatePickerInput_calendarIcon)}
--- a/src/defaultPhrases.js
+++ b/src/defaultPhrases.js
@@ -22,8 +22,8 @@ const moveFocusByOneWeek = 'Move backward (up) and forward (down) by one week.';
 const moveFocusByOneMonth = 'Switch months.';
 const moveFocustoStartAndEndOfWeek = 'Go to the first or last day of a week.';
 const returnFocusToInput = 'Return to the date input field.';
-const keyboardNavigationInstructions = `Press the down arrow key to interact with the calendar and
-  select a date. Press the question mark key to get the keyboard shortcuts for changing dates.`;
+const keyboardForwardNavigationInstructions = 'Navigate forward to interact with the calendar and select a date. Press the question mark key to get the keyboard shortcuts for changing dates.';
+const keyboardBackwardNavigationInstructions = 'Navigate backward to interact with the calendar and select a date. Press the question mark key to get the keyboard shortcuts for changing dates.';
 
 const chooseAvailableStartDate = ({ date }) => `Choose ${date} as your check-in date. It’s available.`;
 const chooseAvailableEndDate = ({ date }) => `Choose ${date} as your check-out date. It’s available.`;
@@ -58,7 +58,8 @@ export default {
   moveFocusByOneMonth,
   moveFocustoStartAndEndOfWeek,
   returnFocusToInput,
-  keyboardNavigationInstructions,
+  keyboardForwardNavigationInstructions,
+  keyboardBackwardNavigationInstructions,
 
   chooseAvailableStartDate,
   chooseAvailableEndDate,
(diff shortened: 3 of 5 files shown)


burtek:
@GfxKai are you working on this PR?

GfxKai:
@burtek it only involves 2 very minor changes, so no idea why the travis build failed the first time round but I think that blocked review / merging. I've just updated the branch and the builds have now passed, so hopefully someone can pick this up and merge it!

burtek:
@GfxKai great! @ljharb @majapw any chance for this PR? 🤔

burtek:
BTW, why render `children` twice?

GfxKai:
@burtek I'm wondering the same thing myself! 😅 good spot, will correct that now.

@ljharb happy to give that a go. just to clarify: are you looking for an enzyme test to check that the calendar stays visible when focus switches between the inputs?

ljharb:
@GfxKai yes - you’re making conditional renderings unconditional, so there should be a test to ensure it doesn’t regress (the commit you referenced was fixed due to a test catching it)

src/components/DateRangePickerInput.jsx
```diff
@@ -269,7 +267,7 @@ function DateRangePickerInput({
         regular={regular}
       />
 
-      {isEndDateFocused && children}
```
ljharb: This changes the rendering order - are you sure this doesn’t have any impact?

GfxKai: Not 100% but the original conditional rendering logic changes the rendering order depending on which input is focused and the component seems fully functional in either case. I've done some manual testing and it appears to work perfectly with the prop in either position.

It might actually make more sense to move `{children}` down to be the last child of the div, since the other elements are all related to the input box and are positioned in a row together, whilst the calendar picker is absolute-positioned below.

ljharb: presumably there was some reason that the two `children`s were not in the same place in the beginning - i'm loath to change that without understanding why.

GfxKai: @monokrome any chance you could shed some light on the reasoning behind https://github.com/airbnb/react-dates/commit/06407e1495761128a935c5169d3f56e8893d3361 ?

GfxKai: After a bit more playing around, I think I've worked out what's going on. If `{children}` is rendered before the second input, pressing tab moves the focus to DayPicker to make picking the start date possible with the keyboard. If the prop is moved after the end date input, pressing tab when the start date input is focused moves focus straight to the end date input.

Could we emulate this focusing behaviour using onTab event handlers rather than relying on the DayPicker's position in the DOM? I imagine this would require logic to: 

1. Focus the DayPicker on tab when the start date input is focused
2. Focus the end date input on tab when start date input is focused **and** the DayPicker's next month button is focused
3. Focus the DayPicker on tab when the end date input is focused

GfxKai: Otherwise, navigating to the DayPicker via the down arrow is already supported and works regardless of the position of `{children}` here. This is the behaviour documented in the screen reader message: "Press the down arrow key to interact with the calendar and select a date. Press the question mark key to get the keyboard shortcuts for changing dates."

ljharb: No, i believe the best approach is always to rely on proper DOM ordering.

GfxKai: Agreed. In that case, I'd say having to stick with using the down arrow to interact with the DayPicker is better than being able to use tab and having the calendar unmounting and re-mounting on every input focus change.

I think unconditionally rendering `{children}` as the last child in the div makes the most sense, as the tabbing order would follow both the visual order and DOM order: focusing start date, end date, the clear dates button (if visible) and **then** the DayPicker, whilst the DayPicker is accessible at any point via the down arrow. Does that sound reasonable to you?

ljharb: @monokrome @majapw @backwardok, what do you think? I'll defer to yall here.

backwardok: We actually want to move away from the "use the down arrow key" functionality since it's not something that's obvious to a keyboard user. Additionally, for any screen reader user who is not using a keyboard (such as on a mobile device), that direction doesn't make any sense.

Similarly, we can't rely on using tab listeners to move focus to the calendar because not all screen reader users will be navigating using the tab key, even if they're using a keyboard to navigate.

One of the main reasons why it doesn't work well to have the calendar be after all elements (the way it was before) is that if you were to tab through everything to get to the calendar, you'd actually be setting the end date instead of the start date since the end date was the last field that had focus. Setting the end date in the calendar first then puts focus back on the start date field. If you then tried to tab forward again to the calendar, you'd be setting the end date again and never be able to set the start date from the calendar.

GfxKai: Perhaps the desired behaviour could be achieved by dynamically defining the tabIndex property for the relevant input and button components, such that the order of keyboard navigation is defined depending on isStartDateFocused and isEndDateFocused. This way, the navigation order can be controlled without having to unmount and remount the DayPicker

backwardok: That was something that @monokrome had considered before, but the biggest problem with using tabindex to handle navigation is it doesn't quite work the way you may expect it to. Setting tabindex to a positive value essentially moves it earlier in natural tab flow. For an entire document, if all the tab-able elements were in an order list based on the DOM order, a positive tabindex value would move from their original spot in that list to the beginning of the list. What that means is that if you have focus on something that is in the natural order, moving focus forward won't get you to the positive tabindex elements. You'd have to manually set the tabindex for every element on the page, which would be a maintenance nightmare (and fairly fragile). 

I created this codepen to demonstrate that: https://codepen.io/backwardok/pen/eQbYEa

GfxKai: Understood. Perhaps the DayPicker could be unconditionally rendered just after the first day picker? This isn't a perfect solution, as it requires a shift-tab to go back and set the end date, but this might potentially make more sense than the DayPicker moving itself to be in a different tab position than before.

If this is not a viable alternative, I would have to argue that the re-rendering of the calendar is a much bigger issue than having to use the slightly unintuitive down arrow for keyboard navigation. The unmount and remount of the calendar is not just undesirable because of the flashing but, depending on the chosen start date, this behaviour can change which months are visible and move the location of the initially selected date to a different panel. This is extremely confusing and could easily result in users picking the wrong set of dates without realising.

Whilst I understand the need to strive for better accessibility, I think any solution that results in this re-rendering issue must be ruled out, as it has a substantially negative impact on the UX for **all** users 😬

backwardok: @GfxKai interesting solution! That might be a decent workaround for now, and solves some of the other concerns. For screen reader users, we could include some extra information for the checkout field that says something like, "navigate backward to interact with the calendar". It's not a perfect solution, since I don't know if it'll be as obvious to a sighted keyboard user that they could access the calendar by navigating backward and not forward, but I think that's a decent option, and one that still addresses the original problem.

GfxKai:
@backwardok I've had a go of implementing the discussed solution. How does that look now?

src/components/DateRangePickerInput.jsx
```diff
@@ -235,6 +236,8 @@ function DateRangePickerInput({
         regular={regular}
       />
 
+      {children}
```
backwardok: should this be checking if either one is focused? seems like this would result in the calendar always showing up, even when the fields are blurred.

GfxKai: {children} here is the DayPicker, passed through all the way from DateRangePicker, where it is conditionally rendered based on `maybeRenderDayPickerWithPortal()`. This function renders null [here](https://github.com/GfxKai/react-dates/blob/master/src/components/DateRangePicker.jsx#L372-L374) if `isOpened()`, which checks whether one of the two inputs are focused, returns false. Hence, the focus check is done in the parent and we can just render unconditionally here and not worry 🎉

patricklafrance:
Hi!

Will this be merged soon?

Thank you

ljharb:
Rebased. This could use some tests for the `SingleDatePickerInput.jsx` changes.

dan-kez:
Hi all - just checking if there is any ability to move this PR forward. We can't upgrade to the latest version of react-dates due to this issue.

ljharb:
It first needs another rebase, and also tests for the `SingleDatePickerInput.jsx` changes.

davidyeiser:
Hello, also wondering if this is planned to be released anytime soon?

monokrome:
@majapw

ljharb:
cc @GfxKai https://github.com/airbnb/react-dates/pull/1553#issuecomment-499228353

GfxKai:
Sorry guys, been extremely busy with the day job. I’ll see if I can get onto this at some point in the next few weeks. If anyone has time to spare, feel free to jump in and write those tests!

EugeneDraitsev:
@GfxKai I forked your repository and added tests for `SingleDatePickerInput.jsx` changes: https://github.com/EugeneDraitsev/react-dates (Latest commit in master, master is fully merged to your `patch-1` and `airbnb/master` branches).  You can pull my changes if they are relevant

PR #1555, opened 2019-02-21
Fix VO issue for date ranges

monokrome:
This fixes an issue where VoiceOver and other accessibility tools don't let people know that dates are selected when they are somewhere between the start and end date of a date range/span.

### DIFF
--- a/src/utils/getCalendarDaySettings.js
+++ b/src/utils/getCalendarDaySettings.js
@@ -1,6 +1,24 @@
 import getPhrase from './getPhrase';
 import { BLOCKED_MODIFIER } from '../constants';
 
+function isSelected(modifiers) {
+  return modifiers.has('selected')
+  || modifiers.has('selected-span')
+  || modifiers.has('selected-start')
+  || modifiers.has('selected-end');
+}
+
+function shouldUseDefaultCursor(modifiers) {
+  return modifiers.has('blocked-minimum-nights')
+  || modifiers.has('blocked-calendar')
+  || modifiers.has('blocked-out-of-range');
+}
+
+function isHoveredSpan(modifiers) {
+  if (isSelected(modifiers)) return false;
+  return modifiers.has('hovered-span') || modifiers.has('after-hovered-start');
+}
+
 function getAriaLabel(phrases, modifiers, day, ariaLabelFormat) {
   const {
     chooseAvailableDate,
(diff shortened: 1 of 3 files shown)


monokrome:
@ljharb Apologies for confusion - this isn't ready for review yet. Does creating a new PR cause it to be considered ready for review? Sorry if you got pinged about this or anything, as I know it's late.

ljharb:
@monokrome there's a new "draft" status you could have used; but either way no worries! ping me when it's ready for review.

monokrome:
@ljharb How do you enable the draft status? Is it a label, a new feature in GitHub, or something else?

ljharb:
@monokrome it's (recently) in a dropdown on the "create pull request" button; non-draft PRs can't be made drafts.

majapw:
In a more old-school way, you can also just clarify in the PR description and add a WIP tag to the PR! :)

monokrome:
@ljharb I added a test (was going to do this anyway) last night, but not 100% sure what other tests I should add. Specifically, I think that we *should* have tests to ensure the styles are applied appropriately to `CalendarDay` but I also don't want to introduce them in this PR since we don't have those tests yet for the already-existing logic. So, I may make another PR for that stuff if it feels like something we should invest in, which I believe it is.

majapw:
Can you add a bit of context to the description as to what this is fixing? QQ: are you intending to follow up on adding a phrase for within range in this PR or in a follow-up (I don't want to merge in early if so)?

monokrome:
@majapw added

monokrome:
@ljharb I'm confused. I'm merged because it said you approved, but I just noticed that it says you requested a review from maja. Do we expect more than one approving review to merge?

monokrome:
(I'll revert if so)

majapw:
If it's stamped, it's stamped! def fine to merge.

ljharb:
@monokrome we don't have any such policy here, was more thinking that it's always better to get more eyes :-) if it was a blocker tho, i'd not have stamped.

src/utils/getCalendarDaySettings.js
```diff
@@ -10,51 +28,34 @@ export default function getCalendarDaySettings(day, ariaLabelFormat, daySize, mo
...
-    } else if (modifiers.has('selected-end') && dateIsSelectedAsEndDate) {
-      ariaLabel = getPhrase(dateIsSelectedAsEndDate, formattedDate);
-    } else {
-      ariaLabel = getPhrase(dateIsSelected, formattedDate);
-    }
-  } else if (modifiers.has(BLOCKED_MODIFIER)) {
-    ariaLabel = getPhrase(dateIsUnavailable, formattedDate);
+  if (modifiers.has('selected-start') && dateIsSelectedAsStartDate) {
+    return getPhrase(dateIsSelectedAsStartDate, formattedDate);
+  } if (modifiers.has('selected-end') && dateIsSelectedAsEndDate) {
```
backwardok: I think you're missing the `else` part of this logic

backwardok: oh wait I just noticed that you're returning so this ends up being technically ok, but visually confusing (because it looks like an else if but it's actually a standalone if statement

ljharb: whoops, that'd be great to clean up in a followup :-) @monokrome

monokrome: @backwardok hmm... Think I should use an `else`? Feels bad to use `let` when we don't need it, but also feels worse to have a final return in a branch instead of in the function's root block

ljharb: It shouldn't use an else; but a newline should be inserted between `}` and `if`.

PR #1562, opened 2019-02-25
fix(DayPickerSingleDateController): Add missing modifiers when loading more months in `VERTICAL_SCROLLABLE` orientation

jtomaszewski:
### Bug

Right now, when we use `DayPickerSingleDateController with `orientation: 'verticalScrollable'`, when i.e. your start date is 15 Feb 2019 and have `numberOfMonths: 3`, it will correctly show additional months from May to July, but when you want to hover with mouse any month days in June (interesting note: it happens only for the 2nd appended month, so no May or July in this example), it will return bunch of errors like those:

image

```
DayPickerSingleDateController.js:781 Uncaught TypeError: Cannot read property '2019-06-01' of undefined
    at DayPickerSingleDateController.addModifier (DayPickerSingleDateController.js:781)
    at DayPickerSingleDateController.onDayMouseEnter (DayPickerSingleDateController.js:525)
    at CalendarDay.onDayMouseEnter (CalendarDay.js:165)
    at onMouseEnter (CalendarDay.js:236)
    at HTMLUnknownElement.callCallback (react-dom.development.js:149)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:199)
    at invokeGuardedCallback (react-dom.development.js:256)
    at invokeGuardedCallbackAndCatchFirstError (react-dom.development.js:270)
    at executeDispatch (react-dom.development.js:561)
    at executeDispatchesInOrder (react-dom.development.js:583)
```

### Fix

I noticed that even though DayPickerSingleDateController and DayPickerRangeController have very similar logic to each other, the DateRangePickerInputController has the logic for updating modifiers in `onMultiplyScrollableMonths` ( https://github.com/airbnb/react-dates/blob/master/src/components/DayPickerRangeController.jsx#L779-L793 ), but DayPickerSingleDateController lacks it.

So basing on that code, I added this functionality (and test) to DayPickerSingleDateController as well, using almost the same code.

### DIFF
--- a/src/components/DayPickerSingleDateController.jsx
+++ b/src/components/DayPickerSingleDateController.jsx
@@ -183,6 +183,7 @@ export default class DayPickerSingleDateController extends React.PureComponent {
     this.onMonthChange = this.onMonthChange.bind(this);
     this.onYearChange = this.onYearChange.bind(this);
 
+    this.onMultiplyScrollableMonths = this.onMultiplyScrollableMonths.bind(this);
     this.getFirstFocusableDay = this.getFirstFocusableDay.bind(this);
   }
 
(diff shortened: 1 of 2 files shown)


gdaolewe:
We've also run into this. Other than resolving conflicts, what is still needed to get this fix merged?

ljharb:
I've rebased this, but it seems like this was added in #1894?

PR #1563, opened 2019-02-25
Focus ring now shows up around custom navigation buttons

nkinser:
This PR fixes an issue where when custom navigation buttons are provided, the focus ring for keyboard-only users did not appear around the custom buttons. This issue is fixed by only supplying a `tabindex` when the default navigation buttons are used. This means that in order for custom navigation buttons to receive keyboard focus a `tabindex` must be specified on the object passed in. 

#### Before
calendar_tabbing_before

#### After
calendar_tabbing_after

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,6 @@ yarn.lock
 package-lock.json
 
 css/styles.css
+
+# Cache
+.cache/
--- a/src/components/DayPickerNavigation.jsx
+++ b/src/components/DayPickerNavigation.jsx
@@ -128,7 +128,7 @@ function DayPickerNavigation({
       )}
     >
       {!isVerticalScrollable && (
-        <div
+        <div // eslint-disable-line jsx-a11y/interactive-supports-focus
           role="button"
           {...navPrevTabIndex}
           {...css(
@@ -166,7 +166,7 @@ function DayPickerNavigation({
         </div>
       )}
 
-      <div
+      <div // eslint-disable-line jsx-a11y/interactive-supports-focus
         role="button"
         {...navNextTabIndex}
         {...css(
--- a/src/components/DayPickerNavigation.jsx
+++ b/src/components/DayPickerNavigation.jsx
@@ -70,7 +70,10 @@ function DayPickerNavigation({
   let navNextIcon = navNext;
   let isDefaultNavPrev = false;
   let isDefaultNavNext = false;
+  let navPrevTabIndex = {};
+  let navNextTabIndex = {};
   if (!navPrevIcon) {
+    navPrevTabIndex = { tabIndex: '0' };
     isDefaultNavPrev = true;
     let Icon = isVertical ? ChevronUp : LeftArrow;
     if (isRTL && !isVertical) {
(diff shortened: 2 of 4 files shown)


.gitignore
```diff
@@ -47,3 +47,6 @@ yarn.lock
 package-lock.json
 
 css/styles.css
+
+# Cache
+.cache/
```
ljharb: what is this a cache of? what produces it?

src/components/DayPickerNavigation.jsx
```diff
@@ -126,7 +126,7 @@ function DayPickerNavigation({
       {!isVerticalScrollable && (
         <div
           role="button"
-          tabIndex="0"
+          tabIndex={isDefaultNavPrev ? '0' : null}
```
ljharb: does `null` work, or does it have to be `undefined`?

nkinser: I wasn't able to find anything very helpful on google about what the value should be, so to be safe, I think I'll go with [this stack overflow ](https://stackoverflow.com/questions/36523225/dynamic-tabindex-attribute-in-jsx-react)solution

.gitignore
nkinser: I'm not sure what produces it, it just always appears in my staged files whenever I work in react dates. Currently the folder contains one file whose contents are:
```{"value":{"success":true,"data":{"next":{"version":"5.0.0-rc.6","info":{"plain":"### Bug Fixes\n\n* Addon-actions: FIX performance by upgrading to telejson 2.1 ([#5751](https://github.com/storybooks/storybook/pull/5751))\n* UI: FIX bad treeview mockdata ([#5741](https://github.com/storybooks/storybook/pull/5741))\n* UI: About page styling fixes ([#5732](https://github.com/storybooks/storybook/pull/5732))\n* UI: Restore the toolbar eject button ([#5737](https://github.com/storybooks/storybook/pull/5737))"}},"latest":{"version":"4.1.13","info":{"plain":"### Documentation\n\n* Fix unkind addon-ondevice-knobs readme. ([#5629](https://github.com/storybooks/storybook/pull/5629))\n\n### Maintenance\n\n* Cleanup CLI version notice ([#5699](https://github.com/storybooks/storybook/pull/5699))\n* Use static versions.json file instead of hacking one in ([#5675](https://github.com/storybooks/storybook/pull/5675))"}}},"time":1551129533811},"type":"Object"}```

ljharb: hm, maybe something in storybook. cool to add then!

nkinser:
@ljharb How would I add regression tests for this?
I do think this could be considered a breaking change if anyone supplies custom navigation buttons, this new change will cause the buttons to be skipped over with keyboard navigation.

ljharb:
Hmm, that's a fair point.

As for a test, I'd assume we could provide one test that uses default buttons, and asserts the tabIndex is present, and another that uses custom ones, and asserts that it's not?

Even better would be adding those test cases to master first, and then updating them in this PR, which would make the breakage very clear.

nkinser:
@ljharb Would it be possible to check in `DayPickerNavigation` if the supplied custom buttons have a `tabIndex` and then only omit the `tabIndex` prop if the custom buttons have it?

ljharb:
Not reliably, no - all we can do is render what they give us.

nkinser:
@ljharb do the test changes look good to you?

PR #1572, opened 2019-03-07
Enable treeShaking import in webpack

samMeow:
As webpack (create react app) have enable treeshaking feature to minimize the bundling size.
(https://webpack.js.org/guides/tree-shaking/)
(https://github.com/webpack/webpack/blob/master/lib/optimize/SideEffectsFlagPlugin.js)

It would be good if react-dates can support this feature

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,6 +3,11 @@
   "version": "20.0.0",
   "description": "A responsive and accessible date range picker component built with React",
   "main": "index.js",
+  "module": "esm/index.js",
+  "sideEffects": [
+    "*.css",
+    "*initialize.js"
+  ],
   "scripts": {
     "prebuild": "npm run clean",
     "build": "npm run build:cjs && npm run build:esm && npm run build:css -- --optimize ",


package.json
```diff
@@ -3,6 +3,8 @@
   "version": "20.0.0",
   "description": "A responsive and accessible date range picker component built with React",
   "main": "index.js",
+  "module": "esm/index.js",
+  "sideEffects": false,
```
ljharb: this is inaccurate; some of our files do have side effects.

samMeow: Thanks for pointing this out. I have just update to `initialize.js` and all css 
using Regex `^\w+\(` to find all self executed function. Hope I didn't miss any.

samMeow:
@ljharb I just curious and want to know more about the usage of `module` field in nodeJS. I have try my best to do some research. 
So the original proposal capture in Nov 2016 where one of the approach is package opt-in [original](https://github.com/nodejs/node-eps/blob/4217dca299d89c8c18ac44c878b5fe9581974ef3/002-es6-modules.md#51-determining-if-source-is-an-es-module)
Then after a long [discussion](https://github.com/nodejs/node-eps/issues/13) around how to determine if a module is es6 or commonjs. Basically arguments around package opt-in / different file extension, etc approach is will it add extra complexity or  extra cost for developer and module system.
Then later the proposal [now/ Aug 2017](https://github.com/nodejs/node-eps/blob/master/002-es-modules.md#44-shipping-both-esm-and-cjs) use `mjs` file extension. 
[Node v11.11.0 experimental es module](https://nodejs.org/api/esm.html)

ljharb:
Yes, the default is going to be the mjs file extension, but there may be additional mechanisms, including a package.json field. What it will *never* be is a string “module” field, because of webpack and other tools poisoning it.

samMeow:
Got you. I will close this merge request as it (module field/ package opt-in) violate the standard node package json purpose in this moment.

PR #1576, opened 2019-03-08
Customize DayPickerKeyboardShortcuts with renderKeyboardShortcutsButton

kypan:
Introducing a functional prop renderKeyboardShortcutsButton to render a custom styled button.

### DIFF
--- a/src/components/DayPickerKeyboardShortcuts.jsx
+++ b/src/components/DayPickerKeyboardShortcuts.jsx
@@ -180,14 +180,7 @@ class DayPickerKeyboardShortcuts extends React.PureComponent {
     return (
       <div>
         { renderKeyboardShortcutsButton
-          ? React.cloneElement(
-            renderKeyboardShortcutsButton(),
-            {
-            // can't pass context-specific callbacks from passed-in button
-              ref: this.setShowKeyboardShortcutsButtonRef,
-              onClick: this.onShowKeyboardShortcutsButtonClick,
-            },
-          ) : (
+          ? renderKeyboardShortcutsButton() : (
             <button
               ref={this.setShowKeyboardShortcutsButtonRef}
               {...css(
--- a/src/components/DayPickerKeyboardShortcuts.jsx
+++ b/src/components/DayPickerKeyboardShortcuts.jsx
@@ -179,8 +179,15 @@ class DayPickerKeyboardShortcuts extends React.PureComponent {
 
     return (
       <div>
-        { renderKeyboardShortcutsButton ? renderKeyboardShortcutsButton()
-          : (
+        { renderKeyboardShortcutsButton
+          ? React.cloneElement(
+            renderKeyboardShortcutsButton(),
+            {
+            // can't pass context-specific callbacks from passed-in button
+              ref: this.setShowKeyboardShortcutsButtonRef,
+              onClick: this.onShowKeyboardShortcutsButtonClick,
+            },
+          ) : (
             <button
               ref={this.setShowKeyboardShortcutsButtonRef}
               {...css(
--- a/src/components/DayPickerKeyboardShortcuts.jsx
+++ b/src/components/DayPickerKeyboardShortcuts.jsx
@@ -22,6 +22,7 @@ const propTypes = forbidExtraProps({
   openKeyboardShortcutsPanel: PropTypes.func,
   closeKeyboardShortcutsPanel: PropTypes.func,
   phrases: PropTypes.shape(getPhrasePropTypes(DayPickerKeyboardShortcutsPhrases)),
+  renderKeyboardShortcutsButton: PropTypes.func,
 });
 
 const defaultProps = {
@@ -31,6 +32,7 @@ const defaultProps = {
   openKeyboardShortcutsPanel() {},
   closeKeyboardShortcutsPanel() {},
   phrases: DayPickerKeyboardShortcutsPhrases,
+  renderKeyboardShortcutsButton: undefined,
 };
 
 function getKeyboardShortcuts(phrases) {
@@ -164,6 +166,7 @@ class DayPickerKeyboardShortcuts extends React.PureComponent {
       closeKeyboardShortcutsPanel,
       styles,
       phrases,
+      renderKeyboardShortcutsButton,
     } = this.props;
 
     const toggleButtonText = showKeyboardShortcutsPanel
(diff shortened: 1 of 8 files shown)


src/components/DayPickerKeyboardShortcuts.jsx
```diff
@@ -176,33 +179,36 @@ class DayPickerKeyboardShortcuts extends React.PureComponent {
...
-        >
-          <span
-            {...css(
-              styles.DayPickerKeyboardShortcuts_showSpan,
-              bottomRight && styles.DayPickerKeyboardShortcuts_showSpan__bottomRight,
-              topRight && styles.DayPickerKeyboardShortcuts_showSpan__topRight,
-              topLeft && styles.DayPickerKeyboardShortcuts_showSpan__topLeft,
-            )}
-          >
+        { renderKeyboardShortcutsButton
```
monokrome: This is a nit, but I think that the code style in react-dates is to do avoid ternaries by creating two conditions instead?

```es7
 { renderKeyboardShortcutsButton && .... }
 { renderKeyboardShortcutsButton || .... }
```

src/components/DayPickerKeyboardShortcuts.jsx
```diff
@@ -176,33 +179,36 @@ class DayPickerKeyboardShortcuts extends React.PureComponent {
...
+              ref={this.setShowKeyboardShortcutsButtonRef}
+              {...css(
+                styles.DayPickerKeyboardShortcuts_buttonReset,
+                styles.DayPickerKeyboardShortcuts_show,
+                bottomRight && styles.DayPickerKeyboardShortcuts_show__bottomRight,
+                topRight && styles.DayPickerKeyboardShortcuts_show__topRight,
+                topLeft && styles.DayPickerKeyboardShortcuts_show__topLeft,
+              )}
+              type="button"
+              aria-label={toggleButtonText}
```
kypan: I might have to change this to `this.getToggleButtonText()` and bind to this scope since it relies on state change in `DayPicker.jsx`

monokrome: I'm confused, what is being referred to as "custom button" here? Seems like `aria-label` would work here unless I misunderstand?

kypan: @monokrome `toggleButtonText` here relies on `showKeyboardShortcuts` which is a state variable set in the parent `DayPicker.jsx`. At first, I didn't think there was a way for me to pass in a similar variable through `renderKeyboardsShortcutsButton` that can read the state of `DayPicker.jsx` unless I bound it somehow.

But actually though, I just realized that I can pass in `toggleButtonText` as a param into the `renderKeyboardShortcutsButton()` here and drill that back into the render logic. Turns out I don't need to change any logic here. Hope that made sense!

src/components/DayPickerKeyboardShortcuts.jsx
ljharb: (in jsx in general)

ljharb: also, jsx curlies should have no bounding spaces inside them

test/components/DayPickerKeyboardShortcuts_spec.jsx
```diff
@@ -135,6 +135,15 @@ describe('DayPickerKeyboardShortcuts', () => {
           expect(mockEvent.currentTarget.blur.callCount).to.equal(1);
         });
       });
+
+      describe('renderKeyboardShortcutsButton', () => {
+        it('renders the provided button', () => {
+          const props = { renderKeyboardShortcutsButton: () => (<button type="button">Success!</button>) };
+          const wrapper = shallow(<DayPickerKeyboardShortcuts {...props} />).dive();
+          const buttonWrapper = wrapper.children().find('button');
```
ljharb: possibly a more robust test would be to render a custom component, and `expect(wrapper.find(Custom)).to.have.lengthOf(1))`, rather than duplicating the text and element name here?

src/components/DayPickerKeyboardShortcuts.jsx
monokrome: Cool! Thanks so much for explaining to me 😺

test/components/DayPickerKeyboardShortcuts_spec.jsx
kypan: The use case for renderKeyboardShortcutsButton is to literally send in a <button> though, should I still create a spec for a React Component?

ljharb: @kypan i mean, the purpose is to ensure that *whatever* `renderKeyboardShortcutsButton` returns is what's rendered; the only way to guarantee that is to use an unforgeable sentinel value - which would thus be a custom component.

kypan: @ljharb That definitely makes sense. I've updated it. Thanks!

src/components/DayPickerKeyboardShortcuts.jsx
```diff
@@ -176,34 +179,43 @@ class DayPickerKeyboardShortcuts extends React.PureComponent {
...
-          <span
+        {renderKeyboardShortcutsButton
+          && renderKeyboardShortcutsButton({
+            // passing in context-specific props
+            ref: this.setShowKeyboardShortcutsButtonRef,
+            onClick: this.onShowKeyboardShortcutsButtonClick,
+            ariaLabel: toggleButtonText,
+          })}
+        {renderKeyboardShortcutsButton
+          || (
```
ljharb: the `|| (` should be on the previous line; there’s no need to have a line break before it

src/components/DayPickerKeyboardShortcuts.jsx
```diff
@@ -176,34 +179,43 @@ class DayPickerKeyboardShortcuts extends React.PureComponent {
...
+                topLeft && styles.DayPickerKeyboardShortcuts_showSpan__topLeft,
+              )}
+            >
             ?
-          </span>
-        </button>
-
+            </span>
+          </button>
+          )}
```
ljharb:
```suggestion
        )}
```

stories/DayPickerRangeController.js
```diff
@@ -68,6 +68,49 @@ const TestCustomInfoPanel = () => (
   </div>
 );
 
+function renderKeyboardShortcutsButton(buttonProps) {
+  const { ref, onClick, ariaLabel } = buttonProps || {};
```
ljharb: Why would buttonProps need to be optional here?

If it is, it should be defaulted in the signature, no?

kypan: @ljharb I think when I wrote this I intended to make buttonProps optional to avoid future implementations breaking, and so I used the defaulting pattern I saw in monorail. But if it's cool with the `react-dates` squad I can just make it required

ljharb: Seems better to require everything until it needs to be optional.

kypan: Cool, I made it required

kypan:
@ljharb Added specs for DayPicker and DayPickerRangeController

stories/DayPickerRangeController.js
```diff
@@ -219,9 +262,9 @@ storiesOf('DayPickerRangeController', module)
         orientation={VERTICAL_SCROLLABLE}
         numberOfMonths={3}
         verticalHeight={300}
-        navNext={
+        navNext={(
           <div style={{ position: 'relative' }}>
-            <span
+  <span
```
ljharb:
```suggestion
            <span
```

stories/DayPickerRangeController.js
```diff
@@ -234,8 +277,8 @@ storiesOf('DayPickerRangeController', module)
             >
               Show More Months
             </span>
-          </div>
-        }
+</div>
```
ljharb:
```suggestion
          </div>
```

stories/DayPickerRangeController.js
```diff
@@ -234,8 +277,8 @@ storiesOf('DayPickerRangeController', module)
...
+)}
```
ljharb:
```suggestion
        )}
```

majapw:
Sick! This looks awesome. Thank you for the thorough reviews @monokrome and @ljharb

Issue #1575, opened 2019-03-08
Unable to have multiple datepickers in different locales


WarrickPardoe:
**react-dates version**
react-dates@20.0.0
moment@2.24.0

**The Bug**
This occurs when changing the global locale for the moments singleton , then creating a new instance of moments in a different locale and using that as the date prop. The month names change, however the week headers (DayPicker_weekHeader) stay in the locale used in the global moments singleton.

This makes it very difficult to have multiple locale versions of the date picker. I've managed to narrow down where I think the issue is:

DayPicker.jsx

```js
    const header = [];
    for (let i = 0; i < 7; i += 1) {
      header.push((
        <li key={i} {...css(styles.DayPicker_weekHeader_li, { width: daySize })}>
          <small>{moment().day((i + firstDayOfWeek) % 7).format(weekDayFormat)}</small>
        </li>
      ));
    }
```

Instead of using the passed in moments instance like currentMonth does, a new instance is being used with the moment's singleton that has the globally set locale.

Possible solutions is to use the passed in moments instance, or allowing a new prop to be passed in to choose how to render the week Headers.

**Source code (including props configuration)**
```js
import React, { useState } from "react"
import { SingleDatePicker } from "react-dates"
import moment from "moment"
import "moment/locale/fr"

function ReactDates() {
  // Create new instance of a moment with a FR locale
  const frMoment = moment().locale("fr")

  // Change the global moment locale singleton
  moment.locale("en")

  const [date1, setDate1] = useState(frMoment)
  const [date1Focus, setDate1Focus] = useState(false)

  const [date2, setDate2] = useState()
  const [date2Focus, setDate2Focus] = useState(false)

  return (
    <>
      <div>
        <SingleDatePicker
          date={date1}
          onDateChange={date => setDate1(date)}
          focused={date1Focus}
          onFocusChange={({ focused }) => setDate1Focus(focused)}
          id="date_one_example"
        />
      </div>
      <div>
        <SingleDatePicker
          date={date2}
          onDateChange={date => setDate2(date)}
          focused={date2Focus}
          onFocusChange={({ focused }) => setDate2Focus(focused)}
          id="date_two_example"
        />
      </div>
    </>
  )
}
```

**Screenshots/Gifs**
![image](https://user-images.githubusercontent.com/45460511/54017386-bda5fc00-417d-11e9-8da3-c903aae2d8dc.png)

**Desktop:**
 - OS: Windows 10
 - Browser: Chrome 
[...]

PR #1577, opened 2019-03-11
Pull Request for bug ticket #1575.

WarrickPardoe:
Pull Request for bug ticket #1575.

### DIFF
--- a/test/components/DayPicker_spec.jsx
+++ b/test/components/DayPicker_spec.jsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import moment from 'moment';
+import moment from 'moment/min/moment-with-locales';
 import { expect } from 'chai';
 import sinon from 'sinon-sandbox';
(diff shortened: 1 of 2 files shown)


WarrickPardoe:
Sure, will look to get it done this weekend

test/components/DayPicker_spec.jsx
```diff
@@ -16,6 +17,8 @@ import {
   VERTICAL_SCROLLABLE,
 } from '../../src/constants';
 
+moment.locale('en');
```
ljharb: this seems like a side effect that shouldn't run at module level - perhaps it should run in a global beforeEach?

WarrickPardoe:
I've updates the moment import to use the import with all the locales. I was having trouble trying to add the reset in the beforeEach callback, it was not resetting the locale correctly for some reason.

PR #1578, opened 2019-03-13
issue #1530 - Fix date selection in the SDP

### LINKED ISSUE #1530: SingleDatePicker doesn't fire onDateChange in IE11
**react-dates version**
e.g. react-dates@18.4.1

**Describe the bug**
In, IE11, SingleDatePicker renders and opens on focus. But when I click on a date, the calendar disappears without updating the date in the input. This works fine on other browsers.

 - OS: Windows
 - Browser IE
 - Version 11.0.9600.19236

https://airbnb.io/react-dates/?selectedKind=SingleDatePicker%20%28SDP%29&selectedStory=default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

GattoLupesco:
Fixes #1530

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -11,6 +11,7 @@ module.exports = (config) => {
     files: ['test/browser-main.js'],
 
     webpack: {
+      mode: 'development',
       externals: {
         sinon: true,
       },
@@ -36,7 +37,6 @@ module.exports = (config) => {
               presets: ['airbnb'],
             },
           },
-          { test: /\.json$/, loader: 'json-loader' },
 
           // Inject the Airbnb shims into the bundle
           { test: /test\/_helpers/, loader: 'imports-loader?shims=airbnb-js-shims' },
--- a/src/components/SingleDatePicker.jsx
+++ b/src/components/SingleDatePicker.jsx
@@ -249,7 +249,16 @@ class SingleDatePicker extends React.PureComponent {
 
   onFocusOut(e) {
     const { onFocusChange } = this.props;
-    if (this.dayPickerContainer.contains(e.relatedTarget || e.target) || this.dayPickerContainer.contains(e.target)) return;
+    // In cases where **relatedTarget** is not null, it points to the right
+    // element here. However, in cases where it is null (such as clicking on a
+    // specific day) or it is **document.body** (IE11), the appropriate value is **event.target**.
+    //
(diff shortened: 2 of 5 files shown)


src/components/SingleDatePicker.jsx
```diff
@@ -248,7 +248,7 @@ class SingleDatePicker extends React.PureComponent {
 
   onFocusOut(e) {
     const { onFocusChange } = this.props;
-    if (this.dayPickerContainer.contains(e.relatedTarget || e.target)) return;
+    if (this.dayPickerContainer.contains(e.relatedTarget || e.target) || this.dayPickerContainer.contains(e.target)) return;
```
ljharb: Can we add a test and/or story to cover this?

It seems like if `e.relatedTarget` is falsy, this line will cause `contains` to be invoked twice with the same `e.target`. Can that be avoided?

jrparish:
@GattoLupesco - do you need help with this at all? This is a critical fix and we'd love to have your changes as soon as possible.

ljharb:
@jrparish if you wanted to write tests, ping me here with a link to your branch and i can pull them into this PR.

jrparish:
@ljharb - I made some tweaks to match the logic to what is being done in the `DateRangePicker` and then added some tests - 
Branch: https://github.com/jrparish/react-dates/tree/fix_singledatepicker_focus_on_ie11
Commit: https://github.com/jrparish/react-dates/commit/b75d56c91a5731f2cfb972872b7681e597a827d7

karma.conf.js
```diff
@@ -36,7 +37,6 @@ module.exports = (config) => {
               presets: ['airbnb'],
             },
           },
-          { test: /\.json$/, loader: 'json-loader' },
```
ljharb: why is this removed?

ljharb:
@jrparish Although the karma tests aren't required, they still failed here - is that something you'd be willing to take a look at? (in a followup PR would be fine too)

ljharb:
actually i'm going to merge this as-is, since it leaves the karma tests in a much better situation.

karma.conf.js
jrparish: This was throwing an error when running karma and is not needed with webpack 4.

jrparish:
I can take a look at the other karma tests in a follow up 👍

PR #1589, opened 2019-03-20
Adding ariaDescribedBy prop

ryanmurakami:
We're using the `SingleDatePicker` inside a table and need to explicitly assign id's to the `aria-describedby` property of the `input` element. This PR allows a user to add the `ariaDescribedBy` attribute to any DatePicker.


src/components/DateInput.jsx
```diff
@@ -29,6 +29,7 @@ const propTypes = forbidExtraProps({
   placeholder: PropTypes.string, // also used as label
   displayValue: PropTypes.string,
   screenReaderMessage: PropTypes.string,
+  ariaLabelledBy: PropTypes.string,
```
backwardok: Suggestion: It might be worth creating a custom prop type here that supports a valid id type that's space separated. 

In a separate project, I created this regex for a single id: 
```const idRegex = /^_|^-|^\./;```

And you'd want to apply that for space separated values so that multiple ids can be passed.

Info on valid id values: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id

File: src/components/DateInput.jsx [hunk 2]
```diff
@@ -232,6 +235,7 @@ class DateInput extends React.PureComponent {
           readOnly={typeof readOnly === 'boolean' ? readOnly : isTouch}
           required={required}
           aria-describedby={screenReaderMessage && screenReaderMessageId}
+          aria-labelledby={ariaLabelledBy}
```
backwardok: You may also want to explicitly not set `aria-label` if `aria-labelledby` is passed in so that there isn't a potential conflict on which one takes precedence.

src/components/DateInput.jsx
ryanmurakami: Great idea!

src/components/DateRangePickerInput.jsx
```diff
@@ -33,6 +34,7 @@ const propTypes = forbidExtraProps({
   startDateId: PropTypes.string,
   startDatePlaceholderText: PropTypes.string,
   screenReaderMessage: PropTypes.string,
+  ariaLabelledBy: validIdList,
```
backwardok: This component has a range that includes two `DateInput` components, so you'll likely need two different props for each input (something like `startDateAriaLabelledBy`)

test/components/DateInput_spec.jsx
```diff
@@ -97,6 +97,34 @@ describe('DateInput', () => {
...
+
+      describe('props.ariaLabelledBy is truthy', () => {
+        beforeEach(() => {
+          wrapper = shallow((
+            <DateInput id={inputId} ariaLabelledBy={ariaLabelledBy} />
+          )).dive();
+        });
+
+        it('has aria-labelledby attribute that starts with ariaLabelledBy value', () => {
+          expect(wrapper.find(`input[aria-labelledby="${ariaLabelledBy}"]`)).to.have.lengthOf(1);
```
backwardok: might be good to also validate that the `aria-labelledby` attribute matches the passed in prop value

src/utils/validIdList.js
```diff
@@ -0,0 +1,21 @@
+const validIdListRegex = /(([A-Za-z]+[\w\-\:\.]*)\s*)*/;
```
backwardok: would be good to add tests for this util, with a variety of different test cases

test/components/DateInput_spec.jsx
ryanmurakami: That should be built into this test since the selector wouldn't match on anything that doesn't match the propType, correct?

backwardok: Oh! I missed that you actually had the value specifically passed in (it wrapped in the side by side view and I thought you were just querying for the presence of the attribute). This seems good, and perhaps then it would be good in the second test to check that `aria-labelledby` isn't set at all (and not just that it's not set with the specific value)

ryanmurakami: Yep, that's a better negative test. Will do.

ryanmurakami: So, I added a test that makes sure that `null` is set on `aria-labelledby` in the react props. When rendered, the attribute won't be there in that case. Do you think that's sufficient? Reading around, it seems that testing the action DOM rendering is an antipattern. Thoughts?

src/components/DateRangePickerInput.jsx
ryanmurakami: Great point. We're just using the `SingleDatePicker` so I missed this.

ryanmurakami: okay, added this in

src/utils/validIdList.js
```diff
@@ -0,0 +1,22 @@
+const validIdListRegex = /(([A-Za-z]+[\w\-:.]*)\s*)*/;
+
+// A more complex regex is used here so that
+// all IDs in an array must work. Even one incorrect
+// id in the list will cause this to return false.
+function isValidIdList(idList) {
+  if (!idList) return true;
+  const matches = idList.match(validIdListRegex);
+  return matches && matches.includes(idList);
```
ljharb: could this be `/^(([A-Za-z]+[\w\-:.]*)\s*)+$/.test(idList || '')`?

src/utils/validIdList.js
```diff
@@ -0,0 +1,22 @@
...
+}
+
+// validates a list of IDs for use in aria-labelledby attributes
+// valid ID format is listed here:
+// https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/id
+export default function (properties, propName, componentName) {
```
ljharb:
```suggestion
export default function validIDList(properties, propName, componentName) {
```

src/utils/validIdList.js
```diff
@@ -0,0 +1,22 @@
+const validIdListRegex = /(([A-Za-z]+[\w\-:.]*)\s*)*/;
```
ljharb: it'd be nice to capitalize all the initialisms (`ID` preferred over `Id`)

src/utils/validIdList.js
ryanmurakami: Yes, thank you thank you. I'm no regexpert so I couldn't get it to do that. The above regex would simplify it for sure.

src/utils/validIdList.js
ryanmurakami: Sounds good

test/components/DateInput_spec.jsx
```diff
@@ -97,6 +97,35 @@ describe('DateInput', () => {
...
+        });
+      });
+
+      describe('props.ariaLabelledBy is falsy', () => {
+        beforeEach(() => {
+          wrapper = shallow(<DateInput id={inputId} />).dive();
+        });
+
+        it('does not have aria-labelledby attribute value', () => {
+          expect(wrapper.find(`input[aria-labelledby="${ariaLabelledBy}"]`)).to.have.lengthOf(0);
```
backwardok: You can simplify these two lines by doing `wrapper.find('input[aria-labelledby]')` instead

test/components/DateInput_spec.jsx
backwardok: I don't personally have a strong preference one way or another, but I know some other people do :) I added a suggestion to simplify your existing test to use a similar pattern as to what exists in other tests in this file

test/components/DateInput_spec.jsx
ryanmurakami: When I try that, the `.find` does return one item bc the property is there in the React component, so I don't think this will work? Maybe I'm doing it wrong?

backwardok: Oh, I think it might be because you're passing `null` and not `undefined`. If I remember correctly, passing `null` will result in `null` getting passed as the value to that prop and passing `undefined` will not set the prop. This is accomplishing what you want so no need to change it!

src/components/DateInput.jsx
```diff
@@ -217,7 +221,7 @@ class DateInput extends React.PureComponent {
             focused && styles.DateInput_input__focused,
             disabled && styles.DateInput_input__disabled,
           )}
-          aria-label={placeholder}
+          aria-label={!ariaLabelledBy && placeholder}
```
ljharb: i think you may want a ternary here, to ensure it gets set to undefined and not false?

src/components/DateInput.jsx
```diff
@@ -56,6 +58,7 @@ const defaultProps = {
   placeholder: 'Select Date',
   displayValue: '',
   screenReaderMessage: '',
+  ariaLabelledBy: null,
```
ljharb: this may need to be undefined? (same with all the added defaults)

ryanmurakami: That would also make handling testing easier

src/components/DateInput.jsx
```diff
@@ -220,7 +224,11 @@ class DateInput extends React.PureComponent {
             focused && styles.DateInput_input__focused,
             disabled && styles.DateInput_input__disabled,
           )}
-          aria-label={ariaLabel === undefined ? placeholder : ariaLabel}
+          aria-label={
+            ariaLabel === undefined 
+              ? ariaLabelledBy ? undefined : placeholder
```
ljharb: Could we maybe do this in an intermediate variable? Nested ternaries are very hard to follow.

ryanmurakami: I separated one ternary into a variable and left the original simplified variable in place. What do you think?

src/components/DateInput.jsx
```diff
@@ -194,6 +198,9 @@ class DateInput extends React.PureComponent {
 
     const value = dateString || displayValue || '';
     const screenReaderMessageId = `DateInput__screen-reader-message-${id}`;
+    // if an aria-labelledby prop exists, then don't use
+    // the placeholder for the aria-label attribute
+    const placeholderArialLabel = ariaLabelledBy ? undefined : placeholder;
```
ryanmurakami:
```suggestion
    const placeholderAriaLabel = ariaLabelledBy ? undefined : placeholder;
```

src/components/DateInput.jsx
```diff
@@ -220,7 +227,7 @@ class DateInput extends React.PureComponent {
             focused && styles.DateInput_input__focused,
             disabled && styles.DateInput_input__disabled,
           )}
-          aria-label={ariaLabel === undefined ? placeholder : ariaLabel}
+          aria-label={ariaLabel === undefined ? placeholderArialLabel : ariaLabel}
```
ryanmurakami:
```suggestion
          aria-label={ariaLabel === undefined ? placeholderAriaLabel : ariaLabel}
```

ljharb:
LGTM but i'll let @backwardok sign off.

PR #1590, opened 2019-03-20
[a11y] Add hasExternalLabel prop to disable aria-label

ryanmurakami:
The `aria-label` attribute is only needed for input fields that aren't labelled. We are using a `label` element to add a label for our DatePicker, but the persistent `aria-label` attribute coming from the `placeholder` prop is causing the `label` element to be ignored. This PR allows the disabling of the `aria-label` attribute if desired.

Documentation on when to use `aria-label` can be found here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-label_attribute


src/components/DateInput.jsx
```diff
@@ -217,7 +220,7 @@ class DateInput extends React.PureComponent {
             focused && styles.DateInput_input__focused,
             disabled && styles.DateInput_input__disabled,
           )}
-          aria-label={placeholder}
+          aria-label={noAriaLabel ? '' : placeholder}
```
backwardok: This would result in the label for the node to be the empty string instead of not setting it. What you'd want to do instead is set the prop to be undefined or null if you don't want it to be set.

src/components/DateInput.jsx
```diff
@@ -180,6 +182,7 @@ class DateInput extends React.PureComponent {
       disabled,
       required,
       readOnly,
+      noAriaLabel,
```
backwardok: instead of `noAriaLabel`, I would recommend something like `hasExternalLabel`, which more clearly indicates that the reason why you wouldn't set the `aria-label` is because there's an external label that's being used. It's also clearer for people who don't understand what ARIA is or why they'd not want it.

src/components/DateInput.jsx
```diff
@@ -217,7 +220,7 @@ class DateInput extends React.PureComponent {
             focused && styles.DateInput_input__focused,
             disabled && styles.DateInput_input__disabled,
           )}
-          aria-label={placeholder}
```
backwardok: Note: This is potentially not an appropriate label, if the placeholder is actually placeholder text and not a label. Arguably, this is a result of the design itself encouraging that behavior, but it's not ideal for accessibility.

src/components/DateInput.jsx
ryanmurakami: I like that better, I'll change it.

src/components/DateInput.jsx
ryanmurakami: Yeah, I agree with that. Perhaps a better design, instead of this PR, is to just require users to pass in a value if they want an `aria-label` set to something, and not use the `placeholder` by default. Accessibility-wise, I think that would be more ideal. What do you think?

backwardok: Agreed - that way it's more explicit about what the label is. Worthy of a separate PR :)

ryanmurakami: Opened it here #1594. It's a way better design than this PR anyways.

ryanmurakami:
Yes, I think that's a better approach. Closing.

PR #1594, opened 2019-03-21
[a11y] Set aria-label explicitly

ryanmurakami:
This PR would supersede #1590.

Currently the `aria-label` of the date picker `input` field is being set with the value of the `placeholder`. This seems to be almost always the wrong choice, as an `aria-label` is totally different from a placeholder. This PR makes it so you can explicitly set the `aria-label` of the input fields and the `placeholder` is no longer used for that attribute.


src/components/DateInput.jsx
```diff
@@ -217,7 +220,7 @@ class DateInput extends React.PureComponent {
             focused && styles.DateInput_input__focused,
             disabled && styles.DateInput_input__disabled,
           )}
-          aria-label={placeholder}
+          aria-label={ariaLabel}
```
ljharb: we'd need to retain the `ariaLabel || placeholder` fallback here, otherwise this is a breaking change.

ryanmurakami: Ah, got it. Didn't think about the breaking nature of this. I'll want some way to not have any `aria-label` field, so what do you think about having a ternary like: `ariaLabel === undefined ? placeholder : null`

ljharb: I’d do `: undefined, but that looks fine to me!

ryanmurakami: Yep, idk why I love `null` so much. I'll change it to `undefined`.

ryanmurakami: Yeah, I goofed the example a bit. I used `ariaLabel` as the second case in the ternary, it shouldn't have been `null`.

ryanmurakami:
@backwardok @ljharb Thanks for the thorough review and good comments!

Issue #18, opened 2016-08-19
DateRangePicker/SingleDatePicker should close on Escape Keypress


majapw:
So I think this can be discussed, but it seems to make sense for the `DayPicker` to hide on escape press. However, `DayPicker` visibility is intrinsically tied to focus right now which is maybe not what we want.

So one option we could do is just to blur the input on escape. The better option, I think, would be to decouple visibility from focus, such that the focus state remains when you hit escape, but the `DayPicker` would close. This begs the question of how you would reopen it from this state... arrow down maybe? I am not sure right now.

ljharb:
I think that until we have a way to reopen it from "focused + closed", escape should always blur the focus, and that once we have that way, we'd want an option to control whether escape blurs the focus or not.

Separately, down arrow makes sense to me as a way to open the datepicker in that scenario.

fullhdpixel:
How can we close the SingleDateSelector programmatically?

phuochau:
Same question with @fullhdpixel

iambrandonn:
I [just submitted a PR](https://github.com/airbnb/react-dates/pull/1595) that closes the `DayPicker` when the input has focus and the escape key is pressed.

erin-doyle:
It looks like this PR has addressed this?  I'm going to close this but if I'm mistaken please feel free to reopen!

erin-doyle:
Actually I'm going to reopen this as it doesn't seem to be working.

PR #1595, opened 2019-03-25
[a11y] Closes calendar when escape is pressed

iambrandonn:
Fixes https://github.com/airbnb/react-dates/issues/18


src/components/DateRangePicker.jsx
```diff
@@ -281,6 +282,17 @@ class DateRangePicker extends React.PureComponent {
...
 
+  closeDayPicker() {
+    const { onFocusChange } = this.props;
+    this.setState({
+      isDateRangePickerInputFocused: false,
+      isDayPickerFocused: false,
+      showKeyboardShortcuts: false,
+    });
+
+    onFocusChange(null);
```
ljharb: should this maybe be in a callback to setState, so it's sure to happen after the state is set?

iambrandonn: Sure I can do that.

iambrandonn:
@ljharb Any other thoughts on this PR?

ljharb:
Seems good to me; I'm going to defer to @backwardok and @majapw here.

src/components/DateRangePicker.jsx
```diff
@@ -281,6 +282,17 @@ class DateRangePicker extends React.PureComponent {
     this.container = ref;
   }
 
+  closeDayPicker() {
+    const { onFocusChange } = this.props;
+    this.setState({
```
backwardok: If the keyboard shortcut screen was open, I'd expect hitting escape would close it. If the calendar was open, I'd expect hitting escape would close it. But, if the calendar AND the keyboard shortcut screen were open, I'd expecting hitting escape would only close one (the one on top, which is the keyboard shortcut screen in this case).

src/components/DateInput.jsx
```diff
@@ -159,6 +162,8 @@ class DateInput extends React.PureComponent {
     } else if (key === '?') {
       e.preventDefault();
       onKeyDownQuestionMark(e);
+    } else if (key === 'Escape') {
```
backwardok: Seems like you'd also want to add escape key handling within the calendar overlay itself as well.

iambrandonn:
@backwardok I've made a few more changes so the calendar will close consistently, whether the user gained focus by pressing the down arrow key or by pushing the tab key.  It seems to be working well for me.  I'm willing to set up a github pages branch so you can see it working... but for some reason `build-storybook` is giving errors like this (even on the master branch without my changes):
```
Property declarations[0] of VariableDeclaration expected node to be of a type ["VariableDeclarator"]
```

src/components/DateRangePicker.jsx
```diff
@@ -267,6 +269,15 @@ class DateRangePicker extends React.PureComponent {
     });
   }
 
+  onKeyDownTab(evt) {
+    const { focusedInput } = this.props;
+    if (focusedInput !== null) {
```
backwardok: There's another change (https://github.com/airbnb/react-dates/pull/1553) that's going to make it so that there's only one instance of the calendar, and it sits between the two inputs. This will end up breaking the behavior you're adding here, since it will cause a keyboard trap on the last input.

In general, it's not advised to manually handle focus from tabbing. What's the need for adding this behavior here?

PR #1598, opened 2019-03-26
Update blocked navigation (min/maxDate) stories

ianduvall:
This fixes two stories around blocked navigation with min and max dates. This change keeps today always between the min and max dates which makes the expected behavior of the story easier to understand.


ianduvall:
@ljharb @majapw thoughts?

lencioni:
Merged in 3b284c9

Issue #1606, opened 2019-04-02
MouseUp manual blur triggers close day picker in IE


DBjelovuk:
**react-dates version**
e.g. react-dates@18.3.1

**Describe the bug**
Manual blur triggers on MouseUp cause the day picker to close in IE.

**Is the issue reproducible in Storybook?**
http://airbnb.io/react-dates/?selectedKind=SingleDatePicker%20%28SDP%29&selectedStory=default&full=0&addons=1&stories=1&panelRight=0&addonPanel=storybook%2Factions%2Factions-panel

**Additional context**
Commenting out the `.blur()` lines in the following files prevents the behaviour:
```
src/components/DayPicker.jsx
src/components/DayPickerKeyboardShortcuts.jsx
src/components/DayPickerNavigation.jsx
```

PR #1607, opened 2019-04-05
Replace programmatic blurs with mousedown preventDefault

DBjelovuk:
Fixes https://github.com/airbnb/react-dates/issues/1606.


ljharb:
I don’t think “remove the blur” is the solution here. Why is the modal closing in IE when the blur happens?

DBjelovuk:
Can't say, but I'd contend the blur is a pretty poor solution to begin with. Don't see what makes this any worse?

backwardok:
I was testing out the bug in IE11 and Edge, and found that Edge seems to work fine but IE does not. What's also interesting is that in IE11, the handling done in the DateRangePicker works fine, but not in the SingleDatePicker. Might be good to investigate why the behavior is different between the two and follow the same behavior.

Separately, as for `blur()`, I agree that using `blur` for the sole purpose of getting rid of a focus ring is not a good practice. I played around with the approach of preventing default on mousedown (https://codepen.io/backwardok/pen/qwryXN) and it seems like it works as an alternative approach to blur, and it doesn't have the negative effect of resetting focus. Tested that codepen with VoiceOver for macOS + Safari, VoiceOver for iOS + Safari, Narrator + Edge, and NVDA + Firefox, and it all seems to work. I think this could be an interesting approach for the same result without the negative impact to keyboard and screen reader users that `blur` has.

PR #1609, opened 2019-04-06
Pass isVisible to renderMonthElement

gidztech:
`renderMonthElement` is called multiple times when you navigate between months. This is because the month before and the month after the currently visible month are still rendered to the DOM, but are hidden with CSS. This is presumably done for smoother animation/performance.

However, it would be very useful to know which month is actually visible when renders occur. This PR passes through the `isVisible` prop received in the `CalendarMonth` component to the `renderMonthElement` event.


gidztech:
I'm struggling to add a test for this. Firstly, I'm not sure what test is relevant here. There doesn't appear to be any tests of the `renderMonthElement` prop currently. I tried to write a test that would check the type of arguments received in the `renderMonthElement` handler after artificially invoking `onPrevMonthClick` , but I don't think it works without using Enzyme's `mount` with JSDom. I'm also not sure that that test actually provides much value here.

ljharb:
It seems like a shallow test that passes in a sinon spy as `renderMonthElement` would be able to assert on the arguments it receives, and also assert that the spy's return value gets rendered?

gidztech:
@ljharb I had already tried that but found that the `renderMonthElement` function wasn't being called. The call count was always 0 on the spy. I'm guessing it has to do with Enzyme's shallow rendering?

```js
const renderMonthElementSpy = sinon.spy();
const wrapper = shallow(<DayPicker renderMonthElement={renderMonthElementSpy} />).dive();
wrapper.instance().onPrevMonthClick();
expect(renderMonthElementSpy.callCount).to.equal(1);
```

ljharb:
It should be called by a shallow test on CalendarMonth directly

test/components/CalendarMonth_spec.jsx
```diff
@@ -1,7 +1,9 @@
 import React from 'react';
 import { expect } from 'chai';
 import { shallow } from 'enzyme';
+import sinon from 'sinon-sandbox';
 import moment from 'moment';
+import isFunction from 'lodash/isFunction';
```
ljharb: There’s no need to use lodash here; let’s compare to typeof directly

src/components/CalendarMonth.jsx
```diff
@@ -197,7 +197,9 @@ class CalendarMonth extends React.PureComponent {
           )}
         >
           {renderMonthElement ? (
-            renderMonthElement({ month, onMonthSelect, onYearSelect })
+            renderMonthElement({
+              month, onMonthSelect, onYearSelect, isVisible,
```
ljharb: If the entire object literal cannot fit on one line, each object property here must be on its own line 
```suggestion
              month,
              onMonthSelect,
              onYearSelect,
              isVisible,
```

test/components/CalendarMonth_spec.jsx
```diff
@@ -26,5 +28,16 @@ describe('CalendarMonth', () => {
         expect(captionWrapper.text()).to.equal(MONTH.format('MMMM YYYY'));
       });
     });
+
+    it('renderMonthElement callback receives correct props', () => {
+      const renderMonthElementSpy = sinon.spy();
+      const wrapper = shallow(<CalendarMonth renderMonthElement={renderMonthElementSpy} />).dive();
+      wrapper.setProps({ month: moment().subtract(1, 'months') });
+      const arg = renderMonthElementSpy.getCall(0).args[0];
```
ljharb:
```suggestion
      const [{ month, onMonthSelect, onYearSelect, isVisible }] = renderMonthElementSpy.getCall(0).args;
```

gidztech:
@ljharb Should be good for review now.

gidztech:
@ljharb Thanks for merging. Is there any chance of a minor release of this. I'd like to take advantage of the change to fix a problem in an app I'm working on.

PR #1611, opened 2019-04-08
test: transform modules in karma so sinon can stub properly

jrparish:
This fixes all the failing tests when running using Karma. 

https://github.com/sinonjs/sinon/issues/1711

### DIFF
--- a/karma.conf.js
+++ b/karma.conf.js
@@ -33,8 +33,13 @@ module.exports = (config) => {
               path.join(__dirname, 'test'),
               require.resolve('airbnb-js-shims'),
             ],
-            query: {
-              presets: ['airbnb'],
+            options: {
+              presets: [
+                // setting modules to false so it does not transform twice
+                ['airbnb', { modules: false }],
+                // transform to cjs so sinon can stub properly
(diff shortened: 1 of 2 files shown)


karma.conf.js
```diff
@@ -33,8 +33,12 @@ module.exports = (config) => {
               path.join(__dirname, 'test'),
               require.resolve('airbnb-js-shims'),
             ],
-            query: {
-              presets: ['airbnb'],
+            options: {
+              presets: [
+                'airbnb',
+                // transform to cjs so sinon can stub properly
```
ljharb: interesting, does the airbnb preset not already do this? i think it accepts a modules option too

jrparish: I originally tried to set it on the airbnb preset, but it prevents you from specifying anything other than true/false/auto (https://github.com/airbnb/babel-preset-airbnb/blob/master/index.js#L35)

ljharb: K, I’ll have to expand that to accept cjs; in the meantime, let’s set this one to false so it doesn’t include the env preset twice?

jrparish: Sure thing 👍

Issue #1623, opened 2019-04-17
[with Resolution] Using withPortal on DateRangePicker, when select a Date doesn't trigger onChange event


phibsholo:
When I use the prop `withPortal` on the component `<DateRangePicker />` and I focus on a startDate, it opens the calendar properly, but after selecting the date, it doesn't trigger the change event, and the calendar closes leaving the input empty.

I used this component some time ago and it was working properly, but after updating dependencies, like `material-ui` and `react`, it started falling.

phibsholo:
I tried to editing the library and this can be fixed doing this:

- Go to DayPickerRangeController and into the function onDayClick edit line 526 adding the following code:
```
if (e) {
      e.stopPropagation();
      e.preventDefault();
 }
```

vijilvpillai:
Hi @phibsholo ,
Could you please share me the fix.I am also facing the issue with bootstrap modal

phibsholo:
HI @vijilvpillai, 
the fix is just explained on the comment before [,](https://github.com/airbnb/react-dates/issues/1623#issuecomment-483999157) in any case, I could solve the problem using Material UI modal and inside use the DayPickerRangeController.
I had to do:
`
import Dialog from '@material-ui/core/Dialog';
<Dialog>
    <DayPickerRangeController
          {...props}
        />
</Dialog>
`
This works perfectly... but in any case, the fix that I set on the previous comment is working too ;)

vijilvpillai:
Thank you @phibsholo . I am using a bootstrap modal and SingleDatePicker. My datepicker is already inside  a bootstrap  parent modal only. As a temporary fix I am not using portal option . As an alternative I reduced the calendar day cell size so that it can fit the screen

PR #1624, opened 2019-04-17
Add stopPropagation on onDayClick method

phibsholo:
This commit resolves this issue: https://github.com/airbnb/react-dates/issues/1623

### DIFF
--- a/src/components/DayPickerRangeController.jsx
+++ b/src/components/DayPickerRangeController.jsx
@@ -523,7 +523,10 @@ export default class DayPickerRangeController extends React.PureComponent {
       disabled,
     } = this.props;
 
-    if (e) e.preventDefault();
+    if (e) {
+      e.preventDefault();
+      e.stopPropagation();
+    }
     if (this.isBlocked(day)) return;
 
     let { startDate, endDate } = this.props;


src/components/DayPickerRangeController.jsx
```diff
@@ -523,7 +523,10 @@ export default class DayPickerRangeController extends React.PureComponent {
       disabled,
     } = this.props;
 
-    if (e) e.preventDefault();
+    if (e) {
+      e.preventDefault();
+      e.stopPropagation();
```
ljharb: i'm pretty uncomfortable with ever stopping event propagation.

can you add a regression test?

phibsholo: Eyyy after update the rest of dependencies of my project it works properly!!
In any case, thanks a lot for the time spent on that.
I think it's a tricky case ;)

phibsholo: I was thinking that it is not your fault... it’s caused by material ui library https://material-ui.com/.

I’m going to explain to myself.

Currently I’m using react-dates into a modal from material ui, and after update from version 1 to 3, some events it wasn’t controlled from them side, but it’s good to you, to know that this error could will be happen into a modal.

This error only happen when I set withPortal option or withFullScreenPortal.

phibsholo: After all, I had to do a tricky fix, because using DateRangePicker with prop withModal it's impossible capture the events of the calendar.
For this reason I had to edit the styles like a modal and capture the event on click background to close the calendar and clean the input values.
It's to tricky... but it's the solution that works to me... If you it will apply this fix, tell me to restore and use DateRangePicker with the prop withModal. 
Thanks!

ljharb:
This still needs a regression test (and now a rebase)

PR #1634, opened 2019-05-03
Fix Detached Container Y Position

yusufff:


mattaningram:
This would make my life dealing with react-dates a lot easier.

ljharb (changes requested):
Can you please add some test cases that show what bug this is fixing?

PR #1644, opened 2019-05-15
[New] `SingleDatePickerInputController`: Allow passing multiple date formats to make validation more permissive

### LINKED ISSUE #345: Validate input using multiple formats

Floriferous:
Closes #345.

This is a proposal at the moment, I didn't yet look at all the other places this needs attention for. So I'd like to get some feedback before going in deeper.

The basic idea is this:

```jsx
<SingleDatePicker
  dateFormat={["D MMMM YYYY", "D/M/YYYY", "D.M.YYYY"]}
/>
```

The first format in the array will be used to display dates, and all the other ones will be used to validate it while typing, to improve the user experience, as fewer inputs will be rejected (there is also no error shown, leaving users stranded).

The main issue with this pattern, is that you can provide conflicting date formats, such as `DD/MM/YYYY` and `MM/DD/YYYY`, in which case it's unclear what will happen. I could try to write a helper to throw when this happens, but documentation might be enough.

However I believe the benefit outweighs the potential misuse, as just my example above with a couple more formats will help a lot!


daviddelusenet:
Would love to have this feature or something similar but I noticed that the Airbnb team barely responds to any issue or pull request.

Pretty disappointing the put it nicely..

Lukasz-Trzaskowski:
+1 for this. Could this be integrated?

src/components/SingleDatePickerInputController.jsx
```diff
@@ -175,7 +175,24 @@ export default class SingleDatePickerInputController extends React.PureComponent
...
     const { displayFormat } = this.props;
-    return typeof displayFormat === 'string' ? displayFormat : displayFormat();
+    
+    if (typeof displayFormat === 'string') {
+      return displayFormat;
+    } else if (Array.isArray(displayFormat)) {
+      return displayFormat[0];
+    }
+    
+    return displayFormat();
```
ljharb, line 185:
```suggestion
    return typeof displayFormat === 'function' ? displayFormat() : [].concat(displayFormat)[0];
```

src/components/SingleDatePickerInputController.jsx
```diff
@@ -175,7 +175,24 @@ export default class SingleDatePickerInputController extends React.PureComponent
...
+  }
+  
+  getValidationDisplayFormats() {
+    const { displayFormat } = this.props;
+
+    if (typeof displayFormat === 'function') {
+      return displayFormat();
+    }
+    
+    return displayFormat;
```
ljharb, line 195:
```suggestion
    return [].concat(displayFormat);
```
this way it's always an array

src/utils/toMomentObject.js
```diff
@@ -3,10 +3,16 @@ import moment from 'moment';
...
-
+  let dateFormats = [];
+  
+  if (Array.isArray(customFormat)) {
+    dateFormats = [...customFormat, DISPLAY_FORMAT, ISO_FORMAT]
+  } else if (customFormat) {
+    dateFormats = [customFormat, DISPLAY_FORMAT, ISO_FORMAT]
+  } else {
+    dateFormats = [DISPLAY_FORMAT, ISO_FORMAT]
+  }
```
ljharb, line 14:
```suggestion
  const dateFormats = [].concat(
    customFormat || [],
    DISPLAY_FORMAT,
    ISO_FORMAT,
  ]);
```

ljharb (changes requested):
Thanks! We'll need a documentation update, and lots of tests to cover this addition.

PR #1654, opened 2019-05-29
Avoid some unnecessary re-rendering of CalendarDays

lencioni:
I have been investigating performance and noticed that when selecting a
date, all of the days for all of the months end up re-rendering. I
discovered that there are two causes for this currently:

1. All of the phrases are passed down to the CalendarDay components, and
   the phrases object can be different when selecting dates because of
   code like this, which toggles one translation back and forth:

   https://github.com/airbnb/react-dates/blob/888f6505/src/components/DayPickerRangeController.jsx#L501-L508

2. Similarly, all of the modifiers end up being different by reference
   because we end up creating new Sets for them every time, even if the
   sets are deeply equal.

To address the first issue, I decided to simplify the API of CalendarDay
a little, so that it takes an ariaLabel prop which is a simple string,
instead of a phrases prop. This allows us to prevent the changing of the
phrases object from automatically deoptimizing our leaf components.

And, to address the second issue, I introduced a simple cache for
modifier Sets. This allows the downstream components to perform a simple
referential equality check on them to avoid re-rendering, which we get
for free with PureComponent.

These two optimizations prevent many dates from re-rendering, namely
dates that are blocked and out of range. In storybook on my Macbook pro
(non-production React) this seems to bring down the time it takes to
update DayPickerRangeController when selecting dates from ~150ms to
~110ms. I think this will have a positive effect on some other
operations or other types of datepickers, but I didn't profile them.

It is worth noting that there is still a lot of re-rendering happening,
but that is because the value of ariaLabel ends up changing while you
select dates from:

> Choose Thursday, May 30, 2019 as your check-in date. It’s available.

to:

> Choose Thursday, May 30, 2019 as your check-out date. It’s available.

This seems important for accessibility, and I don't have any good ideas
at this moment of how to avoid this yet.

Before:
image

After:
image

It looks like there might be a good opportunity to optimize DayPickerRangeController#componentWillReceiveProps--this is where the modifiers are all built on every update.

---
Convert CalendarDay's aria-label to aria-labelledby

I've been working to improve performance of react-dates, and I noticed
that on many interactions every single date ends up being re-rendered.
In my previous commit, I partly fixed this by avoiding passing the
always-new phrases object down to CalendarDay, and to memoize modifiers
so they also don't deoptimize the pure components at the edges.

However, this still left one phrase that always changes when
selecting start and end dates which still caused most dates to be
re-rendered.

To avoid this, instead of passing an aria-label to CalendarDay, which
needs to change when switching from selecting a start date to selecting
an end date, we can render a div that contains the phrase for the
currently focused date, and pass a reference to it to be used by
aria-labelledby.

The downside here is that since ids need to be unique per page and there
is no reliable way to auto-generate an id on the server that will be
consistent on the client, I had to add required id props in a number of
places. For folks using components that already require startDateId, we
can generate a new id based on that value, so the friction there should
be minimal.

Testing this using VoiceOver on Safari and the results seem to be just
what we want.

Profiling this in storybook (non-production React) on my Macbook pro
brings the time it takes to update DayPickerRangeController from ~110ms
after my previous commit to ~75ms. (Before both of these commits, we
were looking at ~150ms).

I think there might still be some opportunity in preventing unnecessary
re-renders of CalendarMonth, but most of the remaining opportunity looks
like it might be in optimizing how modifiers are built in places like
DayPickerRangeController#componentWillReceiveProps.

After:
image

### DIFF
--- a/src/components/CalendarDay.jsx
+++ b/src/components/CalendarDay.jsx
@@ -5,8 +5,6 @@ import { forbidExtraProps, nonNegativeInteger } from 'airbnb-prop-types';
 import { css, withStyles, withStylesPropTypes } from 'react-with-styles';
 import moment from 'moment';
 
-import { CalendarDayPhrases } from '../defaultPhrases';
-import getPhrasePropTypes from '../utils/getPhrasePropTypes';
 import getCalendarDaySettings from '../utils/getCalendarDaySettings';
 import ModifiersShape from '../shapes/ModifiersShape';
 
@@ -24,10 +22,7 @@ const propTypes = forbidExtraProps({
   onDayMouseEnter: PropTypes.func,
   onDayMouseLeave: PropTypes.func,
   renderDayContents: PropTypes.func,
-  ariaLabelFormat: PropTypes.string,
-
-  // internationalization
-  phrases: PropTypes.shape(getPhrasePropTypes(CalendarDayPhrases)),
+  ariaLabel: PropTypes.string.isRequired,
 });
 
 const defaultProps = {
@@ -41,10 +36,6 @@ const defaultProps = {
   onDayMouseEnter() {},
   onDayMouseLeave() {},
   renderDayContents: null,
-  ariaLabelFormat: 'dddd, LL',
-
-  // internationalization
-  phrases: CalendarDayPhrases,
 };
 
 class CalendarDay extends React.PureComponent {
@@ -94,14 +85,13 @@ class CalendarDay extends React.PureComponent {
   render() {
     const {
       day,
-      ariaLabelFormat,
       daySize,
       isOutsideDay,
       modifiers,
       renderDayContents,
       tabIndex,
       styles,
-      phrases,
+      ariaLabel,
     } = this.props;
 
     if (!day) return <td />;
@@ -112,8 +102,7 @@ class CalendarDay extends React.PureComponent {
       selected,
       hoveredSpan,
       isOutsideRange,
-      ariaLabel,
-    } = getCalendarDaySettings(day, ariaLabelFormat, daySize, modifiers, phrases);
+    } = getCalendarDaySettings(daySize, modifiers);
 
     return (
       <td
--- a/src/components/CalendarDay.jsx
+++ b/src/components/CalendarDay.jsx
@@ -22,7 +22,7 @@ const propTypes = forbidExtraProps({
   onDayMouseEnter: PropTypes.func,
   onDayMouseLeave: PropTypes.func,
   renderDayContents: PropTypes.func,
-  ariaLabel: PropTypes.string.isRequired,
+  ariaLabelledby: PropTypes.string.isRequired,
 });
 
 const defaultProps = {
@@ -91,7 +91,7 @@ class CalendarDay extends React.PureComponent {
       renderDayContents,
       tabIndex,
       styles,
-      ariaLabel,
+      ariaLabelledby,
     } = this.props;
 
     if (!day) return <td />;
@@ -131,7 +131,7 @@ class CalendarDay extends React.PureComponent {
         role="button" // eslint-disable-line jsx-a11y/no-noninteractive-element-to-interactive-role
         ref={this.setButtonRef}
         aria-disabled={modifiers.has('blocked')}
-        aria-label={ariaLabel}
+        aria-labelledby={ariaLabelledby}
         onMouseEnter={(e) => { this.onDayMouseEnter(day, e); }}
         onMouseLeave={(e) => { this.onDayMouseLeave(day, e); }}
         onMouseUp={(e) => { e.currentTarget.blur(); }}
--- a/src/components/CalendarMonth.jsx
+++ b/src/components/CalendarMonth.jsx
@@ -7,9 +7,6 @@ import { forbidExtraProps, mutuallyExclusiveProps, nonNegativeInteger } from 'ai
 import { css, withStyles, withStylesPropTypes } from 'react-with-styles';
 import moment from 'moment';
 
-import { CalendarDayPhrases } from '../defaultPhrases';
-import getPhrasePropTypes from '../utils/getPhrasePropTypes';
-
 import CalendarWeek from './CalendarWeek';
 import CalendarDay from './CalendarDay';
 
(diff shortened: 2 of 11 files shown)


lencioni:
If this approach looks good, I'll update the tests.

src/components/CalendarDay.jsx
```diff
@@ -94,14 +85,13 @@ class CalendarDay extends React.PureComponent {
...
       day,
-      ariaLabelFormat,
       daySize,
       isOutsideDay,
       modifiers,
       renderDayContents,
       tabIndex,
       styles,
-      phrases,
+      ariaLabel,
```
lencioni: I'm not really sure what is considered part of the public API here and if this is a breaking change or not.

nkinser (commented):
Overall looks good to me. You will need to make the same changes to `CustomizableCalendarDay` that you made in `CalendarDay`, and please add tests since the props have changed

src/components/CalendarDay.jsx
nkinser: This is a breaking change. This would definitely break the pdp calendar

ljharb: Removal of any prop is generally a breaking change.

ljharb: Could we keep both, and ignore the old one when the new one is present?

src/components/CalendarMonth.jsx
```diff
@@ -86,7 +127,7 @@ const defaultProps = {
   // i18n
   monthFormat: 'MMMM YYYY', // english locale
   phrases: CalendarDayPhrases,
-  dayAriaLabelFormat: undefined,
+  dayAriaLabelFormat: 'dddd, LL',
```
ljharb: this could also be a breaking change if people are relying on the default.

src/components/DayPickerRangeController.jsx
```diff
@@ -1013,7 +1025,7 @@ export default class DayPickerRangeController extends React.PureComponent {
           ...days,
           [monthIso]: {
             ...month,
-            [iso]: modifiers,
+            [iso]: this.getCachedModifiers(Array.from(modifiers)),
```
ljharb: why is Array.from needed here?

lencioni: modifiers is a Set here, and this.getCachedModifiers expects an array, not a Set, because it calls .join on it to generate the cache key.

src/components/CalendarDay.jsx
lencioni: Maybe. Alternatively, I think I might try leaving the phrases prop here and instead pass down `focusedInput` so we don't have to change `phrases` every time. I want to think through how we might get this optimization working for more days without breaking accessibility before we choose a direction to go in here.

erin-doyle:
I just want to put a plug in for also making the relevant changes to `DayPickerSingleDateController` that you've made to `DayPickerRangeController`.

src/components/CalendarDay.jsx
lencioni: I think in order to optimize most dates, we will need to avoid changing `ariaLabel` when `focusedInput` changes. I think in order to do that, we can render at a higher level (`CalendarMonthGrid` maybe) a hidden node that manages the phrase for the focused date, and pass down an id to CalendarDay for `aria-labelledby`. Still need to think through `CustomizableCalendarDay` I think.

lencioni: I've added a commit to this PR that takes the `aria-labelledby` approach.

src/components/CalendarMonthGrid.jsx
```diff
@@ -303,6 +353,15 @@ class CalendarMonthGrid extends React.PureComponent {
         ref={this.setContainerRef}
         onTransitionEnd={onMonthTransitionEnd}
       >
+        <div id={dayLabelledbyId} hidden>
```
majapw: is the `hidden` attribute here sufficient?

src/components/CalendarMonthGrid.jsx
```diff
@@ -303,6 +353,15 @@ class CalendarMonthGrid extends React.PureComponent {
...
+          {!!focusedDate && getAriaLabel(
```
majapw: If this is empty, will it have any effect?

src/components/DayPickerRangeController.jsx
```diff
@@ -944,8 +946,19 @@ export default class DayPickerRangeController extends React.PureComponent {
     return modifiers;
   }
 
+  getCachedModifiers(modifiersArray) {
+    const cacheKey = modifiersArray.join('!');
```
majapw: 🎉 This is great

src/components/CalendarMonthGrid.jsx
backwardok: I don't think that this approach will work - by default in VO on iOS, when you navigate to different elements on the page, it doesn't move keyboard focus. This would mean that this text will not change as you move through the different days and keep saying that the same day.

majapw (commented):
This looks really good! :) Thank you for doing this work @lencioni!

src/components/CalendarMonthGrid.jsx
backwardok: If you're willing to create new phrases, you can do a mixture of the approach you have here and what I had suggested with using two references within `aria-labelledby`. That way this container updates based on the qualification (checkout/checkin), and the id references don't change, but the date strings are tied to the date itself and not this node.

lencioni: rats! It sounds like we might need to break this into two parts then, with one part of the phrase here and the other part of the phrase in the CalendarDay?

I'm not sure we can make that big of a breaking change here. @nkinser what do you think?

lencioni: For reference, here's the pseudo-code you shared with me elsewhere that illustrates what we are talking about here:
```
<div id="endDatePhrase" hidden>Select as your end date</div>
...
<button aria-labelledby="05292019 endDatePhrase">
  <div id="05292019" hidden>Wednesday, May 29, 2019</div>
  29
</button>
```

lencioni: Hmm, this is tricky. Phrases can be functions that take arguments like the day, and I think some people are customizing these voiceover phrases to provide more specific information about individual days than what we currently support out of the box. I think I'll need to work through this with @nkinser to make sure we have all of our use cases properly covered.

nkinser: Sorry for the delayed response. 
I think this would make it impossible to customize the phrase based on the state of the calendar, which is really important for the pdp calendar since there's so much information that we display visually that we need to also convey to VO users. I think this is a really complicated issue so it's probably a good idea to talk about this in person

PR #1656, opened 2019-05-30
Optimize toISODateString and toISOMonthString

lencioni:
These functions get called repeatedly when computing modifiers, which
happens any time components like DayPickerRangeController receives
props.

By manually building this string instead of relying on moment's format
function, we can get better performance. In my profiling, this cuts the
time spent in DayPickerRangeController#componentWillReceiveProps from
~50ms to ~40ms.

Before:
image

After:
image

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -1,6 +1,6 @@
 export const DISPLAY_FORMAT = 'L';
 export const ISO_FORMAT = 'YYYY-MM-DD';
-export const ISO_MONTH_FORMAT = 'YYYY-MM';
+export const ISO_MONTH_FORMAT = 'YYYY-MM'; // TODO delete this line of dead code on next breaking change
 
 export const START_DATE = 'startDate';
 export const END_DATE = 'endDate';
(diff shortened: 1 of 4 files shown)


lencioni:
@ljharb I added a test for toISODateString but not toISOMonthString because the existing tests there were all written in this way.

test/utils/toISOMonthString_spec.js
```diff
@@ -1,9 +1,11 @@
 import moment from 'moment';
 import { expect } from 'chai';
 
-import { ISO_FORMAT, ISO_MONTH_FORMAT } from '../../src/constants';
```
ljharb: why inline this, if the constant is still going to exist?

lencioni: It is only used in one test file now, and in no production code, so I don't think it makes sense to ship this code to clients.

ljharb: sure but this PR doesn’t delete the constant (which would be breaking change). Can the inlining wait til then?

lencioni: Oops, I forgot to save that file. Yeah, I can keep it for now. I'll add a note to delete it as part of the next breaking change.

PR #1659, opened 2019-05-30
Optimize deleteModifier/addModifier

lencioni:
While profiling DayRangePickerController, I noticed that we spend a
significant amount of time generating modifiers whenever anything is
changed. A good amount of this cost comes from these reducers that
unnecessarily create new objects and spread into them on every
iteration. We can make this faster by mutating the accumulator while
inside the reducer.

In my profiling, this reduces the speed of deleteModifier from ~28ms to
~20ms, and I think it will reduce the memory usage, which should further
help reduce garbage collection times.

---

Mutate updatedDaysAfterDeletion in deleteModifier/addModifier

We create this object in this function and then return it at the end, so
it should be safe to just mutate it while we are in this function
instead of rebuilding it immediately again.

This reduces the amount of time spent in deleteModifier from ~20ms to
~16ms.

---

Avoid creating new Sets when deleting/adding modifiers

I noticed that we always create a new set and then delete from it even
if the modifier is not in the set, or create a new set and then add it
even it the modifier is already in the set. This optimization reduces
the cost of DayPickerRangeController#componentWillReceiveProps from
~16ms to ~12ms.

Before:
image

After:
image

### DIFF
--- a/test/components/DayPickerRangeController_spec.jsx
+++ b/test/components/DayPickerRangeController_spec.jsx
@@ -891,9 +891,9 @@ describe('DayPickerRangeController', () => {
             const startOfMonth = today.clone().startOf('month');
             visibleDays = {
               [toISOMonthString(startOfMonth)]: {
-                [toISODateString(startOfMonth)]: [],
-                [toISODateString(startOfMonth.clone().add(1, 'day'))]: [],
-                [toISODateString(startOfMonth.clone().add(2, 'days'))]: [],
+                [toISODateString(startOfMonth)]: new Set(),
+                [toISODateString(startOfMonth.clone().add(1, 'day'))]: new Set(),
+                [toISODateString(startOfMonth.clone().add(2, 'days'))]: new Set(),
               },
             };
           });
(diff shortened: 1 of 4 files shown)


lencioni:
Alright friends, tests are passing and I think this is ready for review!

test/components/DayPickerRangeController_spec.jsx
```diff
@@ -974,9 +974,9 @@ describe('DayPickerRangeController', () => {
             const startOfMonth = today.clone().startOf('month');
             visibleDays = {
               [toISOMonthString(startOfMonth)]: {
-                [toISODateString(startOfMonth)]: [],
-                [toISODateString(startOfMonth.clone().add(1, 'day'))]: [],
-                [toISODateString(startOfMonth.clone().add(2, 'days'))]: [],
+                [toISODateString(startOfMonth)]: new Set(),
```
ljharb: this type change suggests an observable api change - is that the case, or am i missing context?

lencioni: I could be wrong but I think this is just a care of these fixtures being incorrect. Previously they were being converted to sets in the code by happenstance, but I think all of there are sets throughout the actual code.

lencioni: On phone now but I think these might be able to be passed in as props, so I might need to add a conversation step somewhere like in the constructor. I'll dig in more tomorrow.

lencioni: Nope, as far as I can tell, this is all component state. These tests are injecting it into the component by calling setState directly.

This is in the constructor where this variable begins its life: https://github.com/airbnb/react-dates/blob/0e2ed9a322dd0d8f9782614b40a362413406a76b/src/components/DayPickerRangeController.jsx#L217

And getStateForNewMonth uses getModifiers for visibleDates, which always uses Set for the set of modifiers: https://github.com/airbnb/react-dates/blob/0e2ed9a322dd0d8f9782614b40a362413406a76b/src/components/DayPickerRangeController.jsx#L965-L971

I'm pretty sure this is all internal stuff.

lencioni:
Yeah, it seems that there's a fair amount of duplicated code here, so abstracting something might make sense.

PR #1661, opened 2019-05-30
Add caching to isDayVisible

lencioni:
After the optimizations I applied to generating modifiers in
https://github.com/airbnb/react-dates/pull/1659, I noticed that
about 50% of the time in componentWillReceiveProps is spent in
isDayVisible() doing moment operations to generate the boundaries for
the check. Since these boundaries remain very stable across each call,
we can reduce most of the work done here to very little by adding some
caching of these values.

This brings down the time spent in componentWillReceiveProps from
~46ms to ~22ms.

I'm not doing anything to limit the size of these Maps, so this will
leak some memory, but I think the size of the cache will be really small
in most cases, so I'm not sure it matters.

Before:
image

After: 
image

### DIFF
--- a/src/utils/isDayVisible.js
+++ b/src/utils/isDayVisible.js
@@ -2,26 +2,55 @@ import moment from 'moment';
 
 import isBeforeDay from './isBeforeDay';
 import isAfterDay from './isAfterDay';
+import toISOMonthString from './toISOMonthString';
+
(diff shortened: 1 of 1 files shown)


lencioni:
@nkinser I think it is safe because I have two separate sets of caches and two branches of logic for that option. So the keys are the same, but they are pulling from different caches.

PR #1663, opened 2019-05-31
Cache some moment instances when rebuilding modifiers

lencioni:
When modifiers need to be rebuilt, which happens pretty frequently, we
iterate over every day and create a new moment instance for these
operations. These moment instances are never mutated, so we can safely
store and reuse them to improve performance.

This optimization reduces the time spent in
DayPickerRangeController#componentWillReceiveProps when selecting dates
from ~16ms to ~6-10ms.

Before:
image

After:
image

And here it is with a production build with 4x CPU slowdown.

Before:
image

After:
image

### DIFF
--- a/src/utils/getPooledMoment.js
+++ b/src/utils/getPooledMoment.js
@@ -1,0 +1,10 @@
+import moment from 'moment';
+
+const momentPool = new Map();
+export default function getPooledMoment(dayString) {
+  if (!momentPool.has(dayString)) {
+    momentPool.set(dayString, moment(dayString));
+  }
+
+  return momentPool.get(dayString);
+}
--- a/test/utils/getPooledMoment_spec.js
+++ b/test/utils/getPooledMoment_spec.js
@@ -1,0 +1,24 @@
+import { expect } from 'chai';
+import moment from 'moment';
+
+import getPooledMoment from '../../src/utils/getPooledMoment';
+
+describe('getPooledMoment', () => {
+  it('returns a moment given a day string', () => {
+    const momentObj = getPooledMoment('2017-12-10');
+    expect(moment.isMoment(momentObj)).to.equal(true);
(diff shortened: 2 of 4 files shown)


src/utils/getPooledMoment.js
```diff
@@ -0,0 +1,10 @@
+import moment from 'moment';
+
+const momentPool = new Map();
+export default function getPooledMoment(dayString) {
```
ljharb: unit tests?

lencioni: I added some tests

lencioni:
@ljharb any idea why coverage dropped?

ljharb:
@lencioni looks like the coverage run failed: https://travis-ci.org/airbnb/react-dates/jobs/539840359

```sh
Deprecation warning: value provided is not in a recognized RFC2822 or ISO format. moment construction falls back to js Date(), which is not reliable across all browsers and versions. Non RFC2822/ISO date formats are discouraged and will be removed in an upcoming major release. Please refer to http://momentjs.com/guides/#/warnings/js-date/ for more info.
Arguments: 
[0] _isAMomentObject: true, _isUTC: [secure], _useUTC: [secure], _l: undefined, _i: 10/10/2017, _f: undefined, _strict: undefined, _locale: [object Object]
```

lencioni:
I think that always happens when these tests are run: https://travis-ci.org/airbnb/react-dates/jobs/539840355#L365

test/utils/getPooledMoment_spec.js
```diff
@@ -0,0 +1,24 @@
+import { expect } from 'chai';
+import moment from 'moment';
+
+import getPooledMoment from '../../src/utils/getPooledMoment';
+
+describe.only('getPooledMoment', () => {
```
ljharb:
```suggestion
describe('getPooledMoment', () => {
```

lencioni: wherps

PR #1665, opened 2019-06-03
npmignore .cache, .github, .lgtm, and _gh-pages

lencioni:
I don't think there is any value in distributing these files in the
package.


lencioni:
Merged in 9862d28

PR #1667, opened 2019-06-04
Add guard for undefined objects in addModifier

lencioni:
After updating to v20.2.1, we started seeing a handful of errors about
`undefined is not an object` coming from these addModifier calls. It
isn't entirely clear to me how this happens, but I think we can prevent
it from being a problem by adding some guards in these codepaths.


lencioni:
@ljharb Totally agreed about the test, but I wasn't sure how this could happen so I wasn't sure how to write the test. Reading over the code again though, I think I see what I got confused about and I'll try to get a test in (I think I can also simplify my change).

lencioni:
@ljharb PR updated, PTAL

PR #1678, opened 2019-06-08
fixed: wrong props to SDP onClose(#1674)

### LINKED ISSUE #1674: SingleDatePicker: startDate and endDate passed to onClose when clicking outside
react-dates@20.2.0

While implementing the `SingleDatePicker` I found out that `onDayClick` passes `{ date: Moment }` and `onOutsideClick` passes `{ startDate: undefined, endDate: undefined }` to `onClose`. See: https://github.com/airbnb/react-dates/blob/11de2016bf6e7111ddebd5a7f7fc1fe929a3600c/src/components/SingleDatePicker.jsx#L206

Should the latter even be the case for `SingleDatePicker`, since there's only one date?

AlokTakshak:
Fixes #1674.

### DIFF
--- a/src/components/SingleDatePicker.jsx
+++ b/src/components/SingleDatePicker.jsx
@@ -202,7 +202,7 @@ class SingleDatePicker extends React.PureComponent {
     });
 
     onFocusChange({ focused: false });
-    onClose({ date});
+    onClose({ date });
   }
 
   onInputFocus({ focused }) {
--- a/src/components/SingleDatePicker.jsx
+++ b/src/components/SingleDatePicker.jsx
@@ -188,8 +188,7 @@ class SingleDatePicker extends React.PureComponent {
       focused,
       onFocusChange,
       onClose,
-      startDate,
-      endDate,
+      date,
       appendToBody,
     } = this.props;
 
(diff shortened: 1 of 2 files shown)


src/components/SingleDatePicker.jsx
```diff
@@ -203,7 +202,7 @@ class SingleDatePicker extends React.PureComponent {
     });
 
     onFocusChange({ focused: false });
-    onClose({ startDate, endDate });
+    onClose({ date});
```
ljharb:
```suggestion
    onClose({ date });
```

AlokTakshak: done

AlokTakshak: https://github.com/airbnb/react-dates/blob/11de2016bf6e7111ddebd5a7f7fc1fe929a3600c/test/components/SingleDatePicker_spec.jsx#L407
in this test suite should i add a test for `calls props.onFocusChange ` with  `date` props ?

ljharb: that sounds great, yes

AlokTakshak: can i run particular test file?

AlokTakshak: done

lencioni:
Merged in 3290bb4

PR #1684, opened 2019-06-14
Added support for custom input components

### LINKED ISSUE #1549: Allow to use custom input
It would be nice to be able to use custom input element, i.e. input from `react-input-autosize` or any other controllable component with at least `value` and `onChange` props

lucasconstantino:
Adds support for custom input components, not modifying the date-pickers at all.

Fixes #1549

Example usage:

```jsx
import { SingleDatePicker } from 'react-dates'
import TextField from '@material-ui/core/TextField'

<SingleDatePicker component={ TextField } />
```

> One gotcha is you haver to make sure you handle the passed `innerRef`. Attach it to an HTML element, for gods sake.

### DIFF
--- a/src/utils/componentPropType.js
+++ b/src/utils/componentPropType.js
@@ -1,0 +1,31 @@
+/* eslint-disable consistent-return */
+import { isValidElementType } from 'react-is';
+
+const createComponentPropType = isRequired => (
+  props,
+  propName,
+  componentName,
+  location,
+  propFullName,
+) => {
+  const prop = `${location} \`${propFullName}\``;
+  const comp = `\`${componentName}\``;
+  const value = props[propName];
+
+  if (isRequired && value == null) {
+    throw new Error(
+      `The ${prop} is marked as required in ${comp}, but its value is \`${
+        value === null ? 'null' : 'undefined'
+      }\``,
+    );
+  } else if (value && !isValidElementType(value)) {
+    return new Error(
+      `Invalid prop ${prop} supplied to ${comp}: the prop is not a valid React component`,
+    );
+  }
+};
+
+const componentPropType = createComponentPropType(false);
+componentPropType.isRequired = createComponentPropType(true);
+
+export default componentPropType;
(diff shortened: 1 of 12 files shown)


ljharb:
Can you elaborate a bit more on why we’d want to add this complexity?

src/utils/componentPropType.js
```diff
@@ -0,0 +1,31 @@
+/* eslint-disable consistent-return */
+import { isValidElementType } from 'react-is';
+
+const createComponentPropType = isRequired => (
```
ljharb: This already exists in airbnb-prop-types as elementType.

src/utils/getRenderComponent.jsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+
+export default function getRenderComponent(Component) {
+  return typeof Component === 'string'
+    ? Component
+    : React.forwardRef((props, ref) => <Component {...props} innerRef={ref} />);
```
ljharb: We definitely can’t add forwardRef here, because we support down to react 0.14.

tangcundi:
support for custom input components is really useful, hope this PR can be merged ASAP :)

GoPro16:
We were able to get pretty close with the `reactstrap-date` that runs on `bootstrap`. If this was implemented we would be able to write less style overrides and hook more into the error styles the bootstrap applies by default with its classes.

[Storybook Link](https://availity.github.io/availity-react/?path=/story/components-avdate--avdaterangefield)

ndbroadbent:
This would be really useful! My use-case is that I really need to set a dynamic font-size as an inline style property on the `<input>` element, and I couldn't see any way to do that. This change would make it possible to pass in a functional component so that I can render my own  `<input>` with additional props.

ndbroadbent:
I've just tried this out in a fork, and I think there is something wrong with the way the event handlers are being passed. If I just use `<SingleDatePicker ... />` without the `component` prop, then I can type in a date. 

But the input loses focus after the first keypress if I do this:

```
<SingleDatePicker
  component={props => <input ref={props.innerRef} {...props} />}
/>
```

(But I'm not sure if this is due to some keyboard handlers in my own code.)

@lucasconstantino Am I setting up the component correctly, and should this have the same behavior as the original `'input'`?

JavaJamie:
What's happening with this??? I'm using Chakra UI so need to use a custom textfield...

PR #1685, opened 2019-06-14
Added basic className support

lucasconstantino:
Added initial className based styling support (focus on styled-components).

### DIFF
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -49,6 +49,7 @@ const defaultProps = {
   focusedInput: null,
 
   // input related props
+  className: undefined,
   startDatePlaceholderText: 'Start Date',
   endDatePlaceholderText: 'End Date',
   startDateAriaLabel: undefined,
@@ -539,6 +540,7 @@ class DateRangePicker extends React.PureComponent {
 
   render() {
     const {
+      className,
       startDate,
       startDateId,
       startDatePlaceholderText,
@@ -638,6 +640,7 @@ class DateRangePicker extends React.PureComponent {
         {...css(
           styles.DateRangePicker,
           block && styles.DateRangePicker__block,
+          className,
         )}
       >
         {enableOutsideClick && (
--- a/src/components/SingleDatePicker.jsx
+++ b/src/components/SingleDatePicker.jsx
@@ -46,6 +46,7 @@ const defaultProps = {
   focused: false,
 
   // input related props
+  className: undefined,
   id: 'date',
   placeholder: 'Date',
   ariaLabel: undefined,
@@ -520,6 +521,7 @@ class SingleDatePicker extends React.PureComponent {
 
   render() {
     const {
+      className,
       id,
       placeholder,
       ariaLabel,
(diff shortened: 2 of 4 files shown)


mattaningram:
Preventing people from using basic styling concepts like a wrapping class name makes the barrier to entry for using `react-dates` much higher. Not only is it a bit much to ask newer designers/developers to switch to CSS-in-JS for styling just one component, it also can make a codebase much more complex if existing styling uses "traditional" CSS.

It also feels inconsistent since you already allow react-dates to initialize with styleable class names, and this would be just one more wrapping class name.

Considering how few robust date selector libraries there are for React out there, it would be nice if react-dates was more friendly to projects that don't use CSS-in-JS.

ljharb:
You can already use the `.css` files we provide, and override them with normal CSS - `className` and `style` props *are* CSS-in-JS already.

mattaningram:
I can't specifically target a `appendToBody` date picker because I can't add a custom class name to it, or wrap it in a custom class. So all the provided CSS files give me a theme which I have changed to match my branding, but it doesn't let me fix positioning issues that haven't been addressed or bug-fixed yet.

For example we offset and overflow lock our page when a modal is open so that it doesn't jump to the top of the page if you opened it when scrolled down. To get date pickers to appear nicely in the modal they have to be appended to the body and have `position: fixed` set on them because `react-dates` does not take into account the body offset in it's positioning. I can append them to the body with the `appendToBody` prop, but then they are no longer in the modal so I can't specifically target it without styling date pickers not in that context

All I need from `className` is the ability to add a class to the panel wrapper so I can style it with normal CSS in such a situation.

k-funk:
> Preventing people from using basic styling concepts like a wrapping class name makes the barrier to entry for using `react-dates` much higher. Not only is it a bit much to ask newer designers/developers to switch to CSS-in-JS for styling just one component, it also can make a codebase much more complex if existing styling uses "traditional" CSS.
> 
> It also feels inconsistent since you already allow react-dates to initialize with styleable class names, and this would be just one more wrapping class name.
> 
> Considering how few robust date selector libraries there are for React out there, it would be nice if react-dates was more friendly to projects that don't use CSS-in-JS.

To this point, I am trying to use bootstrap with this datepicker. I need to add `.form-control` to the `<input />` so that the input field stylistically fits with my other form elements, but looks like that's not possible in this lib. This means that I need to break my project's pattern of using css classnames to style generic components, and do something like
```
.DateInput_input {
  // copy-paste from bootstrap's .form-control + possibly other .form-control related classes or nested styles
  // hope bootstrap doesn't change too much in between version bumps
}
```
Maybe if scss's `@extend` wasn't [complete trash](https://webinista.com/updates/dont-use-extend-sass/), but that's another topic.

macqm:
I'd love to see this merged to be able to target styles to a date picker that is rendered with `appendToBody`.
We use `appendToBody` to overcome issue of the parent container element setting `overflow: hidden`, however we have multiple users of the react-date control across the app. Currently, we cannot target the styles to our specific usage.

oriooctopus:
I have the same issue as @macqm . Without a wrapper class, I can't target styles to a specific usage on the page @ljharb

ljharb:
You can add your own wrapper CSS class name on a container; you don’t need react-dates to have one.

PR #1690, opened 2019-06-17
Run SVGs through SVGO with most aggressive settings

lencioni:
I noticed that these SVGs could be optimized a little more, by reducing
the precision. The end result is slightly different, but at the size
they are used, I don't think it will really be perceptible.

I left close.svg alone because it has a 12x12 viewBox instead of a
1000x1000 viewbox, so the precision here is actually pretty important
for the shape.

I think some more bytes could maybe be squeezed out of these by
recreating them from hand using different shapes instead of paths, but
that's a bit more work than I'd like to do at this time.

This optimization reduces the size of the svg directory from 2678 bytes
to 2106 bytes.

Screenshots before:

image

image

image

image

Screenshots after: 

image

image

image

image


ljharb:
Can you share a visual diff? 500bytes is pretty tiny.

lencioni:
@ljharb I've added some screenshots

lencioni:
@ljharb probably not worth the effort. Also, svgo removes the focusable attribute which I had to manually add back in.

PR #1689, opened 2019-06-17
Use @babel/plugin-transform-runtime for helpers

lencioni:
Currently when this project is built, a bunch of babel helpers end up
being inlined into each module as needed. With Babel 7, we can use
@babel/plugin-transform-runtime to add imports for @babel/runtime
modules instead of inlining and duplicating these helper functions. This
should reduce bundle size impact of this project a fair amount.

Here's an example diff of one of the components:

```diff
0a1,4
> import _extends from "@babel/runtime/helpers/esm/extends";
> import _assertThisInitialized from "@babel/runtime/helpers/esm/assertThisInitialized";
> import _inheritsLoose from "@babel/runtime/helpers/esm/inheritsLoose";
> import _objectSpread from "@babel/runtime/helpers/esm/objectSpread";
2,12d5
<
< function _extends() { _extends = Object.assign || function (target) { for (var i = 1; i < arguments.length; i++) { var source = arguments[i]; for (var key in source) { if (Object.prototype.hasOwnProperty.call(source, key)) { target[key] = source[key]; } } } return target; }; return _extends.apply(this, arguments); }
<
< function _assertThisInitialized(self) { if (self === void 0) { throw new ReferenceError("this hasn't been initialised - super() hasn't been called"); } return self; }
<
< function _inheritsLoose(subClass, superClass) { subClass.prototype = Object.create(superClass.prototype); subClass.prototype.constructor = subClass; subClass.__proto__ = superClass; }
<
< function _objectSpread(target) { for (var i = 1; i < arguments.length; i++) { var source = arguments[i] != null ? arguments[i] : {}; var ownKeys = Object.keys(source); if (typeof Object.getOwnPropertySymbols === 'function') { ownKeys = ownKeys.concat(Object.getOwnPropertySymbols(source).filter(function (sym) { return Object.getOwnPropertyDescriptor(source, sym).enumerable; })); } ownKeys.forEach(function (key) { _defineProperty(target, key, source[key]); }); } return target; }
<
< function _defineProperty(obj, key, value) { if (key in obj) { Object.defineProperty(obj, key, { value: value, enumerable: true, configurable: true, writable: true }); } else { obj[key] = value; } return obj; }
<
```

This change reduces the size of the esm build directory from 380,051 bytes to 361,346 bytes

PR #1692, opened 2019-06-18
Update babel-preset-airbnb 3.2.1 → 4.0.0

lencioni:
This version adds `@babel/plugin-transform-runtime` to help us reduce
bundle size.

This should have the same effect as
https://github.com/airbnb/react-dates/pull/1689

### DIFF
--- a/package.json
+++ b/package.json
@@ -67,7 +67,7 @@
     "babel-plugin-inline-svg": "^1.0.0",
     "babel-plugin-istanbul": "^5.1.4",
     "babel-plugin-transform-replace-object-assign": "^2.0.0",
-    "babel-preset-airbnb": "^3.2.1",
+    "babel-preset-airbnb": "^4.0.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -110,6 +110,7 @@
     "why-did-you-update": "^1.0.6"
   },
   "dependencies": {
+    "@babel/runtime": "^7.4.5",
```
ljharb: This should also be a peer dep as well as a dep (oops, v3.3 should have been a v4 because of the peer dep requirement)

lencioni: Why should it be a peerDep?

ljharb: i suppose it doesn't have to me; it'd ensure maximal deduping, but since it's not stateful i guess we can rely on npm for that.

lencioni: Cool, that's what I thought as well but peerDeps are always a bit confusing to me so I wanted to make sure.

PR #1707, opened 2019-06-25
Defer day focusing until next animation frame

lencioni:
At Airbnb we noticed a bug affecting only mobile devices or desktop with
mobile device emulation enabled. In months that span 5 weeks that are
followed by months that span 6 weeks (e.g. May 2019), the scroll
position of the calendar days was scrolled down some, causing the
numbers to awkwardly overlap with the days of the week and look broken.
This only happened when first opening the date picker, and once you go
to a different month it would fix itself.

After some sleuthing, I determined that this was caused by these
.focus() calls happening at a weird time, causing a parent container (I
think maybe the DayPicker transition container with overflow hidden) to
have a > 0 scrollTop.

It seems that we can prevent this from happening by deferring the focus
until the next animation frame. I think this could also help us avoid
some layout thrashing, which could give us a little performance boost
here.

I was entirely unable to reproduce this in react-dates storybook, but
testing this fix out in the Airbnb codebase seems to fix the issue.

Before fix:
image

After fix:
image

### DIFF
--- a/src/components/CalendarDay.jsx
+++ b/src/components/CalendarDay.jsx
@@ -58,7 +58,11 @@ class CalendarDay extends React.PureComponent {
     const { isFocused, tabIndex } = this.props;
     if (tabIndex === 0) {
       if (isFocused || tabIndex !== prevProps.tabIndex) {
-        this.buttonRef.focus();
+        requestAnimationFrame(() => {
+          if (this.buttonRef) {
+            this.buttonRef.focus();
+          }
+        });
       }
     }
   }
--- a/src/components/CustomizableCalendarDay.jsx
+++ b/src/components/CustomizableCalendarDay.jsx
@@ -228,7 +228,11 @@ class CustomizableCalendarDay extends React.PureComponent {
     const { isFocused, tabIndex } = this.props;
     if (tabIndex === 0) {
       if (isFocused || tabIndex !== prevProps.tabIndex) {
-        this.buttonRef.focus();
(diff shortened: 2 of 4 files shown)


src/components/CalendarDay.jsx
```diff
@@ -58,7 +58,11 @@ class CalendarDay extends React.PureComponent {
     const { isFocused, tabIndex } = this.props;
     if (tabIndex === 0) {
       if (isFocused || tabIndex !== prevProps.tabIndex) {
-        this.buttonRef.focus();
+        requestAnimationFrame(() => {
```
nkinser: should we `cancelAnimationFrame` when the component unmounts?

ljharb: do we already reference rAF elsewhere in the repo, or is this dependency a potential breaking change?

lencioni: This looks like the first usage

lencioni: I don't think we need to, since I included a check that this.buttonRef is still truthy, so if it unmounts in between, it will essentially be a noop anyway.

lencioni:
@nkinser I think it could, but hard to say.

lencioni:
@ljharb I added some tests

src/components/CalendarDay.jsx
ljharb: any chance we could use https://npmjs.com/raf instead of the global?

Issue #1703, opened 2019-06-24
RangePicker: start date of the range not colored correctly


ktrungha:
I have a date range picker with default range.
When the picker is first opened, the start date of the range is not colored correctly.
Only the end date of the range has the darker color.

The start date is colored correctly if I manually pick the range after the picker opens.

I am not using css to style this picker. I use theming.

In the photo, you can see that the 24th day is not colored correctly.

![SelectedStart](https://user-images.githubusercontent.com/6115540/59986365-c0f7c180-9660-11e9-99d2-12e235b9d83f.png)

mmarkelov:
@ktrungha your code will be helpful

teagueash:
This issue has been the default behavior for a while now. Just look at their storybook example:
[http://airbnb.io/react-dates/?path=/story/drp-input-props--default](url)

Source code here: [https://github.com/airbnb/react-dates/blob/master/examples/DateRangePickerWrapper.jsx](url)

ktrungha:
I look at the html element. It looks like the css class 'selected' is applied to the end date but not the start date. That's why the end date is colored  with the darker color but not the start data.

ktrungha:
Looks like this is a bug reported here https://github.com/airbnb/react-dates/issues/1599

lencioni:
Fixed by #1710

PR #1710, opened 2019-06-26
Fix-Start-Date-Style

### LINKED ISSUE #1599: startDate UI is broken if startDate time is not 12:00:00
startDate UI is broken even on airbnb.com (works correctly in examples at http://airbnb.io/react-dates/)

this is because of this stuff  https://github.com/airbnb/react-dates/pull/310

Screenshot 2019-03-27 13 23 19

mmarkelov:
Will fix #1599, #1703


lencioni:
@mmarkelov thanks for the PR! Any chance you could add a regression test to prevent this from being reintroduced?

lencioni:
Merged in 13ee107

PR #1718, opened 2019-07-02
[Fix] `CalendarMonth`: recompute month title height when rerendering

cedrics:
**react-dates version**
e.g. react-dates@20.2.5

**Describe the bug**
When passing in a different function for `renderMonthTitle` on each render call, the `DayPicker` is hidden on subsequent rerenders.

If you change the [`DateRangePickerWrapper`](https://github.com/airbnb/react-dates/blob/master/examples/DateRangePickerWrapper.jsx#L141-L148) to the following code, you can reproduce the issue in storybook.

```
<DateRangePicker
  startDateId="your_unique_start_date_id"
  endDate={this.state.endDate}
  endDateId="your_unique_end_date_id"
  onDatesChange={({ startDate, endDate }) => this.setState({ startDate, endDate })}
  onFocusChange={focusedInput => this.setState({ focusedInput })}
  renderMonthText={() => "test"}
/>
```

Select a start date and then select an end date that is before the start date (which forces a rerender and not the daypicker to destroyed and readded).

**Screenshots/Gifs**

bug-react-dates

**Bug Description**

When the `renderMonthTitle` functions changes the calculated `monthTitleHeight` gets reset (See [code](https://github.com/airbnb/react-dates/blob/master/src/components/DayPicker.jsx#L308-L312)). However, the `CalendarMonth` currently only computes the height on mount. This changes the behavior of the component to recompute the height whenever it is passed the callback to set it.

### DIFF
--- a/test/components/CalendarMonth_spec.jsx
+++ b/test/components/CalendarMonth_spec.jsx
@@ -1,8 +1,9 @@
 import React from 'react';
 import { expect } from 'chai';
-import { shallow } from 'enzyme';
+import { shallow, mount } from 'enzyme';
 import sinon from 'sinon-sandbox';
 import moment from 'moment';
+import describeIfWindow from '../_helpers/describeIfWindow';
 
 import CalendarMonth from '../../src/components/CalendarMonth';
 
@@ -45,6 +46,43 @@ describe('CalendarMonth', () => {
       expect(typeof onYearSelect).to.equal('function');
       expect(typeof isVisible).to.equal('boolean');
       expect(wrapper.find('#month-element').exists()).to.equal(true);
+    });
+
+    describeIfWindow('setMonthTitleHeight', () => {
+      beforeEach(() => {
+        sinon.stub(window, 'setTimeout').callsFake((handler) => handler());
+      });
+
+      it('sets the title height after mount', () => {
+        const setMonthTitleHeightStub = sinon.stub();
+        mount(
+          <CalendarMonth
+            isVisible
+            setMonthTitleHeight={setMonthTitleHeightStub}
+          />,
+        );
+
+        expect(setMonthTitleHeightStub).to.have.property('callCount', 1);
+      });
+
+      describe('if the callbacks gets set again', () => {
+        it('updates the title height', () => {
+          const setMonthTitleHeightStub = sinon.stub();
+          const wrapper = mount(
+            <CalendarMonth
+              isVisible
+              setMonthTitleHeight={setMonthTitleHeightStub}
+            />,
+          );
+
+          expect(setMonthTitleHeightStub).to.have.property('callCount', 1);
+
+          wrapper.setProps({ setMonthTitleHeight: null });
+
+          wrapper.setProps({ setMonthTitleHeight: setMonthTitleHeightStub });
+          expect(setMonthTitleHeightStub).to.have.property('callCount', 2);
+        });
+      });
     });
   });
 });
--- a/src/components/CalendarMonth.jsx
+++ b/src/components/CalendarMonth.jsx
@@ -106,7 +106,7 @@ class CalendarMonth extends React.PureComponent {
   }
 
   componentDidMount() {
-    this.setMonthTitleHeightTimeout = setTimeout(this.setMonthTitleHeight, 0);
+    this.queueSetMonthTitleHeight();
   }
 
   componentWillReceiveProps(nextProps) {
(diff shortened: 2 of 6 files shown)


cedrics:
@ljharb sure, I have added some regression tests. As the code to test uses both refs and setTimeout the test is fairly complicated and kind of ugly 😄 If you have some pointers how to improve it, I would appreciate it.

I also stole the `describeIfWindow` function from the [DateRangePicker spec](https://github.com/airbnb/react-dates/blob/master/test/components/DateRangePicker_spec.jsx#L18). Should I extract it to a common util method? If so should I just add it to [test/_helpers](https://github.com/airbnb/react-dates/tree/master/test/_helpers)

test/components/CalendarMonth_spec.jsx
```diff
@@ -46,5 +48,50 @@ describe('CalendarMonth', () => {
       expect(typeof isVisible).to.equal('boolean');
       expect(wrapper.find('#month-element').exists()).to.equal(true);
     });
+
+    describeIfWindow('setMonthTitleHeight', () => {
+      const awaitChange = fn => new Promise((resolve, reject) => {
```
ljharb: This seems very brittle; it’s explicitly racing against the change. These tests should find a way to explicitly wait for the proper time instead.

test/components/CalendarMonth_spec.jsx
```diff
@@ -46,5 +48,50 @@ describe('CalendarMonth', () => {
...
+      it('sets the title height after mount', () => {
+        const setMonthTitleHeightStub = sinon.stub();
+        mount(
+          <CalendarMonth
+            isVisible
+            setMonthTitleHeight={setMonthTitleHeightStub}
+          />,
+        );
+
+        return awaitChange(() => expect(setMonthTitleHeightStub.callCount).to.equal(1));
```
ljharb:
```suggestion
        return awaitChange(() => expect(setMonthTitleHeightStub).to.have.property('callCount', 1));
```
this has a nicer failure message; same throughout

ljharb:
Yes, moving it to a common place would be great.

test/components/CalendarMonth_spec.jsx
```diff
@@ -1,11 +1,36 @@
...
+      try {
+        fn();
+        resolve();
+      } catch (e) {
+        if (Date.now() - startTimestamp > MAX_WAIT_TIME_IN_MS) {
+          reject(e);
+          return;
+        }
+
+        setTimeout(checkAndWait, 10);
```
ljharb: Sorry if i wasn’t clear - any usage of setTimeout at all, or anything based on the concept of waiting for some amount of time, is inherently brittle and shouldn’t be used. We need access to the actual async behavior so we can explicitly await on it.

cedrics: I am not sure what you are proposing. Could you describe in more concrete terms how this change might look like?

ljharb: Since you want to wait until `wrapper.instance().setMonthTitleHeightTimeout` has ran, at the very least, you'd want to poll until that had become null. However, in this case, I think you could do something like:
```jsx
let timeout = wrapper.instance().setMonthTitleHeightTimeout;
const stub = sinon.stub(wrapper.instance(), 'setMonthTitleHeightTimeout');
stub.get(() => timeout);
const nulled = new Promise((resolve) => {
  stub.set((value) => {
    if (timeout !== null && value === null) resolve();
    timeout = value;
  });
});
return nulled.then(() => {
  expect(whatever);
});
```

cedrics: Thanks, I think I get what you want now. Your solution is already coupled to the actual implementation, so a much more straight forward solution would be to just stub out `setTimeout` for the this test and make it synchronous. The asynchronicity is irrelevant for this test. This would reduce coupling to between the implementation and the tests but may result in some weird behavior if some other piece of code relies on setTimeout be asynchronous. Your call 😸

ljharb: Coupling tests to the implementation is fine as long as changing the impl in a way that breaks it, breaks the tests.

That includes stubbing setTimeout, which could certainly work.

huvber:
👍

ljharb (approved):
Moved the setTimeout to a shared method so that it'd have one source of truth.

Also, made the describeIfWindow a default export, since default exports are almost always preferred.

ljharb:
Coverage is going down because we're not testing mount things right now; merging.

PR #1722, opened 2019-07-05
chore(package): update react-with-styles to version 3.2.3

scottsheffield:
Previous version is breaking (resolving to 3.2.2)

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,7 +124,7 @@
     "react-outside-click-handler": "^1.2.0",
     "react-portal": "^4.1.5",
     "react-with-direction": "^1.3.0",
-    "react-with-styles": "^3.2.0",
+    "react-with-styles": "^3.2.3",
     "react-with-styles-interface-css": "^4.0.2"
   },
   "peerDependencies": {


oshanley:
For context for anyone else: https://github.com/airbnb/react-dates/issues/1720

oshanley:
@majapw @ljharb Could either of you please take a look when you have the chance? This dependency is breaking the master build of a large product

ljharb:
This isn’t a chore, and isn’t necessary, since 3.2.3 is included in the semver range for `^3.2.0`.

What problem is this solving for you that an updated lockfile wouldn’t?

scottsheffield:
Greenkeeper provided the commit message and I figured it was configured well enough for this task. You've got a point there, and we don't have a lockfile to begin with, so I surmised that whatever built entity we're pulling from ends up old. 

I'll see about fixes that don't necessitate this change on our end, though I don't see much harm in cutting a new minor to explicitly avoid a broken dep.

ljharb:
It’d be a new patch, and it’s fine to do the update; but if you don’t have a lockfile then `rm -rf node_modules && npm install` should be installing the latest version of r-w-s.

scottsheffield:
Yes, sorry, misspoke (:s/minor/patch). I think our issue is coming from our use of artifactory.

ljharb:
I’ve definitely seen issues where artifactory fails to pick up public updates; if that’s the case then this PR wouldn’t fix your problem, it would make react-dates unusable.

scottsheffield:
I don't follow that conclusion, but yes I do think seeing about refreshing artifactory should be my focus.

ljharb:
The conclusion is because the implication is that your artifactory doesn't contain v3.2.3 of r-w-s at all.

PR #1732, opened 2019-07-25
Bump the dependency on react-with-styles to ^3.2.3

### LINKED ISSUE #1198: Uncaught TypeError: modifiers.has is not a function

### LINKED ISSUE #1493: Always have an error : Failed prop type: Invalid prop `modifiers` of type `Object` supplied to `withStyles(CalendarDay)`, expected instance of `Set`.
Hi, 

The minimum REQUIRED setup seems to not work well... What I missed?

The thing is the calendar seems working but I always have an error in console. I don't know why. 
It's the same thing with `DateRangerPicker`, `SingleDatePicker`...

Help please

```
class DatePicker extends React.PureComponent {
  constructor(props) {
    super(props);
    this.state = {
      focused: null,
      date: moment(),
    };
  }

  render() {
    return (
      <SingleDatePicker
        date={this.state.date} // momentPropTypes.momentObj or null
        onDateChange={date => this.setState({ date })} // PropTypes.func.isRequired
        focused={this.state.focused} // PropTypes.bool
        onFocusChange={({ focused }) => this.setState({ focused })} // PropTypes.func.isRequired
      />
    );
  }
}
```

capture d ecran 2018-12-13 a 10 51 31

NMinhNguyen:
This closes #1198, closes #1493.

/cc @ljharb

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,7 +124,7 @@
     "react-outside-click-handler": "^1.2.0",
     "react-portal": "^4.1.5",
     "react-with-direction": "^1.3.0",
-    "react-with-styles": "^3.2.0",
+    "react-with-styles": "^3.2.3",
     "react-with-styles-interface-css": "^4.0.2"
   },
   "peerDependencies": {


NMinhNguyen:
> should automatically bring in the updates

@ljharb yeah, except when you already have `react-with-styles` pinned to that version in your lockfile. Can't speak for npm as I use Yarn, but Yarn in the presence of an existing `^3.2.0 -> 3.2.1` resolution, will keep it, even if you bump `react-dates` :slightly_smiling_face:

ljharb:
Sure, but that's the burden you take on when you have a lockfile - you have to manually update transitive deps. Yarn's behavior seems broken to me; `npm install react-dates` generally will update all the transitive deps of it in the lockfile.

NMinhNguyen:
I'll prob struggle to find references right now, but I'm pretty sure I read somewhere on Yarn's GitHub that it behaves that way so as to not introduce accidental regressions (it is a lockfile after all) - i.e. things are working fine with resolution X, why should a subsequent install change that? Or perhaps, you even have this dependency in your own `package.json` (or maybe another package has it), and you know it works with that exact version - installing another dependency shouldn't change that 🙂

ljharb:
It should because that’s how semver works.

Issue #1323, opened 2018-08-23
renderCalendarInfo cause memory leak


WatsonFungHK:
I am trying to insert a text link in calendar.
`<DateRangePicker renderCalendarInfo={() => <CustomText />} ...moreProps />`

But after user select the date, and the calendar is hided, a warning is thrown

> Warning: Can't call setState (or forceUpdate) on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.

I confirmed it is not caused by my custom react component because event if I render `() => <b>HELLO WORLD</b>`, the warning is still here

Is it a hidden issue of react-dates?

majapw:
Um, can you share all of the props you are using give a demo app? Do you no longer see the issue if you remove `renderCalendarInfo`? What version of react-dates/react are you using?

mstanaland:
This issue is aging, but I'm seeing the same thing. It's even reproducible for me running react-dates's own storybook locally (i.e. cloning react-dates > `npm install` > `npm run storybook` > then going to the "with info panel bottom" story). 

However, I do not see the error at `http://airbnb.io/react-dates/?path=/story/drp-calendar-props--with-info-panel-bottom`. 🤷‍♂️

The error is not consistent, but is most reproducible when tabbing through the inputs instead of clicking.

The problem appears to be in DayPicker.jsx, line 322 where `this.setCalendarInfoWidthTimeout` is assigned a timeout. I _think_ what is happening is that timeout is sometimes being duplicated, so it's not being cleared correctly in componentWillUnmount.

Adding `clearTimeout(this.setCalendarInfoWidthTimeout);` above line 322 resolved the issue for me.
```
if (this.calendarInfo) {
  clearTimeout(this.setCalendarInfoWidthTimeout);
  this.setCalendarInfoWidthTimeout = setTimeout(() => {
    ...
```

ljharb:
@mstanaland thanks, a PR adding that would be appreciated.

mstanaland:
@ljharb My pleasure >> PR #1736

PR #1736, opened 2019-07-28
Clear renderCalendarInfo timeout before assigning a new timeout (#1323)

mstanaland:
Clears the setTimeout assigned to this.setCalendarInfoWidthTimeout before a new setTimeout is assigned/reassigned to prevent a possible memory leak.

Fixes issue #1323

### DIFF
--- a/src/components/DayPicker.jsx
+++ b/src/components/DayPicker.jsx
@@ -319,6 +319,8 @@ class DayPicker extends React.PureComponent {
     // breaks the CSS transition.
     // The setTimeout will wait until the transition ends.
     if (this.calendarInfo) {
+      clearTimeout(this.setCalendarInfoWidthTimeout);
+
       this.setCalendarInfoWidthTimeout = setTimeout(() => {
         const { calendarInfoWidth } = this.state;
         const calendarInfoPanelWidth = calculateDimension(this.calendarInfo, 'width', true, true);


JimmyGray:
I'm getting the same issue. Any chance of getting the fix merged?

ljharb:
Sure, if it adds a regression test. Feel free to post a link to a branch and I'll pull it into the PR.

smit1625:
I'm running into this issue as well, and willing to try and help out. Can we define what is meant by "regression test" in this context? I am having trouble wrapping my head around the idea of writing a regression test for a memory leak warning.

ljharb:
@smit1625 that's a totally fair question :-) mainly, a test that would have otherwise set the timeout twice, then advances the clock past both, and then asserts that only one setState happened.

smit1625:
@ljharb okay, understood. I have cloned this fork of the codebase and am starting to try and get such a test together. A few notes:

- As far as I can tell, this project is not using jest, so I don't have access to the usual tools that I am accustomed to testing `setTimeout` with (see https://jestjs.io/docs/en/timer-mocks). I am not very well-versed in enzyme, and would greatly appreciate being pointed in the right direction as to how a test should deal with these concepts.
- Are you aware that the tests for `DayPicker`'s lifecycle methods (including `componentWillUpdate`) have been nested within a `describe.skip` (line 808 on this fork, 897 on master) since this project's _initial commit_? I have un-skipped them and spent about a half hour trying to figure out which of these tests are useful, but unsurprisingly they don't appear to have been kept up to date with the project. So, I'm left wondering, if we haven't been diligently testing the functionality of these (now-deprecated) lifecycle methods, why are we drawing this line in the sand here and now? My personal feeling is that this is the type of situation where the existing tests should be enough of a safeguard, and there is probably very little value to counting the number of times that `setTimeout` is invoked.

smit1625:
@ljharb I have created a new branch off of master with the test I think you were looking for. Please let me know if there's anything else I can do to be helpful in this matter.

https://github.com/airbnb/react-dates/pull/2021

PR #1741, opened 2019-08-02
Fix ios keyboard accessibility

mmarkelov:
Why it was done?

I just notice some bugs with accessibility in IOs devices with opened keyboard.
It is especially noticeable with DatePickers without **readOnly** property:
bug
So after clicking on 'Done' keyboard buttons nothing happened, but picker should close on it.

Also there are some rare cases with DatePickers with **readOnly** property:
As example airbnb.com:
bug2
I can open picker, don't select any date, but then I just click navigation arrow button inside of keyboard - and also nothing happened, but it should be closed.

### DIFF
--- a/src/components/DateInput.jsx
+++ b/src/components/DateInput.jsx
@@ -43,6 +43,7 @@ const propTypes = forbidExtraProps({
 
   onChange: PropTypes.func,
   onFocus: PropTypes.func,
+  onBlur: PropTypes.func,
   onKeyDownShiftTab: PropTypes.func,
   onKeyDownTab: PropTypes.func,
 
@@ -71,6 +72,7 @@ const defaultProps = {
 
   onChange() {},
   onFocus() {},
+  onBlur() {},
   onKeyDownShiftTab() {},
   onKeyDownTab() {},
 
@@ -180,6 +182,7 @@ class DateInput extends React.PureComponent {
       focused,
       showCaret,
       onFocus,
+      onBlur,
       disabled,
       required,
       readOnly,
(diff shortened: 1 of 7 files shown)


backwardok:
Will this cause a problem with when focus is inside the calendar and no longer keyboard focused on the input itself?

mmarkelov:
@backwardok I've checked lot of cases and it was OK. If you are unsure with something I can modify PR

mmarkelov:
Another catch:
After merging https://github.com/airbnb/react-dates/commit/75bba6b7eb92f57727b45bacdb83221eee6667b6 I can't move between inputs in **DateRangePicker** and adding onBlur to **DateInput** also fix this problem.
bug

backwardok:
@mmarkelov looks like from that gif that you can no longer tab into the calendar. That would make it impossible to set the start date from the calendar with a keyboard

mmarkelov:
@backwardok yeah all these gifs represent current behavior. I can make gifs to show what would this PR change

backwardok:
@mmarkelov Is the gif you posted in https://github.com/airbnb/react-dates/pull/1741#issuecomment-518559542 not the latest behavior? Can you verify whether or not you can tab into the calendar and set the date that way?

mmarkelov:
@backwardok all of posted gifs represent master branch. Gif, that I posted in comment show that there is problem right now with moving throw inputs in iOS right now. It’s ok in latest release. And with this PR it is totally fine. Hope, it’s clear now :)

backwardok:
Hmm... @majapw or @nkinser mind taking a look? I feel like you two have more context into the blur handling than I do

PR #1746, opened 2019-08-08
[a11y] Add `aria-roledescription`

rcourtie:
As per the [recommendation of the W3C](https://www.w3.org/TR/wai-aria-1.1/#application), adding an `aria-roledescription` to describe the role of the datepicker application. Since this is supposed to be a human-readable and localized description of the element it's passed in through the existing phrases prop, and a default value was included.

### DIFF
--- a/src/components/DayPicker.jsx
+++ b/src/components/DayPicker.jsx
@@ -1027,6 +1027,7 @@ class DayPicker extends React.PureComponent {
     return (
       <div
         role="application"
+        aria-roledescription={phrases.roleDescription}
         aria-label={phrases.calendarLabel}
         {...css(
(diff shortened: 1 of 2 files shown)


alejandrogarciasalas:
@rcourtie @backwardok This seems to be failing with axe tests, I am getting the following error:

Screen Shot 2019-09-18 at 4 12 24 PM

what's the best way to proceed, should we ignore this tests or remove the label?

backwardok:
@alejandrogarciasalas since it's providing additional information and nothing super crucial (there are additional pieces of information that are also provided to help with what this component is), I think it's ok to ignore that warning

Issue #1484, opened 2018-12-07
Prevent first available day of month from focusing


dvprz:
Hello,

Thank you all for the wonderful library!

So we're working with a team of a11y auditors and one of their suggestion goes against what react-dates does out of the box. When a user tabs to a prev/next button and hits enter after the month transition ends the first available calendar day receives focus. Our consultants want us to keep the focus on the prev/next button. 

I have looked at the source and perhaps I have missed it but I don't see anyway to modify this using the available api. Any ideas?

backwardok:
I agree with this and actually was just looking through issues to see if it was already filed. I think this behavior should be the default, and I would recommend updating the library to change the behavior

mmarkelov:
@backwardok I suppose it is here:
https://github.com/airbnb/react-dates/blob/75bba6b7eb92f57727b45bacdb83221eee6667b6/src/components/CalendarDay.jsx#L58-L69
But I not sure that it won't broke screen readers

PR #1747, opened 2019-08-09
[A11y] Keep keyboard focus on navigation buttons after click

nkinser:
This PR makes it so that when a keyboard only user selects one of the navigation buttons the focus stays on the button rather than shifting to the first available date on the calendar. This makes it much easier for a keyboard user to see the availability of a month more than 1 month in the future. Here is the [jira ticket](https://jira.airbnb.biz/browse/PRODUCT-83356) filed for this bug.

#### Before
calendar-nav-tabbing-bug

#### After
calendar-nav-tabbing-fix

### DIFF
--- a/src/components/CalendarDay.jsx
+++ b/src/components/CalendarDay.jsx
@@ -55,10 +55,10 @@ class CalendarDay extends React.PureComponent {
     this.setButtonRef = this.setButtonRef.bind(this);
   }
 
-  componentDidUpdate(prevProps) {
+  componentDidUpdate() {
     const { isFocused, tabIndex } = this.props;
     if (tabIndex === 0) {
-      if (isFocused || tabIndex !== prevProps.tabIndex) {
+      if (isFocused) {
         raf(() => {
           if (this.buttonRef) {
             this.buttonRef.focus();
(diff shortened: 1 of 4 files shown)


src/components/CalendarDay.jsx
```diff
@@ -55,10 +55,10 @@ class CalendarDay extends React.PureComponent {
     this.setButtonRef = this.setButtonRef.bind(this);
   }
 
-  componentDidUpdate(prevProps) {
+  componentDidUpdate() {
     const { isFocused, tabIndex } = this.props;
     if (tabIndex === 0) {
-      if (isFocused || tabIndex !== prevProps.tabIndex) {
+      if (isFocused) {
```
backwardok: I think the change here is doing more than desired, and it's not super obvious that this behavior is being done specifically for month navigation clicks. Is it possible to hook into the actual click calls in https://github.com/airbnb/react-dates/blob/e8f8c3c86ce0d6885009068e98c51639d99ff323/src/components/DayPicker.jsx#L476-L479 to set some state information or something in order to validate in the previous if case so that it doesn't try to focus in those cases?

backwardok:
Related: https://github.com/airbnb/react-dates/issues/1484

erin-doyle:
Please see the PR I've posted here as an alternate approach to fixing this issue: https://github.com/airbnb/react-dates/pull/1954.

PR #1750, opened 2019-08-11
Fix DateInput cannot fully show Chinese issue

LeoShi:
**Before:**
Screen Shot 2019-08-11 at 7 13 55 PM

**After:**
Screen Shot 2019-08-11 at 7 14 48 PM

### DIFF
--- a/src/theme/DefaultTheme.js
+++ b/src/theme/DefaultTheme.js
@@ -172,7 +172,7 @@ export default {
     },
 
     sizing: {
-      inputWidth: 130,
+      inputWidth: 140,
       inputWidth_small: 97,
       arrowWidth: 24,
     },


LeoShi:
@ljharb @backwardok could you please help stamp?

ljharb:
Is this a change that’s needed for all locales? Or is this something that could be applied to a locale-specific react-with-styles theme?

LeoShi:
@ljharb I only tried on English and Chinese. it works. 
I don‘t think this value can be override by a locale-specific react-with-styles theme. Or could you help show me some pseudo code about how to change it?

ljharb:
It’s currently defined in a theme, so a replacement theme would be able to define any value for it that’s desired.

LeoShi:
@ljharb Define a new theme will bring more work to do which I think it's not necessary in this case.
What's your concern about the change in this PR? I don't see any of side affect from a width size change but it indeed fix the problem.

ljharb:
It could break layouts that depend on the current width. Changing every layout due to one language or locale is a lot more disruptive than using a theme.

PR #1758, opened 2019-08-15
Remove all direct imports of css in favor of injected prop

noratarano:
## Summary

Remove all direct imports of css in favor of the css injected prop. This is the preferred usage and is a pre-req to using contextual react-with-styles.

## Reviewers

@majapw @TaeKimJR @ahuth


ljharb:
Failing test seemed due to flakiness at UTC midnight; i reran.

PR #1787, opened 2019-09-17
Add a render function for customizable week header text

caseklim:
## Summary
Add a function `renderWeekHeaderText` (would `renderWeekHeaderElemement` be more appropriate?) that allows the week header text for a calendar month to be customized.

### Default
image

### With custom week header text
image

### DIFF
--- a/src/components/DayPicker.jsx
+++ b/src/components/DayPicker.jsx
@@ -88,6 +88,7 @@ const propTypes = forbidExtraProps({
   // month props
   renderMonthText: mutuallyExclusiveProps(PropTypes.func, 'renderMonthText', 'renderMonthElement'),
   renderMonthElement: mutuallyExclusiveProps(PropTypes.func, 'renderMonthText', 'renderMonthElement'),
+  renderWeekHeaderElement: PropTypes.func,
 
   // day props
   modifiers: PropTypes.objectOf(PropTypes.objectOf(ModifiersShape)),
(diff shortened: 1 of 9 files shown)


src/components/DayPicker.jsx
```diff
@@ -88,6 +88,7 @@ const propTypes = forbidExtraProps({
   // month props
   renderMonthText: mutuallyExclusiveProps(PropTypes.func, 'renderMonthText', 'renderMonthElement'),
   renderMonthElement: mutuallyExclusiveProps(PropTypes.func, 'renderMonthText', 'renderMonthElement'),
+  renderWeekHeaderText: PropTypes.func,
```
ahuth: I like your suggestion of `renderWeekHeaderElement`, since an element would be a valid return value.

src/components/DayPicker.jsx
```diff
@@ -149,6 +150,7 @@ export const defaultProps = {
   // month props
   renderMonthText: null,
   renderMonthElement: null,
+  renderWeekHeaderText: null,
```
ahuth: Not sure if this is better or worse, but did you consider making the default for this:

```
renderWeekHeaderText: day => <small>{day}</small>
```

?

That would collapse the number of conditionals from 2 (the ternary in the render prop and the default prop) to 1 (just the default prop).

But maybe it would be harder to see what the default is when scanning the component?

caseklim: ooh I like that idea! though I tend to see this pattern more where the function prop is optional and then some functionality is used in its place, rather than assigning that functionality as the default to the function prop—so I'm inclined to keep this as-is for consistency.

PR #1788, opened 2019-09-18
Add a render function for keyboard shortcuts panel

caseklim:
## Summary

Add a prop `renderKeyboardShortcutsPanel` that allows a custom panel for keyboard shortcuts to be rendered.

### Default
image

### renderKeyboardShortcutsPanel
image

### DIFF
--- a/examples/DayPickerRangeControllerWrapper.jsx
+++ b/examples/DayPickerRangeControllerWrapper.jsx
@@ -74,7 +74,7 @@ const defaultProps = {
   renderDayContents: null,
   minimumNights: 1,
   isDayBlocked: () => false,
-  isOutsideRange: (day) => !isInclusivelyAfterDay(day, moment()),
+  isOutsideRange: day => !isInclusivelyAfterDay(day, moment()),
   isDayHighlighted: () => false,
   enableOutsideDays: false,
 
@@ -145,13 +145,12 @@ class DayPickerRangeControllerWrapper extends React.Component {
 
     return (
       <div style={{ height: '100%' }}>
-        {showInputs
-          && (
+        {showInputs && (
           <div style={{ marginBottom: 16 }}>
             <input type="text" name="start date" value={startDateString} readOnly />
             <input type="text" name="end date" value={endDateString} readOnly />
           </div>
-          )}
+        )}
 
         <DayPickerRangeController
           {...props}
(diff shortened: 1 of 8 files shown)


examples/DayPickerRangeControllerWrapper.jsx
```diff
@@ -73,7 +74,7 @@ const defaultProps = {
   renderDayContents: null,
   minimumNights: 1,
   isDayBlocked: () => false,
-  isOutsideRange: day => !isInclusivelyAfterDay(day, moment()),
+  isOutsideRange: (day) => !isInclusivelyAfterDay(day, moment()),
```
ljharb:
```suggestion
  isOutsideRange: day => !isInclusivelyAfterDay(day, moment()),
```

examples/DayPickerRangeControllerWrapper.jsx
```diff
@@ -143,12 +145,13 @@ class DayPickerRangeControllerWrapper extends React.Component {
 
     return (
       <div style={{ height: '100%' }}>
-        {showInputs &&
+        {showInputs
+          && (
```
ljharb: personally i'd usually do this like `{showInputs && (`

caseklim: @ljharb fixed and added more tests :)

caseklim:
@ljharb how long should the coverage check take to run? 🤔

ljharb:
This seems to be the issue @noratarano and i were working on; i believe recent permissions changes caused this to be broken. I’m not sure how to fix it :-/

A short term workaround could be to disable coveralls as a required check; a longer term workaround could be to replace coveralls with another service; ideally we can just figure out how to fix coveralls.

caseklim:
@ljharb ahh that'd do it, any chance it's fixed now? coveralls finished running at least but now I'm seeing "Coverage decreased (-68.7%) to 11.246%" on this PR which seems odd

ljharb:
@caseklim yep, should be all set now! try a rebase, and if it's still claiming coverage has dropped then more tests need to be added.

caseklim:
@ljharb but a coverage drop of -68.7% to 11.246%? I added a test for my new render function, and it's complaining about files I didn't even touch. e.g.,

image

ljharb:
It might be related to the issues coveralls was having - ie, if master was updated while coveralls was down, it might not know about the new coverage.

@noratarano can merge through the failing coverage check, which i think is fine here.

noratarano:
@caseklim @ljharb Merging!

PR #1789, opened 2019-09-18
add .gitpod.yml

garyfeng:
adding .gitpod.yml configuration file to support GitPod in-browser editing and live previewing. Live editing and previewing tested to work in Chrome.


ljharb:
What is gitpod, and why would we want this config file?

PR #1793, opened 2019-09-19
Add minDate and maxDate props to DateRangePicker

v47:
Currently, `DateRangePicker` does not support `minDate` and `maxDate` props. This PR will propagate them down to `DayPickerRangeController`.

PR #1794, opened 2019-09-19
Add tests for minDate and maxDate props in DateRangePicker

v47:
Missing tests for minDate and maxDate props in DateRangePicker.

From #1793


alanhamlett:
We're not testing for presence of other props that are passed through, should we add tests for all passed through props?

v47:
@ljharb, I sorry for the absent tests in a first place.

v47:
@alanhamlett comunications between `DateRangePickerInputController` and `DayPickerRangeController` not realy tested too...

PR #1796, opened 2019-09-20
Fix for getWeekHeaders(), prevents it from changing state.currentMonth

v47:
Fix for a bug that prevents `getWeekHeaders()` from changing `this.state.currentMonth`.

A bug that was introduced in v21.1.0, it resets `startDate` to the last day in the `startDate` week, because`getWeekHeaders()` changing `this.state.currentMonth`.

dayPickerBug


alanhamlett:
Thanks for finding and fixing this!

markotom:
is there a plan to release this fix asap? Because I think it's critical

noratarano:
@markotom et al, this was released! See the changelog.

Issue #1795, opened 2019-09-20
Date is changed on SingleDatePicker when picker is opened again


terohotti:
**react-dates@21.1.0**

Can be reproduced in: https://airbnb.io/react-dates/?path=/story/singledatepicker-sdp--default

Select SingleDatePicker -> Default
Pick a date -> Click on date field to open picker again -> Date on input field is changed to the last day of the week

![image](https://user-images.githubusercontent.com/4482090/65321869-0408f880-dbae-11e9-829d-3ed990d1cb1a.png)

mmarkelov:
@terohotti will be fixed by #1796

PR #1798, opened 2019-09-23
unit test for getWeekHeaders()

v47:
Will prevent regressions like in #1795. 
Ensures that state stays immuted while executing `getWeekHeaders()`.

### DIFF
--- a/test/components/DayPicker_spec.jsx
+++ b/test/components/DayPicker_spec.jsx
@@ -3,6 +3,7 @@ import moment from 'moment/min/moment-with-locales';
 import { expect } from 'chai';
 import sinon from 'sinon-sandbox';
 import { mount, shallow } from 'enzyme';
+import { cloneDeep } from 'lodash';
 
 import * as isDayVisible from '../../src/utils/isDayVisible';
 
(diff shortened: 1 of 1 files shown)


test/components/DayPicker_spec.jsx
```diff
@@ -3,6 +3,7 @@ import moment from 'moment/min/moment-with-locales';
 import { expect } from 'chai';
 import sinon from 'sinon-sandbox';
 import { mount, shallow } from 'enzyme';
+import _ from 'lodash';
```
```suggestion
import { cloneDeep } from 'lodash';
```

test/components/DayPicker_spec.jsx
```diff
@@ -832,6 +833,17 @@ describe('DayPicker', () => {
     });
   });
 
+  describe('#getWeekHeaders', () => {
+    it('returns unmuted weekday headers for currentMonth in a future', () => {
```
```suggestion
    it('returns unmutated weekday headers for currentMonth in a future', () => {
```

test/components/DayPicker_spec.jsx
```diff
@@ -832,6 +833,17 @@ describe('DayPicker', () => {
...
+      const INITIAL_MONTH = moment().add(2, 'Months').week(3).weekday(3);
+      const dayPicker = PureDayPicker.prototype.constructor(PureDayPicker.defaultProps);
```
ljharb: constructors can't be called; you may want `new PureDayPicker()` here?

test/components/DayPicker_spec.jsx
```diff
@@ -832,6 +833,17 @@ describe('DayPicker', () => {
...
+      dayPicker.state.currentMonth = moment(INITIAL_MONTH);
```
ljharb: it'd be better to use an enzyme wrapper here, rather than manually constructing an instance.

test/components/DayPicker_spec.jsx
```diff
@@ -832,6 +833,17 @@ describe('DayPicker', () => {
...
   });
 
+  describe('#getWeekHeaders', () => {
+    it('returns unmutated weekday headers for currentMonth in a future', () => {
+      const INITIAL_MONTH = moment().add(2, 'Months').week(3).weekday(3);
+      var dayPicker = new PureDayPicker(PureDayPicker.defaultProps);
+      dayPicker.state.currentMonth = moment(INITIAL_MONTH);
+      const state = cloneDeep(dayPicker.state);
+      expect(dayPicker.getWeekHeaders()).to.be.eql(INITIAL_MONTH.localeData().weekdaysMin());
+      expect(dayPicker.state).to.eql(state);
```
ljharb: if you also want to ensure they're not mutated, add this:
```suggestion
      expect(dayPicker.state).not.to.equal(state);
      expect(dayPicker.state).to.eql(state);
```

test/components/DayPicker_spec.jsx
v47: I wanted to prevent render to call `getWeekHeaders()`, because it was changing state already on that stage.

ljharb: `sinon.spy(PureDayPicker.prototype, 'render', () => null)`?

v47:
@ljharb Thanks! Feels like you are always watching :)

PR #1831, opened 2019-10-07
Force render prop to fix calendar rendering issue in Safari

### LINKED ISSUE #1829: Days are not rendered in Safari browsers
**react-dates version**
Latest

**Describe the bug**
I am using DayPickerRangeController and every time this component is rendered outside container (overflow: auto or overflow: scroll) calendar days are not rendered (see foto). So far this issue was reported only by users who are using Safari browsers.

 - OS: iOS, Windows
 - Browser safari

**Smartphone (please complete the following information):**
 - Device: iPad
 - OS: iOS latest
 - Browser safari

71546701_3044072915663040_3078288900259577856_n

povilasDadelo:
Closes #1829 issue

### DIFF
--- a/src/components/DayPicker.jsx
+++ b/src/components/DayPicker.jsx
@@ -54,7 +54,9 @@ const YEAR_SELECTION_TRANSITION = 'year_selection';
 
 const propTypes = forbidExtraProps({
   ...withStylesPropTypes,
-
+  // forked additional props
+  forceRender: PropTypes.bool,
+  
   // calendar presentation props
   enableOutsideDays: PropTypes.bool,
   numberOfMonths: PropTypes.number,
@@ -120,6 +122,7 @@ const propTypes = forbidExtraProps({
 });
 
 export const defaultProps = {
+  forceRender: false,
   // calendar presentation props
   enableOutsideDays: false,
   numberOfMonths: 2,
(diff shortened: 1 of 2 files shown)


src/components/DayPicker.jsx
```diff
@@ -54,7 +54,9 @@ const YEAR_SELECTION_TRANSITION = 'year_selection';
 
 const propTypes = forbidExtraProps({
   ...withStylesPropTypes,
-
+  // forked additional props
+  forceRender: PropTypes.bool,
```
ljharb: I’m confused why this is needed - can you not force a render by passing a different object down?

src/components/DayPicker.jsx
```diff
@@ -361,6 +364,10 @@ class DayPicker extends React.PureComponent {
     if (!prevProps.isFocused && isFocused && !focusedDate) {
       this.container.focus();
     }
+    
+    if (prevProps.forceRender !== forceRender) {
```
ljharb: this means that going from true to false also triggers the logic

src/components/DayPickerRangeController.jsx
```diff
@@ -1174,6 +1176,7 @@ export default class DayPickerRangeController extends React.PureComponent {
       transitionDuration,
       verticalBorderSpacing,
       horizontalMonthPadding,
+      forceRender
```
ljharb: this prop isn’t forcing this component to rerender, so it seems confusing to only pass it down.

src/components/DayPicker.jsx
povilasDadelo: Yes this prop is passed so that we could trigger the rerender of DatePicker component by calling onMonthChange. For example when a calendar is drawn somewhere in scrollable modal the days are not rendered, so when you scroll to DatePicker it is empty.

PR #1834, opened 2019-10-11
Call getStateForNewMonth when date/startDate/endDate is set to a date that is not visible

caseklim:
If `date`, `startDate`, or `endDate` is set externally by a different date input, this will bring the month with the newly selected date into view.

### DIFF
--- a/src/components/DayPickerRangeController.jsx
+++ b/src/components/DayPickerRangeController.jsx
@@ -18,6 +18,7 @@ import isBeforeDay from '../utils/isBeforeDay';
 import getVisibleDays from '../utils/getVisibleDays';
 import isDayVisible from '../utils/isDayVisible';
 
+import getPooledMoment from '../utils/getPooledMoment';
 import getSelectedDateOffset from '../utils/getSelectedDateOffset';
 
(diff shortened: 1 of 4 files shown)


src/components/DayPickerRangeController.jsx
```diff
@@ -321,10 +335,10 @@ export default class DayPickerRangeController extends React.PureComponent {
         && !prevFocusedInput
         && didFocusChange
       )
+      || !isDateVisible
     ) {
       const newMonthState = this.getStateForNewMonth(nextProps);
-      const { currentMonth } = newMonthState;
-      ({ visibleDays } = newMonthState);
+      ({ currentMonth, visibleDays } = newMonthState);
```
noratarano: Not real feedback: Why does this need to be wrapped in parens like this? 🤔

caseklim: @noratarano I thiiiink it’s to destructure into a `let` assignment—this is also the first time I’ve seen this :)

majapw: yup you are right!

PR #1835, opened 2019-10-12
Turn off react/deprecated lint rule and remove erroneous `.only`

majapw:
The lint rule was likely introduced by a downstream dependency. Currently turning off the rule to unblock development.

Also removing an erroneous `.only`.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -27,6 +27,8 @@
     "no-restricted-imports": 0, // TODO: enable with full RTL support
 
     "react/jsx-props-no-spreading": 0, // TODO: re-evaluate
+
+    "react/no-deprecated": 1, // TODO: update to UNSAFE_componentWillReceiveProps
   },
 
   "overrides": [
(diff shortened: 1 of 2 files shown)


ljharb:
looks like master

majapw:
Looks like the lint failure is on master, likely due to an eslint plugin minor bump I bet. I'll turn that rule off for now.

In addition, a `.only` snuck into a test which unfortunately led to https://github.com/airbnb/react-dates/pull/1798 passing when I don't think it should have. I'm going to revert that change for now and merge this one in to unblock @caseklim's PR

.eslintrc
```diff
@@ -27,6 +27,8 @@
     "no-restricted-imports": 0, // TODO: enable with full RTL support
 
     "react/jsx-props-no-spreading": 0, // TODO: re-evaluate
+
+    "react/no-deprecated": 1, // TODO: update to UNSAFE_componentWillReceiveProps
```
ljharb: note that this TODO will be a semver-major, hoisting support to 16.3+

test/components/DayPickerRangeController_spec.jsx
```diff
@@ -4492,7 +4492,7 @@ describe('DayPickerRangeController', () => {
       });
     });
 
-    describe.only('renderKeyboardShortcutsPanel prop', () => {
```
ljharb: eesh, good catch

PR #1837, opened 2019-10-12
Revert "Revert "[Tests] `DayPicker`: unit test for getWeekHeaders()""

majapw:
This reverts commit defb7fd680cab4154d12e9416c9f28e4a47e9e5b.

@v47 @ljharb Unfortunately, due to an accidental `describe.only` that slipped into master, we weren't running the full test suite on https://github.com/airbnb/react-dates/pull/1798. It seems like the test actually fails on React 15. If you'd like to take another look, we can try to get it merged in.

Thanks!

### DIFF
--- a/test/components/DayPicker_spec.jsx
+++ b/test/components/DayPicker_spec.jsx
@@ -3,6 +3,7 @@ import moment from 'moment/min/moment-with-locales';
 import { expect } from 'chai';
 import sinon from 'sinon-sandbox';
 import { mount, shallow } from 'enzyme';
+import { cloneDeep } from 'lodash';
 
 import * as isDayVisible from '../../src/utils/isDayVisible';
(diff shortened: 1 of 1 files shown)


test/components/DayPicker_spec.jsx
```diff
@@ -871,6 +872,23 @@ describe('DayPicker', () => {
     });
   });
 
+  describe('#getWeekHeaders', () => {
+    it('returns unmutated weekday headers for currentMonth in a future', () => {
+      sinon.stub(PureDayPicker.prototype, 'render');
```
```suggestion
      sinon.stub(PureDayPicker.prototype, 'render', () => null);
```
I think the issue is that this stub is causing it to render `undefined`, which isn't valid.

v47:
@majapw Will look on it today or tomorrow.

Issue #1737, opened 2019-07-30
Upgrade Lodash, security


LuisSaybe:
**react-dates version**
react-dates@20.2.5

**Describe the bug**
lodash has a security vulnerability described here https://www.npmjs.com/advisories/1065

Can you upgrade to latest lodash to fix this?

```
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ high          │ Prototype Pollution                                          │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ lodash                                                       │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Patched in    │ >=4.17.12                                                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ react-dates                                                  │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ react-dates > lodash                                         │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://www.npmjs.com/advisories/1065                        │
└───────────────┴──────────────────────────────────────────────────────────────┘
```

ljharb:
This is both a false positive, and no action is necessary since we use a semver range. If you’re not using a lockfile, reinstalling react-dates should update lodash, and if you are, you can manually update your transitive deps with npm audit fix.

josgraha:
what about the transitive deps from the old `babel-X` packages?

ljharb:
They're only dev deps, and also irrelevant to the CVE (most CVEs are false positives).

PR #1838, opened 2019-10-15
[Security] Bump lodash to 4.17.15

josgraha:
With the carat dependency installing should bump `lodash` to latest but it seemed best to be explicit to address the following [Snyk Issue](https://www.npmjs.com/advisories/1065)


ljharb:
See #1737.

PR #1839, opened 2019-10-15
[Security] bump all deps with `lodash@4.17.12` or lower

josgraha:
This is a more aggressive PR bumping all versions that depend on `lodash@4.17.12` or lower.  Unfortunately this required bumping `babel-X` packages to `@babel/` and required config changes.


ljharb:
See #1838 as well as #1737.

There's no security issue in *this* package wrt use of older lodash.

ljharb:
Also, updating to babel 7 will fail tests in older node versions that remain supported by this package.

PR #1851, opened 2019-10-30
Revert "Call getStateForNewMonth when date/startDate/endDate is set to a date that is not visible"

caseklim:
Reverts airbnb/react-dates#1834


noratarano:
@caseklim Can you add a label for what semver this should be?

PR #1854, opened 2019-11-04
Update flakey test for no-selected-start-before-selected-end

krissalvador27:
**What's this fix?**
Fixes the breaking test in master from merged in PR #1608 

**What happened?**
The broken test was validating that a change in end date would call `deleteModifier` for `no-selected-start-before-selected-end` on all visible days. Turns out the test was flakey because the visible days object calculated by DayPickerRangeController is determined by the `initialVisibleMonthThunk`.
```
const initialVisibleMonthThunk = initialVisibleMonth || (
   startDate ? () => startDate : () => this.today
);
``` 
[DayPickerRangeController.jsx Line 1065](https://github.com/airbnb/react-dates/blob/master/src/components/DayPickerRangeController.jsx#L1065). 

The default `initialVisibleMonthThunk` does not take into account a passed in end date and by defaulting to today, the test cannot rely on a hardcoded number of times that `deleteModifier `will run. This isn't a major issue since DayPickerRangeController is most likely consumed by using the DateRangePicker kitchen sink that determines `initialVisibleMonthThunk` like so
``` 
const initialVisibleMonthThunk = initialVisibleMonth || (
    () => (startDate || endDate || moment())
);
```
[DayPickerRangeController.jsx Line 460](https://github.com/airbnb/react-dates/blob/master/src/components/DateRangePicker.jsx#L460). 

Happy to update the default of DayPickerRangeController to match the default on the DateRangePicker for both consistency and expected behavior, but wanted to prioritize fixing the flakey test for now.

Let me know if this works @majapw @ljharb! 😁

### DIFF
--- a/test/components/DayPickerRangeController_spec.jsx
+++ b/test/components/DayPickerRangeController_spec.jsx
@@ -1622,10 +1622,11 @@ describe('DayPickerRangeController', () => {
               />,
             );
             const newStartDate = today;
-            const visibleDays = wrapper.instance().state.visibleDays;
-            const numberOfVisibleDays = Object.values(visibleDays).reduce((total, visibleDayArray) => {
-              return total + Object.keys(visibleDayArray).length;
-            }, 0);
+            const { visibleDays } = wrapper.instance().state;
+            const numberOfVisibleDays = Object.values(visibleDays).reduce(
+              (total, visibleDayArray) => total + Object.keys(visibleDayArray).length,
+              0,
+            );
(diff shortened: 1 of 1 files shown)


krissalvador27:
@ljharb Would we be able to merge and publish a new minor? I feel some slight nervousness for being the pull request that is breaking master right now 😅

ljharb:
No need to publish anything just for tests; I’m going to defer to @majapw to merge this tho.

PR #1857, opened 2019-11-06
Relaxing propTypes for CalendarWeek

kevinthepan:
Since Airbnb's own monorepo uses the wrapper `PdpCalendarDay` around `CustomizableCalendarDay`, it triggers a `TypeError` every time the component tree is mounted in any sort of unit test.

Relaxing the propTypes for CalendarWeek to accept any sort of node.


majapw:
The existing failure is unrelated to this change. @krissalvador27 is looking into it.

krissalvador27:
Hey @majapw, I have a fix open here #1854 😁

kevinthepan:
@ljharb We have a bunch of business logic surrounding CustomizableCalendarDay (day-specific styling, day contents styling, day dependent phrases, etc) that makes things a lot cleaner and more manageable when encapsulated as part of the day instead of at the date picker level.

majapw:
Thanks @krissalvador27! 

And yeah, agree with @kevinthepan's statement. I think we are more generally interested in transforming this library into a more flexible solution so as to better accommodate needs both internally and externally.

PR #1859, opened 2019-11-13
Allow for custom month navigation buttons

caseklim:
[Image from Gyazo](https://gyazo.com/80426812d12022a60639160a84e68686)

image

### DIFF
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -103,6 +103,8 @@ const defaultProps = {
   navPosition: NAV_POSITION_TOP,
   navPrev: null,
   navNext: null,
+  renderNavPrevButton: null,
+  renderNavNextButton: null,
 
   onPrevMonthClick() {},
   onNextMonthClick() {},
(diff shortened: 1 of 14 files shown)


src/components/DateRangePicker.jsx
```diff
@@ -469,8 +473,10 @@ class DateRangePicker extends React.PureComponent {
 
     const withAnyPortal = withPortal || withFullScreenPortal;
 
+    /* eslint-disable jsx-a11y/no-static-element-interactions */
```
kevinthepan: Is this not doable in the same inline comment?

caseklim: oooh TIL https://stackoverflow.com/a/56714489

src/components/DateRangePicker.jsx
```diff
@@ -469,8 +473,9 @@ class DateRangePicker extends React.PureComponent {
 
     const withAnyPortal = withPortal || withFullScreenPortal;
 
+    /* eslint-disable jsx-a11y/no-static-element-interactions, jsx-a11y/click-events-have-key-events */
```
kevinthepan: Can we move it down to where it originally was and use `eslint-disable-line`

caseklim: done 👌 I was trying to stay within the character limit but the line length is too long either way lol

PR #1869, opened 2019-11-28
Fix invalid font-weight for regular option

### LINKED ISSUE #1868: Invalid font-weight style for regular option
When `regular` option is true, style for `DateInput_input__regular` has `font-weight: auto`. 

But `auto` is an invalid value for `font-weight` in Chrome (I have not checked other browsers). The value should be `unset`.

mmarkelov:
Fix #1868 
Set it to **inherit** to support IE browsers

### DIFF
--- a/src/components/DateInput.jsx
+++ b/src/components/DateInput.jsx
@@ -335,7 +335,7 @@ export default withStyles(({
   },
 
   DateInput_input__regular: {
-    fontWeight: 'auto',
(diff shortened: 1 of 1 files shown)


ljharb:
Unfortunately the original fork this PR was issued from was deleted, so this PR is unrecoverable. I will cherry-pick the commit onto the default branch to close it.

PR #1878, opened 2019-12-05
Fix keyboard accessibility for minDate and maxDate

### LINKED ISSUE #1860: minDate and maxDate do not work if you navigate with keyboard shortcut
**Describe the bug**
When you set up a minDate or a maxDate you suppose to not be able to navigate before or after the month you set up. However if you use the keyboard shortcut to navigate it seems to not care about minDate or maxDate and we can continue navigate even if it should be block.

**To Reproduce**
1. Go to https://airbnb.io/react-dates/?path=/story/daterangepicker-drp--with-navigation-blocked-mindate-and-maxdate
2. Click to open the DatePicker
3. Use any arrow key to navigate (arrow up to navigate by week) navigate until september. The min date setup is 1st of september so you should not be able to keep navigate back. 
4. You can keep navigate forever.

**Expected behavior**
You should not be able to continue navigate with arrow if you try to navigate before or after the minDate , maxDate

mmarkelov:
Fix #1860

Sorry for multiple pr about this fix, I got some troubles with test configuration on my machine ;(


erin-doyle:
This looks like a worthwhile fix to me!  @mmarkelov can you resolve the conflicts shown here?  I wonder if we can then get this reviewed and merged.

mmarkelov:
@erin-doyle youp! Done here

src/components/DayPicker.jsx
```diff
@@ -96,6 +98,9 @@ const propTypes = forbidExtraProps({
   onNextMonthClick: PropTypes.func,
   onMonthChange: PropTypes.func,
   onYearChange: PropTypes.func,
+  onMultiplyScrollableMonths: PropTypes.func, // VERTICAL_SCROLLABLE daypickers only
```
erin-doyle: I believe since you started this PR `onMultiplyScrollableMonths` has been removed (https://github.com/airbnb/react-dates/commit/a3e76feb193d09af640e226bf33f7d98f628cad8#diff-87325f2795c2738c7a5b679475232f8e).  This doesn't look like it's being used anywhere anymore so you should probably remove it.

PR #1884, opened 2019-12-20
Add VERTICAL_SCROLLABLE as an orientation for DateRangePicker #898

bobrobcpp:
Previously functional but leaves a prop type warning in the console


bobrobcpp:
Yeah, planning to add these tests and stories when I can get the time

PR #1890, opened 2019-12-31
Fix SingleDatePicker when appendToBody is enabled.

### LINKED ISSUE #1550: appendToBody property breaks date selection in SingleDatePicker when in a Material-UI Dialog
**react-dates version**

- react-dates@19.0.3

**Describe the bug**

When `appendToBody` is set in a `SingleDatePicker` and the date picker is in a Material-UI dialog, the picker displays correctly on click but `onDateChange` callback is never called and the picker gets closed immediately. The date doesn't update.

As I said in #1535 , #1540 has fixed the most the obvious cases (thanks again @danedavid for looking into this before I had a chance), yet some edge cases like this one still remain unfortunately.

I was able to reproduce the issue by wrapping the picker in a Material-UI Dialog component:

```jsx
class DP extends Component {
  constructor(props) {
    super(props);
    this.state = {
      focused: props.autoFocus,
      date: props.initialDate,
    };

    this.onDateChange = this.onDateChange.bind(this);
  }

  onDateChange(date) {
    this.setState({ date });
  }

  onFocusChange({ focused }) {
    this.setState({ focused });
  }

  render() {
   const { focused, date } = this.state; 
    return (
        <SingleDatePicker
          id="mydatepicker"
          date={date}
          focused={focused}
          onDateChange={this.onDateChange}
          appendToBody
        />
    );
  }
}

class App extends Component {
  render() {
    return (
      <Dialog open>
        <DialogContent>
          <DP />
        </DialogContent>
      </Dialog>
    );
  }
}
```

See repro on CodeSandbox [here](https://codesandbox.io/s/nl9q7qzyj).

Would probably need some further investigations to understand why Material-UI's Dialogs produce such behaviors on React-Dates.

**Screenshots/Gifs**

None.

* OS: Archlinux
* Browser: Chromium, Firefox
* Version: 72, 65

Not sure.

gilbsgilbs:
Fixes #1550

**Disclaimer:** this fix is probably very naive and I make it for the sole purpose of bringing attention of contributors and maintainers on this bug that has been around for way too long now. Given that neither of:
- the very long comment
- the git commit history
- the tests
- my very dysfunctional brain

are of any help at understanding why `relatedTarget` was looked for in the first place, and given that it introduces a bug in the `appendToBody={true}` case, this is the most logical PR I could come up with.


gilbsgilbs:
By the way, a less controversial fix might be what I suggested [here](https://github.com/airbnb/react-dates/issues/1550#issuecomment-464922437) about a year ago.

viniciuscdavila:
This is happening in my project as well

ljharb (changes requested):
I am very concerned about removing what the comment implies is a workaround for IE < 11, or other older browsers.

PR #1893, opened 2020-01-09
[Fix] `DateRangePickerInputController`: fix onClose method

### LINKED ISSUE #1892: onClose returns incorrect endDate if it's been entered manually in the input
If dates are not choosed from calendar but entered manually in inputs onClose handler returns null or previous date in endDate. It is visible in storybook https://airbnb.io/react-dates/?path=/story/drp-calendar-props--with-onclose-handler

mmarkelov:
Fixes #1892 
I suppose following issue because of firing **onClearFocus** with start and end dates from props, that's not right it should use values inside **onEndDateChange** function

### DIFF
--- a/test/components/DateRangePickerInputController_spec.jsx
+++ b/test/components/DateRangePickerInputController_spec.jsx
@@ -221,6 +221,23 @@ describe('DateRangePickerInputController', () => {
           expect(isSameDay(onDatesChangeArgs.endDate, futureDate)).to.equal(true);
         });
 
+        it('calls props.onClose with props.startDate and provided end date', () => {
+          const onCloseStub = sinon.stub();
+          const wrapper = shallow((
+            <DateRangePickerInputController
+              onClose={onCloseStub}
+              startDate={startDate}
+            />
+          ));
+          wrapper.instance().onEndDateChange(validFutureDateString);
+          expect(onCloseStub).to.have.property('callCount', 1);
+
+          const [onCloseArgs] = onCloseStub.getCall(0).args;
+          const futureDate = moment(validFutureDateString);
+          expect(onCloseArgs).to.have.property('startDate', startDate);
+          expect(isSameDay(onCloseArgs.endDate, futureDate)).to.equal(true);
+        });
+
         describe('props.onFocusChange', () => {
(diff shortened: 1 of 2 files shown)


test/components/DateRangePickerInputController_spec.jsx
```diff
@@ -222,6 +222,23 @@ describe('DateRangePickerInputController', () => {
...
+        it('calls props.onClose with props.startDate and provided end date', () => {
+          const onCloseStub = sinon.stub();
+          const wrapper = shallow((
+            <DateRangePickerInputController
+              onClose={onCloseStub}
+              startDate={startDate}
+            />
+          ));
+          wrapper.instance().onEndDateChange(validFutureDateString);
+          expect(onCloseStub.callCount).to.equal(1);
```
ljharb:
```suggestion
          expect(onCloseStub).to.have.property('callCount', 1);
```

test/components/DateRangePickerInputController_spec.jsx
```diff
@@ -222,6 +222,23 @@ describe('DateRangePickerInputController', () => {
...
+
+          const onCloseArgs = onCloseStub.getCall(0).args[0];
```
ljharb:
```suggestion
          const [onCloseArgs] = onCloseStub.getCall(0).args;
```

test/components/DateRangePickerInputController_spec.jsx
```diff
@@ -222,6 +222,23 @@ describe('DateRangePickerInputController', () => {
...
+          const futureDate = moment(validFutureDateString);
+          expect(onCloseArgs.startDate).to.equal(startDate);
```
ljharb:
```suggestion
          expect(onCloseArgs).to.have.property('startDate', startDate);
```

tjhart:
If I understand correctly, these changes will fix #1892 ? It's causing issues for me as well. What's left to get a release out that includes this?

tjhart:
@ljharb Leaving another ping about this. I see that the changes have been approved, and I haven't found that another issue superseding this one. Can I do anything to help get this into a release?

PR #1894, opened 2020-01-10
Add functionality to see previous months for vertical scrollable calendar

nkinser:
This PR adds the ability to view months before the current month when the calendar is vertically scrollable. There's a small issue where when you click to view more dates above the calendar shifts down and you see the farthest date in the past rather than staying on the month you were just viewing. I will try to fix this in a followup PR.

#### Before
react-dates-scroll-before

#### After
react-dates-scroll-after

### DIFF
--- a/src/components/DayPicker.jsx
+++ b/src/components/DayPicker.jsx
@@ -679,6 +679,29 @@ class DayPicker extends React.PureComponent {
     this.transitionContainer = ref;
   }
 
+  getNextScrollableMonths(e) {
+    const { onGetNextScrollableMonths } = this.props;
+    if (e) e.preventDefault();
+
+    if (onGetNextScrollableMonths) onGetNextScrollableMonths(e);
+
+    this.setState(({ scrollableMonthMultiple }) => ({
+      scrollableMonthMultiple: scrollableMonthMultiple + 1,
+    }));
+  }
+
+  getPrevScrollableMonths(e) {
+    const { numberOfMonths, onGetPrevScrollableMonths } = this.props;
+    if (e) e.preventDefault();
+
+    if (onGetPrevScrollableMonths) onGetPrevScrollableMonths(e);
+
+    this.setState(({ currentMonth, scrollableMonthMultiple }) => ({
+      currentMonth: currentMonth.clone().subtract(numberOfMonths, 'month'),
+      scrollableMonthMultiple: scrollableMonthMultiple + 1,
+    }));
+  }
+
   maybeTransitionNextMonth(newFocusedDate) {
     const { numberOfMonths } = this.props;
     const { currentMonth, focusedDate } = this.state;
@@ -707,29 +730,6 @@ class DayPicker extends React.PureComponent {
     }
 
     return false;
-  }
-
-  getNextScrollableMonths(e) {
-    const { onGetNextScrollableMonths } = this.props;
-    if (e) e.preventDefault();
-
-    if (onGetNextScrollableMonths) onGetNextScrollableMonths(e);
-
-    this.setState(({ scrollableMonthMultiple }) => ({
-      scrollableMonthMultiple: scrollableMonthMultiple + 1,
-    }));
-  }
-
-  getPrevScrollableMonths(e) {
-    const { numberOfMonths, onGetPrevScrollableMonths } = this.props;
-    if (e) e.preventDefault();
-
-    if (onGetPrevScrollableMonths) onGetPrevScrollableMonths(e);
-
-    this.setState(({ currentMonth, scrollableMonthMultiple }) => ({
-      currentMonth: currentMonth.clone().subtract(numberOfMonths, 'month'),
-      scrollableMonthMultiple: scrollableMonthMultiple + 1,
-    }));
   }
 
   isHorizontal() {
@@ -879,15 +879,13 @@ class DayPicker extends React.PureComponent {
       return null;
     }
 
-    const onPrevMonthClick =
-      orientation === VERTICAL_SCROLLABLE
-        ? this.getPrevScrollableMonths
-        : this.onPrevMonthClick;
+    const onPrevMonthClick = orientation === VERTICAL_SCROLLABLE
+      ? this.getPrevScrollableMonths
+      : this.onPrevMonthClick;
 
-    const onNextMonthClick =
-      orientation === VERTICAL_SCROLLABLE
-        ? this.getNextScrollableMonths
-        : this.onNextMonthClick;
+    const onNextMonthClick = orientation === VERTICAL_SCROLLABLE
+      ? this.getNextScrollableMonths
+      : this.onNextMonthClick;
 
     return (
       <DayPickerNavigation
--- a/src/components/DayPicker.jsx
+++ b/src/components/DayPicker.jsx
@@ -51,6 +51,8 @@ const PREV_TRANSITION = 'prev';
 const NEXT_TRANSITION = 'next';
 const MONTH_SELECTION_TRANSITION = 'month_selection';
 const YEAR_SELECTION_TRANSITION = 'year_selection';
+const PREV_NAV = 'prev_nav';
+const NEXT_NAV = 'next_nav';
 
 const propTypes = forbidExtraProps({
   ...withStylesPropTypes,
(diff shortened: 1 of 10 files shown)


src/components/DayPicker.jsx
```diff
@@ -842,7 +859,7 @@ class DayPicker extends React.PureComponent {
     });
   }
 
-  renderNavigation() {
+  renderNavigation(navDirection = null) {
```
ljharb: curious, why default it to null? this only has the effect of making `undefined` always be `null`, but it's not clear to me why that would matter.

src/components/DayPickerNavigation.jsx
```diff
@@ -151,19 +157,21 @@ function DayPickerNavigation({
...
-        }) : (
+      {showNavPrevButton &&
+        (renderNavPrevButton ? (
+          renderNavPrevButton({
+            ariaLabel: phrases.jumpToPrevMonth,
+            disabled: disablePrev,
+            onClick: disablePrev ? undefined : onPrevMonthClick,
+            onKeyUp: disablePrev ? undefined : (e) => {
+              const { key } = e;
+              if (key === 'Enter' || key === ' ') onPrevMonthClick(e);
```
ljharb:
```suggestion
              if (key === 'Enter' || key === ' ') {
                onPrevMonthClick(e);
              }
```

src/components/DayPickerNavigation.jsx
```diff
@@ -151,19 +157,21 @@ function DayPickerNavigation({
...
-          },
-        }) : (
+      {showNavPrevButton &&
+        (renderNavPrevButton ? (
+          renderNavPrevButton({
+            ariaLabel: phrases.jumpToPrevMonth,
+            disabled: disablePrev,
+            onClick: disablePrev ? undefined : onPrevMonthClick,
+            onKeyUp: disablePrev ? undefined : (e) => {
+              const { key } = e;
```
ljharb:
```suggestion
```

src/components/DayPickerNavigation.jsx
```diff
@@ -151,19 +157,21 @@ function DayPickerNavigation({
...
-            e.currentTarget.blur();
-          },
-        }) : (
+      {showNavPrevButton &&
+        (renderNavPrevButton ? (
+          renderNavPrevButton({
+            ariaLabel: phrases.jumpToPrevMonth,
+            disabled: disablePrev,
+            onClick: disablePrev ? undefined : onPrevMonthClick,
+            onKeyUp: disablePrev ? undefined : (e) => {
```
ljharb:
```suggestion
            onKeyUp: disablePrev ? undefined : ({ key }) => {
```

src/components/DayPickerNavigation.jsx
```diff
@@ -201,62 +210,63 @@ function DayPickerNavigation({
...
+        ))}
+
+      {showNavNextButton &&
+        (renderNavNextButton ? (
+          renderNavNextButton({
+            ariaLabel: phrases.jumpToNextMonth,
+            disabled: disableNext,
+            onClick: disableNext ? undefined : onNextMonthClick,
+            onKeyUp: disableNext ? undefined : (e) => {
+              const { key } = e;
```
ljharb:
```suggestion
```

src/components/DayPickerNavigation.jsx
```diff
@@ -201,62 +210,63 @@ function DayPickerNavigation({
...
-                !isRTL && styles.DayPickerNavigation_rightButton__horizontalDefault,
+        ))}
+
+      {showNavNextButton &&
+        (renderNavNextButton ? (
+          renderNavNextButton({
+            ariaLabel: phrases.jumpToNextMonth,
+            disabled: disableNext,
+            onClick: disableNext ? undefined : onNextMonthClick,
+            onKeyUp: disableNext ? undefined : (e) => {
```
ljharb:
```suggestion
            onKeyUp: disableNext ? undefined : ({ key }) => {
```

src/components/DayPickerNavigation.jsx
```diff
@@ -201,62 +210,63 @@ function DayPickerNavigation({
...
+              const { key } = e;
+              if (key === 'Enter' || key === ' ') onNextMonthClick(e);
```
ljharb:
```suggestion
              if (key === 'Enter' || key === ' ') {
                onNextMonthClick(e);
              }
```

src/components/DayPickerNavigation.jsx
```diff
@@ -201,62 +210,63 @@ function DayPickerNavigation({
...
+            },
+            onMouseUp: disableNext ? undefined : e => {
```
ljharb:
```suggestion
            onMouseUp: disableNext ? undefined : (e) => {
```

src/components/DayPickerNavigation.jsx
```diff
@@ -201,62 +210,63 @@ function DayPickerNavigation({
...
-          {navNextIcon}
-        </div>
-      )}
+            )}
+            aria-disabled={disableNext ? true : undefined}
+            aria-label={phrases.jumpToNextMonth}
+            onClick={disableNext ? undefined : onNextMonthClick}
+            onKeyUp={ disableNext ? undefined : (e) => {
+              const { key } = e;
+              if (key === 'Enter' || key === ' ') onNextMonthClick(e);
```
ljharb:
```suggestion
              if (key === 'Enter' || key === ' ') {
                onNextMonthClick(e);
              }
```

src/components/DayPickerNavigation.jsx
```diff
@@ -201,62 +210,63 @@ function DayPickerNavigation({
...
-        >
-          {navNextIcon}
-        </div>
-      )}
+            )}
+            aria-disabled={disableNext ? true : undefined}
+            aria-label={phrases.jumpToNextMonth}
+            onClick={disableNext ? undefined : onNextMonthClick}
+            onKeyUp={ disableNext ? undefined : (e) => {
+              const { key } = e;
```
ljharb:
```suggestion
```

src/components/DayPickerNavigation.jsx
```diff
@@ -201,62 +210,63 @@ function DayPickerNavigation({
...
-          }}
-        >
-          {navNextIcon}
-        </div>
-      )}
+            )}
+            aria-disabled={disableNext ? true : undefined}
+            aria-label={phrases.jumpToNextMonth}
+            onClick={disableNext ? undefined : onNextMonthClick}
+            onKeyUp={ disableNext ? undefined : (e) => {
```
ljharb:
```suggestion
            onKeyUp={ disableNext ? undefined : ({ key }) => {
```

ljharb:
What makes this a breaking change exactly? It seems like a semver-minor but I might be missing something.

nkinser:
> What makes this a breaking change exactly? It seems like a semver-minor but I might be missing something.

Actually, I think you're right. I was thinking it was incompatible because it could mess up a user's calendar navigation, but I guess that's not really incompatibility

nkinser:
@ljharb are there any additional stories you think would be helpful? Since this change doesn't rely on any props from the user I thought change to the existing vertical scrollable stories was enough

ljharb:
oh if it's already accessible in existing stories, then there's def no point in adding more :-D i just saw no story files in the diff :-p

src/components/DayPickerRangeController.jsx
```diff
@@ -1005,6 +1006,22 @@ export default class DayPickerRangeController extends React.PureComponent {
...
 
+  onGetPrevScrollableMonths() {
+    const { numberOfMonths, enableOutsideDays } = this.props;
+    const { currentMonth, visibleDays } = this.state;
+
+    const firstPreviousMonth = currentMonth.clone().subtract(numberOfMonths, 'month');
+    const newVisibleDays = getVisibleDays(firstPreviousMonth, numberOfMonths, enableOutsideDays, true);
+
+    this.setState({
+      currentMonth: firstPreviousMonth.clone(),
```
kevinthepan: @nkinser Is this why the calendar resets to the farthest date in the past? You're setting the `currentMonth` to the `firstPreviousMonth`

nkinser: @kevinthepan no, `currentMonth` has to be set to `firstPreviousMonth` in order for the previous months to be rendered. I think the reason it shows the earliest month is because when the past months are rendered they push the rest of the months down. I'm not sure how to get around that though

src/components/DayPicker.jsx
nkinser: you're right, it doesn't matter. removing the default

src/components/DayPickerNavigation.jsx
nkinser: i don't think i can do this since i need to pass e into `onPrevMonthClick`, and same for `onNextMonthClick` below

ljharb: ah, good point, my bad

PR #1896, opened 2020-01-13
feat(a11y): added ariaInvalid and ariaDescribedby props to SDP

jrparish:
- `ariaInvalid` allows for more accessible form validation
- `ariaDescribedby` allows users to use custom elements for the field description instead of the built in `screenReaderMessage` behavior.


src/components/DateInput.jsx
```diff
@@ -79,6 +79,10 @@ const defaultProps = {
 
   // accessibility
   isFocused: false,
+  ariaLabel: undefined,
+  ariaInvalid: undefined,
```
ljharb: shouldn't this be false, since it's a boolean?
```suggestion
  ariaInvalid: false,
```

src/components/SingleDatePicker.jsx
```diff
@@ -50,6 +50,8 @@ const defaultProps = {
   id: 'date',
   placeholder: 'Date',
   ariaLabel: undefined,
+  ariaInvalid: undefined,
```
ljharb:
```suggestion
  ariaInvalid: false,
```

src/components/SingleDatePickerInput.jsx
```diff
@@ -59,6 +61,8 @@ const defaultProps = {
   children: null,
   placeholder: 'Select Date',
   ariaLabel: undefined,
+  ariaInvalid: undefined,
```
ljharb:
```suggestion
  ariaInvalid: false,
```

src/components/SingleDatePickerInputController.jsx
```diff
@@ -109,6 +109,10 @@ const defaultProps = {
 
   // accessibility
   isFocused: false,
+  ariaLabel: undefined,
+  ariaInvalid: undefined,
```
ljharb:
```suggestion
  ariaInvalid: false,
```

jrparish:
> Can you elaborate on "allows for more accessible form validation"?

This attribute can be set to let a screen reader know a field is not valid. 
For instance, a form may require a proper date. Say a user types in a partial date, which is invalid for the form, we would set `aria-invalid` to `true` to indicate to the user that the field is not correct.

src/components/DateInput.jsx
jrparish: If we set this to `false` then the attribute will always appear in the DOM. Is that the behavior we want? If not set, the default is `false`.

I'm now also realizing this property can accept a couple string values - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-invalid_attribute

ljharb: ah, true. in that case `undefined` is the proper default, and i'm glad to see the more accurate propType :-)

src/components/DateInput.jsx
```diff
@@ -51,13 +49,15 @@ const propTypes = forbidExtraProps({
 
   // accessibility
   isFocused: PropTypes.bool, // describes actual DOM focus
+  ariaLabel: PropTypes.string,
+  ariaInvalid: PropTypes.oneOfType([PropTypes.bool, PropTypes.oneOf(['grammar', 'spelling'])]),
```
ljharb: can this propType be shared, instead of redefined in multiple places?

src/components/DateInput.jsx
```diff
@@ -79,6 +79,10 @@ const defaultProps = {
 
   // accessibility
   isFocused: false,
+  ariaLabel: undefined,
+  ariaInvalid: undefined,
+  ariaDescribedby: undefined,
+  screenReaderMessage: '',
```
backwardok: This isn't part of your change, but I wonder if this should default to `undefined` instead of an empty string

src/components/DateInput.jsx
```diff
@@ -234,7 +240,8 @@ class DateInput extends React.PureComponent {
           disabled={disabled}
           readOnly={typeof readOnly === 'boolean' ? readOnly : isTouch}
           required={required}
-          aria-describedby={screenReaderMessage && screenReaderMessageId}
+          aria-describedby={ariaDescribedby || (screenReaderMessage && screenReaderMessageId)}
```
backwardok: There may be cases where someone may want to have both `screenReaderMessage` and `ariaDescribedBy` set. This creates a bit of a complication as to the order, since it's hard to know what order to put them without knowing what content is there. Alternatively, it might be good to make the two props mutually exclusive so that consumers know that they can only use one.

src/components/SingleDatePickerInputController.jsx
```diff
@@ -109,6 +109,10 @@ const defaultProps = {
...
+  ariaDescribedby: undefined,
+  screenReaderMessage: 'Date',
```
backwardok: Not from your code, but this is probably a bad value to set as a default. It's both set to something that's not particularly useful and also is hardcoded in English.

src/components/DateInput.jsx
jrparish: Which direction do you think is best?

backwardok: Perhaps for simplicity, they can be mutually exclusive. If someone wants to use `ariaDescribedby` and have another screen reader message, they could do that manually and pass in the related `id` with `ariaDescribedby`.

PR #1900, opened 2020-01-14
Add ability to not show prev or next navigation buttons

nkinser:
Add capability to not show prev or next navigation buttons.

#### No nav prev
no-nav-prev

#### No nav next
no-nav-next

### DIFF
--- a/src/components/DayPickerNavigation.jsx
+++ b/src/components/DayPickerNavigation.jsx
@@ -66,220 +66,226 @@ const defaultProps = {
   showNavNextButton: true,
 };
 
-function DayPickerNavigation({
-  inlineStyles,
-  isRTL,
-  disablePrev,
-  disableNext,
-  navPosition,
-  navPrev,
-  navNext,
-  onPrevMonthClick,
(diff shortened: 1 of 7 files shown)


nkinser:
@ljharb Tests are failing for only certain versions of node and react. Any suggestions on how to fix this?

ljharb:
ah - `return null` from an SFC requires React 15+; one solution is to make it a class component (simplest); another is to return an empty div; another is to try to detect react < 15 and return an empty div only in that case.

src/components/DayPickerNavigation.jsx
```diff
@@ -66,220 +66,228 @@ const defaultProps = {
...
-  renderNavNextButton,
-  showNavPrevButton,
-  showNavNextButton,
-  styles,
-}) {
-  if (!showNavNextButton && !showNavPrevButton) {
-    return null;
+class DayPickerNavigation extends React.PureComponent {
+  constructor(props) {
+    super(props);
```
ljharb: this is just the default constructor, so you can omit it entirely

PR #1902, opened 2020-01-15
Keep scroll position when prev months rendered on vertical scrollable calendar

nkinser:
Keep track of the scroll position so that the visible month stays in view when prev months are rendered for the vertical scrollable calendar.

#### Before
wrong-cal-position

#### After
correct-cal-position

### DIFF
--- a/src/components/DayPicker.jsx
+++ b/src/components/DayPicker.jsx
@@ -25,6 +25,7 @@ import getCalendarMonthWidth from '../utils/getCalendarMonthWidth';
 import calculateDimension from '../utils/calculateDimension';
 import getActiveElement from '../utils/getActiveElement';
 import isDayVisible from '../utils/isDayVisible';
+import isSameMonth from '../utils/isSameMonth';
 
 import ModifiersShape from '../shapes/ModifiersShape';
 import NavPositionShape from '../shapes/NavPositionShape';
@@ -215,6 +216,7 @@ class DayPicker extends React.PureComponent {
 
     this.hasSetInitialVisibleMonth = !props.hidden;
     this.state = {
+      currentMonthScrollTop: null,
       currentMonth,
       monthTransition: null,
       translationValue,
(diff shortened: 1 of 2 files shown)


nkinser:
@ljharb I'm failing the coverage check but most of the lines highlighted as untested are not additions by me in this PR. Any suggestions?

ljharb:
If every line you've added is fully covered - every branch - then this means you've added enough lines to change the fraction, which makes coverage decrease. The desired effect is to force you to add tests to cover the *preexisting* uncovered lines.

However, it's fine to merge through as well; it'll just put some future PR author in the same boat.

src/components/DayPicker.jsx
```diff
@@ -355,11 +379,13 @@ class DayPicker extends React.PureComponent {
...
   }
 
-  componentDidUpdate(prevProps) {
+  componentDidUpdate(prevProps, prevState) {
     const {
       orientation, daySize, isFocused, numberOfMonths,
     } = this.props;
-    const { focusedDate, monthTitleHeight } = this.state;
+    const {
+      currentMonth, currentMonthScrollTop, focusedDate, monthTitleHeight,
```
ljharb:
```suggestion
      currentMonth,
      currentMonthScrollTop,
      focusedDate,
      monthTitleHeight,
```

test/components/DayPicker_spec.jsx
```diff
@@ -1027,6 +1027,28 @@ describe('DayPicker', () => {
...
+        });
+
+        it('updates transitionContainer ref`s scrollTop currentMonth changes', () => {
+          sinon.spy(DayPicker.prototype, 'setTransitionContainerRef');
+          const wrapper = mount(<DayPicker orientation={VERTICAL_SCROLLABLE} />);
+          const prevScrollTop = wrapper.transitionContainer.scrollTop;
+          wrapper.setState({
+            currentMonth: moment().subtract(1, 'months'),
+          });
+          expect(wrapper.transitionContainer.scrollTop).to.not.equal(prevScrollTop);
```
```suggestion
          expect(wrapper.transitionContainer).to.not.have.property(’scrollTop’, prevScrollTop);
```

PR #1911, opened 2020-01-28
Fixing render conditional in DayPickerKeyboardShortcuts

kevinthepan:
Fixing conditional so we don't render the function definition.


ljharb:
@kevinthepan it'd be great if you didn't merge PRs both without approvals and also without regression tests. Please add tests in a followup.

kevinthepan:
@ljharb Yeah that's totally on me, will follow up. Thanks for the reminder!

PR #1913, opened 2020-01-28
Add option to allow days violating min nights to be clicked

nkinser:
This PR adds a prop that allows days violating min nights to be selected as the end date. I would like to show an error message when a day blocked by min nights is pressed, but since currently `onDayClick` returns immediately if the day violates the min nights requirement, I don't have a way of knowing that the day has been clicked. This new logic makes it so that all the same modifiers are applied to days violating min nights, but in the `onDayClick` function it doesn't immediately return and it calls `onDatesChange`.
cal-min-nights-clicking

### DIFF
--- a/examples/DayPickerRangeControllerWrapper.jsx
+++ b/examples/DayPickerRangeControllerWrapper.jsx
@@ -29,6 +29,7 @@ const propTypes = forbidExtraProps({
   isOutsideRange: PropTypes.func,
   isDayBlocked: PropTypes.func,
   isDayHighlighted: PropTypes.func,
+  daysViolatingMinNightsCanBeClicked: PropTypes.bool,
 
   // DayPicker props
   enableOutsideDays: PropTypes.bool,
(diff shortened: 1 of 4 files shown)


examples/DayPickerRangeControllerWrapper.jsx
```diff
@@ -122,7 +124,17 @@ class DayPickerRangeControllerWrapper extends React.Component {
...
 
   onDatesChange({ startDate, endDate }) {
-    this.setState({ startDate, endDate });
+    const { daysViolatingMinNightsCanBeClicked, minimumNights } = this.props;
+    let doesNotMeetMinNights = false;
+    if (daysViolatingMinNightsCanBeClicked && startDate && endDate) {
+      const dayDiff = endDate.diff(startDate.clone().startOf('day').hour(12), 'days');
+      doesNotMeetMinNights = dayDiff < minimumNights && dayDiff >= 0;
+    }
+    if (doesNotMeetMinNights) alert('That day does not meet the minimum nights requirement');
```
ljharb: this needs to be removed before merging

nkinser: @ljharb just curious why this needs to be removed?

ljharb: oh oops, i didn't realize this was an example; so that makes it not blocking, but `alert` blocks the main JS thread so it shouldn't ever be used in any scenario, including examples.

nkinser: @ljharb updated to show the message in the calendar info instead

PR #1924, opened 2020-02-07
fix(Daypicker): add blur handler to capture all blur events

### LINKED ISSUE #1571: Unexpected onBlur behavior for DayPickerRangeController?
**react-dates version**
react-dates@20.0.0

**Describe the bug**
I've been playing around with the `DayPickerRangeController` and the onBlur behavior does not seem to work as expected when tabbing through. I added a window alert on the default storybook example and the onBlur alert did not show up when tabbing 'out' of the calendar, however it did show up on date selection.

In storybook:
```
  .add('default', withInfo()(() => (
    <DayPickerRangeControllerWrapper
      onOutsideClick={action('DayPickerRangeController::onOutsideClick')}
      onPrevMonthClick={action('DayPickerRangeController::onPrevMonthClick')}
      onNextMonthClick={action('DayPickerRangeController::onNextMonthClick')}
      onBlur={() => window.alert('onBlur')}
    />
```
If you have custom methods that you are passing into a `react-dates` component, e.g. `onDatesChange`, `onFocusChange`, `renderMonth`, `isDayBlocked`, etc., please include the source for those as well.

**Screenshots/Gifs**
https://www.useloom.com/share/8600023d0ce64ff1b082b8fffafa2d85

 - OS: MacOS Mojave 10.14.3
 - Browser: Chrome
 - Version:  72.0.3626.119

With custom onBlur

tanmoyopenroot:
#### PR checklist

- [x] Addresses an existing issue: fixes #1571 
- [x] bug
- [x] Includes tests
- [ ] Documentation update

#### Overview of change:
Ordinarily, the focus and blur events in the DOM dot not bubble. React will bubble these events through its event model. So we can listen for blur events at the root of a component and react to these events fired on its child nodes.

### DIFF
--- a/test/components/DayPicker_spec.jsx
+++ b/test/components/DayPicker_spec.jsx
@@ -31,6 +31,14 @@ describe('DayPicker', () => {
     sinon.restore();
   });
 
+  it('should pass onBlur to <DayPicker />', () => {
+    const onBlurStub = sinon.stub();
+    const wrapper = shallow(<DayPicker onBlur={onBlurStub} />);
+    expect(wrapper.prop('onBlur')).to.equal(onBlurStub);
+    wrapper.prop('onBlur')();
+    expect(onBlurStub.callCount).to.equal(1);
+  });
+
   describe('#render', () => {
     describe('renderWeekHeader', () => {
       it('there are 7 elements on each .DayPicker__week-header class', () => {
--- a/src/components/DayPicker.jsx
+++ b/src/components/DayPicker.jsx
@@ -240,6 +240,7 @@ class DayPicker extends React.PureComponent {
     this.setCalendarMonthGridHeightTimeout = null;
 
     this.onKeyDown = this.onKeyDown.bind(this);
+    this.onBlur = this.onBlur.bind(this);
     this.throttledKeyDown = throttle(this.onFinalKeyDown, 200, { trailing: false });
     this.onPrevMonthClick = this.onPrevMonthClick.bind(this);
     this.onPrevMonthTransition = this.onPrevMonthTransition.bind(this);
@@ -428,6 +429,12 @@ class DayPicker extends React.PureComponent {
     if (!MODIFIER_KEY_NAMES.has(e.key)) {
       this.throttledKeyDown(e);
     }
+  }
+
+  onBlur(e) {
+    const { onBlur } = this.props;
+
+    onBlur(e);
   }
 
   onFinalKeyDown(e) {
@@ -1179,6 +1186,7 @@ class DayPicker extends React.PureComponent {
               role="application"
               aria-roledescription={phrases.roleDescription}
               aria-label={phrases.calendarLabel}
+              onBlur={this.onBlur}
             >
               {!verticalScrollable && navPosition === NAV_POSITION_TOP && this.renderNavigation()}
 


test/components/DayPicker_spec.jsx
```diff
@@ -31,6 +31,14 @@ describe('DayPicker', () => {
     sinon.restore();
   });
 
+  it('should pass onBlur to <DayPicker />', () => {
+    const onBlurStub = sinon.stub();
+    const wrapper = shallow(<DayPicker onBlur={onBlurStub} />);
+    expect(wrapper.prop('onBlur')).to.equal(onBlurStub);
+    wrapper.prop('onBlur')();
+    expect(onBlurStub.callCount).to.equal(1);
```
```suggestion
    expect(onBlurStub).to.have.property('callCount', 1);
```

test/components/DayPicker_spec.jsx
```diff
@@ -31,6 +31,14 @@ describe('DayPicker', () => {
     sinon.restore();
   });
 
+  it('should pass onBlur to <DayPicker />', () => {
+    const onBlurStub = sinon.stub();
+    const wrapper = shallow(<DayPicker onBlur={onBlurStub} />);
+    expect(wrapper.prop('onBlur')).to.equal(onBlurStub);
```
```suggestion
    expect(wrapper.props()).to.have.property('onBlur', onBlurStub);
```

tanmoyopenroot:
> This makes the `onBlur` prop not just apply to the keyboard picker, but also to the DayPicker. Do you think the consumer would want to know which one is triggering it? Do you think it could be a breaking change for someone who's already passing onBlur?

Could you please clarify it for me?

ljharb:
Someone who is currently passing `onBlur` expects it to only be fired from the current components it's passed to; silently and suddenly passing it to new components means their callback will suddenly be invoked more times than they expect, on different components - that makes this a breaking change for them.

tanmoyopenroot:
> Someone who is currently passing `onBlur` expects it to only be fired from the current components it's passed to; silently and suddenly passing it to new components means their callback will suddenly be invoked more times than they expect, on different components - that makes this a breaking change for them.

ok, I will fix it.
Do we need to listen to all onblur events within the calendar(like blur event from navigation button, shortcut and day click) or only when it loses focus from the calendar boundary?

src/components/DayPickerRangeController.jsx
```diff
@@ -668,8 +675,6 @@ export default class DayPickerRangeController extends React.PureComponent {
     } else {
       onDatesChange({ startDate, endDate });
     }
-
-    onBlur();
```
ljharb: removing this call is a breaking change

tanmoyopenroot: But blur event from `CalendarDay.jsx` is being captured but this(`<div onBlur={this.onBlur}>`) in `DayPickerRangeController.jsx`

tanmoyopenroot: the reason for `onBlur` call inside `onDayClick` function in `DayPickerRangeController` is because of [this](https://github.com/airbnb/react-dates/blob/master/src/components/CalendarDay.jsx#L160).
We can pass custom `renderCalendarDay` which might not trigger blur event and on day click may unnecessary call `onBlur` if placed inside of `onDayClick`

PR #1929, opened 2020-02-25
[Fix] getVisibleDays

### LINKED ISSUE #698: Disabled dates missing styles if outside current month
I am using ```SingleDatePicker``` with the following configuration:

```
<SingleDatePicker
    date={ this.state.date }
    onDateChange={ this.handleChange }
    onFocusChange={() => null}
    numberOfMonths={1}
    focused
    displayFormat="DD/MM/YYYY"
    firstDayOfWeek={1}
    isDayBlocked={isDayBlocked}
    enableOutsideDays
/>
```

my ```isDayBlocked``` function:

```
const isDayBlocked = momentDate => {
    const dayString = momentDate.format('ddd');
    return dayString === 'Sat' || dayString ==='Sun'
}
```

This effectively renders all dates as valid apart from previous dates and weekend dates
However when I render the dates (as an example shown below):

calendar sept
calendar october

note that September 30th in September month is disabled but October 1st is not, when it should be. In the October month September 30th is not showing as disabled and neither is November 5th, but November 4th is. On inspection it appears the last date of the month is only being rendered with class ```CalendarDay CalendarDay--outside``` when it should have class ```CalendarDay CalendarDay--outside CalendarDay--blocked CalendarDay--blocked-calendar``` (comparing November 4th and November 5th classes in the table). I think there is a similar problem with the final outside date on the preceding month.

I have a workaround for the final date of the month view in the css which is to add

```
.CalendarDay--outisde:last-of-type {
[STYLE AS YOU WANT HERE]
}
```

But actually fixing the issue would be helpful! It should be noted these dates aren't selectable, they just appear to be missing styling

d1ckeys:
fix #698 

Pass firstDayOfWeek to getVisibleDays


ljharb (changes requested):
This looks great! Can we also add tests for the two components that are changed?

d1ckeys:
@ljharb 
Thanks for looking at this. Do you mean I should add tests for these components before merging this PR?

ljharb:
@d1ckeys yes

PR #1941, opened 2020-03-06
Issue 1937 | Pass invalid date string to date change handlers

marcgarygray:
Add `invalidDateString` to args for date change handlers (`onDateChange` and `onDatesChange`) whenever the date string is invalid. Add `storybook` stories and unit tests to support changes.

### DIFF
--- a/examples/DateRangePickerWrapper.jsx
+++ b/examples/DateRangePickerWrapper.jsx
@@ -24,6 +24,7 @@ const propTypes = {
   stateDateWrapper: PropTypes.func,
   initialStartDate: momentPropTypes.momentObj,
   initialEndDate: momentPropTypes.momentObj,
+  onInvalidInput: PropTypes.func,
 
   ...omit(DateRangePickerShape, [
     'startDate',
@@ -40,6 +41,7 @@ const defaultProps = {
   autoFocusEndDate: false,
   initialStartDate: null,
   initialEndDate: null,
+  onInvalidInput: () => null,
 
   // input related props
   startDateId: START_DATE,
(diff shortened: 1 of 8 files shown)


marcgarygray:
I think the CI tests collided because one was running while I merged `master` into my branch. My branch itself appears to have passed CI but the merge commit has a cancelled `travis-ci` job and that's the one that's linked in the review summary box. Not sure what the steps are to re-run the build.

renaudl:
Hey, Looks good to me but not enough of an expert to fully review.

amast09:
Having this merged would be AMAZING. Based on the current `SingleDatePicker` props, there is no way to distinguish between an invalid date and an empty date. This means we are unable to have an "optional" date input because we won't be able to tell if the user has a true invalid date or they just cleared the input. To overcome this we are using the following hack to determine whether or not the user has "cleared" the input or not,

```
useEffect(() => {
  const input = document.getElementById(props.id);

  const changeHandler = (e: Event) => {
    if (e.target) {
      setCurrentInputValue((e.target as HTMLInputElement).value);
    }
  };

  if (input) {
    input.addEventListener('change', changeHandler);
  }

  return () => {
    const inputToRemoveHandlerFrom = document.getElementById(props.id);

    if (inputToRemoveHandlerFrom) {
      inputToRemoveHandlerFrom.removeEventListener('change', changeHandler);
    }
  };
}, []);
```

which is admittedly fragile and prone to insidious bugs because we are operating outside of React. We felt this approach was better as opposed to completely forking this repository.

Not sure who the official maintainers are of this repo but it would be wonderful if they merged this PR or something close to it that enables this sort of "optional" behavior.

pranjaljately:
Any updates on this? Seems to be something highly sought after...

AndyxWood:
Bumping this - would love to see more information returning from date change handlers in all relevant DateChangePicker components.

ljharb (approved):
Seems reasonable. It'll need to be rebased.

marcgarygray:
> Seems reasonable. It'll need to be rebased.

Sure! To which branch?

G07cha:
Bumping, the PR has been hanging for quite a while awaiting for feedback.

Issue #208, opened 2016-12-06
Switch from moment to date-fns?


aaronjensen:
I don't know what it would take to do, but reducing dependencies on moment would be a great thing.

https://date-fns.org/

jpollard-cs:
@aaronjensen I wouldn't assume folks are familiar with this library. It would probably help if you outlined the advantages of this library over moment (and in the context of react-dates).

aaronjensen:
There are some reasons on this issue: https://github.com/date-fns/date-fns/issues/275

The primary reason for me is the size of moment with or without its locales (I hack them out using webpack, but moment is still pretty big). If you don't need all of the functionality from date-fns you can just use the functions you need, making the size even smaller. You could potentially get rid of the peerDependency and simply bundle the functionality you need.

ljharb:
If `date-fns` gains traction in the community, then it might be reasonable to support *both* `moment` and `date-fn` objects - although that wouldn't make the dependencies any smaller.

The author of date-fns points out some reasons in https://github.com/date-fns/date-fns/issues/275#issuecomment-264934189 that we wouldn't want to support it just yet - and when it *did* have those capabilities, there's no reason not to switch over to it completely.

In general, *nobody* doesn't need "every locale known to humans" - because every locale you fail to account for, is a human you're excluding. That choice goes against the spirit of both the project and Airbnb.

andrewmclagan:
I would be very interested in this, date-fns has gained allot of traction and looking forward seems to have allot more momentum and a better footing.

shinzui:
@ljharb `date-fns` supports 32 locales now with more coming. It's also quickly gaining traction in the community.

ljharb:
When it supports every locale that moment does, and passes all of moments test cases, then it is worth having this conversation again. :-)

andrewmclagan:
Agreed. Although, without digging too much @ljharb do you think there is allot of work to swap out implementations?

ljharb:
That would depend on the moment APIs we're using, and how date-fns' API compares - assuming it's a change we want to make.

rubencodes:
@andrewmclagan Just browsed through the moment usage in this project (I'm also interested in this swap) and it's very minimal (IMO), should be a relatively easy swap I think... Might make a fork to try it out. 

FWIW Moment.js is great but just *huge*, and if we're talking about "excluding humans", being data-usage-savvy is also a super important consideration.

ljharb:
[...]

PR #1947, opened 2020-03-16
WIP: Date library support via driver interface.

tonyhb:
STATUS: WIP

This PR opens the door to supporting multiple date libraries.  Out of the
box it includes support for Moment, and I'm currently testing Luxon. I'd
also love to add js-joda, that library is dope.

## How

Instead of calling moment directly within each component, we instead
create a driver which wraps all functions we need from moment.  This
driver supports (theoretically) multiple runtimes;  we can use moment or
swap out for Luxon/date-fns/js-joda as we wish.

## Driver

The implementation of the driver is similar to a go interface.  A driver
must implement all of the calls that we use throughout our components to
fulfil the interface.

The driver also assumes that any function call is stateless;  calling
`driver.add(date, { days: 2 })` must return the new date without
mutating the argument.

As most date functions accept the library's date instance as the first
argument we also provide a utility to curry these functions w/ partial
application:

```js

import { driver, chain } from '../src/drivers/drivers';

chain(date).add({ days: 1}).startOf('day').value();
```

## Status

This is a WIP PR.  I'm not particulary happy with the way that drivers
are defined, with the way that they're instantiated (via mutating), and
there are a few errors that I need to get to.  But it's the direction
I'm heading in after wanting to use this awesome datepicker with Luxon.

Thoughts appreciated as I work on it :)


tonyhb:
This aims to fix the long standing (4 years!) issue here: https://github.com/airbnb/react-dates/issues/208

src/components/CalendarDay.jsx
```diff
@@ -42,7 +42,7 @@ const defaultProps = {
   onDayMouseEnter() {},
   onDayMouseLeave() {},
   renderDayContents: null,
-  ariaLabelFormat: 'dddd, LL',
+  ariaLabelFormat: 'cccc, DDD',
```
tonyhb: TODO: aria label inside drivers

src/components/DateRangePickerInputController.jsx
```diff
@@ -118,9 +116,10 @@ const defaultProps = {
   reopenPickerOnClearDates: false,
   withFullScreenPortal: false,
   minimumNights: 1,
-  isOutsideRange: (day) => !isInclusivelyAfterDay(day, moment()),
+  isOutsideRange: (day) => !isInclusivelyAfterDay(day, driver.now()),
   isDayBlocked: () => false,
-  displayFormat: () => moment.localeData().longDateFormat('L'),
+  // TODO: @tonyhb driver.formats
+  displayFormat: 'D',
```
tonyhb: TODO: driver display format

src/drivers/luxon/driver.js
```diff
@@ -0,0 +1,149 @@
...
+  const dateObj = valid(d) ? d : toLuxonObject(d);
+  if (!dateObj) return null;
+  return dateObj.daysInMonth;
+}
+
+function firstDayOfWeek() {
+  // Luxon does not have this capability
+  // Note that Luxon's days of week range from 1-7;  internally we expect a range
+  // from 0-6.
+  return 0;
```
tonyhb: TBD:  This is when I was keeping this moment compatible, but the driver should really dictate the first day of week.  I need to figure out if returning `1` here messes with things.

src/drivers/luxon/toLuxonObject.js
```diff
@@ -0,0 +1,24 @@
...
+    ? [customFormat, DISPLAY_FORMAT, ISO_FORMAT]
+    : [DISPLAY_FORMAT, ISO_FORMAT];
+
+  let date;
+  dateFormats.find((format) => {
+    date = DateTime.fromFormat(dateString, format);
+    return date.isValid;
+  });
+
+  return date ? date.hour(12) : null;
```
tonyhb: `date ? date.set({ hours: 12 }) : null`

src/constants.js
```diff
@@ -1,7 +1,3 @@
-export const DISPLAY_FORMAT = 'L';
-export const ISO_FORMAT = 'YYYY-MM-DD';
-export const ISO_MONTH_FORMAT = 'YYYY-MM'; // TODO delete this line of dead code on next breaking change
```
ljharb: removing these exports is a breaking change

src/drivers/luxon/driver.js
```diff
@@ -0,0 +1,149 @@
+import PropTypes from 'prop-types';
+import { DateTime } from 'luxon';
```
ljharb: this means that react-dates would cause both moment and luxon to be installed.

moment is a peer dep because react-dates, and the user, need to be using the same version of moment to ensure that things work properly. For luxon, it'd be the same requirement, which means that adding luxon would cause react-dates to have a peer dep requirement on *both* moment and luxon, and users would have to add *both* of them to their own package.json (which is a breaking change)

package.json
```diff
@@ -119,6 +120,7 @@
     "enzyme-shallow-equal": "^1.0.0",
     "is-touch-device": "^1.0.1",
     "lodash": "^4.1.1",
+    "luxon": "^1.22.0",
```
ljharb: this needs to be a peer dep, not a regular dep, because it's critical that the user and the library be using the same version of luxon.

src/constants.js
tonyhb: Yeah I figured it would be

package.json
tonyhb: Thanks.  Probably did `yarn install` while testing.

src/drivers/luxon/driver.js
tonyhb: yep, i get you.  it's in the same repo to test (it's all a WIP), though with this change I anticipated that every driver would be moved to its own repo... even temporal when it comes along.

tonyhb:
Yeah I get you.  I'm trying to generalize the functions used within the components so that switching to any time system is easy (temporal or not), though given temporal should be the "standard" it's less important.

Essentially, if we: 1) know we're moving to temporal, and 2) know that we have moment... we could move forward with drivers to release a new major breaking change with backwards compatibility (via the moment driver)

src/drivers/luxon/driver.js
tonyhb: The luxon driver has been removed and will be uploaded to a separate repo.

tonyhb:
It looks as though someone is going to have to go through the pain of changing all moment/datetime related functions within a component when the move to Temporal comes out.

With this PR, that is now a _lot_ easier (a 10m job). 

This PR also provides backwards compatibility to those wanting to use Moment on older browsers... if you're using moment everywhere in your app, you'd want to refactor moment out of your code altogether.

Lastly, it allows people to start using other date libraries (even Temporal with the current polyfill) right now.

ljharb:
When Temporal is available, this library will need to do a pretty major breaking change to support it - it won't just be one driver to another, but entirely different sets of props (since a specific Temporal type might cover what is currently 3-4 separate props, for example).

tonyhb:
I dont' think that changing props and adding drivers are mutually exclusive.

package.json
```diff
@@ -93,6 +93,7 @@
     "karma-mocha": "^1.3.0",
     "karma-sinon": "^1.0.5",
     "karma-webpack": "^4.0.2",
+    "luxon": "^1.22.0",
```
tonyhb, line 96: tood: remove

CoryDanielson:
What would the usage of the library look like with these changes?

```js
import "react-dates/driver/luxon"
import { DateRangePicker } from "react-dates";
```

It doesn't seem like importing the driver has any side-effect that would cause the DateRangePicker to start using it.

olso:
@tonyhb Awesome work!

tonyhb:
@CoryDanielson Yeah, honestly not the biggest fan of those mutable imports!  I was hacking away to get a POC done to see what it'd look like, whilst also trying to minimize API damage.

I'm back in the 3D/backend world right now and likely won't be able to take this further, but if anyone wants to carry it feel free!

ljharb:
To make sure expectations are clear, though: given that moment remains the defacto ecosystem standard, and Temporal will hopefully be stage 3 within the next year (at which point, every date library will immediately become either largely obsolete, or will interoperate smoothly with Temporal, and thus react-dates would drop moment and switch to Temporal exclusively), it's going to be a really tough sell to add complexity for the common case (using moment) in the hopefully brief meantime.

olso:
https://momentjs.com/docs/#/-project-status/

> We now generally consider Moment to be a legacy project in maintenance mode. It is not dead, but it is indeed done.

ljharb:
@olso that doesn't change the thought process here whatsoever.

PR #1643, opened 2019-05-14
Accessibility: added aria-hidden=true to navigation.

nicolad:
Fixes: https://github.com/WordPress/gutenberg/issues/15330

PR #1957, opened 2020-03-30
Accessibility: added aria-hidden=true to navigation.

nicolad:
This PR attempts to fix the `a11y` issue described here: https://github.com/WordPress/gutenberg/issues/15330


ljharb:
(replacement for #1643, unless that one can be reopened)

nicolad:
>This change shouldn't need to be made.

@backwardok is it ok to close this PR, then?
***
I am asking this in context of [this comment](https://github.com/airbnb/react-dates/pull/1643#issuecomment-606158964) from @ljharb

Issue #1966, opened 2020-04-07
unable to change DayPickerSingleDateController visibility using focused prop


louisbyrne89:
**react-dates version**
 react-dates@21.8.0

**Describe the bug**
I'm having trouble controlling the visibility of the DayPicker using the DayPickerSingleDateController component. Setting the focused property to false was having no effect on the display of the DayPicker. I noticed while going through the source code, that the application of the DayPicker__hidden class is controlled by the 'monthTitleHeight' prop (line 1147, DayPicker.jsx). This stikes me as very odd, shouldn't this use the hidden prop?

louisbyrne89:
Can't seem to link my PR but its here: https://github.com/airbnb/react-dates/pull/1967

camille-dev:
I seem to have the same issue for the SingleDatePicker: once clicked, DayPicker stays open no matter what (click outside, or select a date). This is despite the focused prop being set to false.
The package is up-to-date (21.8.0)

adrien3d:
Don't know if anyone have still the issue, but, as the `SingleDatePicker` component `onFocusChange` returns an object like `focused: true`, I had to pass the object.focused like that to make it work:
```
focused={focusSDP}
onFocusChange={(foc) => setFocusSDP(foc.focused)}
```

PR #1967, opened 2020-04-07
fix set visibility to hidden for day picker component

louisbyrne89:


louisbyrne89:
Fixes this bug: https://github.com/airbnb/react-dates/issues/1966

Issue #1023, opened 2018-02-14
SingleDatePicker: No way to distinguish between invalid date and empty input


viktoriyavulfson:
`SingleDatePicker`'s `onDateChange` handler is fired with `null` when user's input doesn't validate to a date. This can happen when the date is, in fact, invalid `13/13/2018` or when user input is empty. 

Is there a way to tell these scenarios apart?

charliedavison:
+1

austenpayan:
This is also tripping me up a little. My scenario is that I want to send a request to the server if the user "clears" the input, but I have no way of knowing if the `null` date arg was because of an actual clearing of the input or a temporary edit that put the input into an invalid state. 

Perhaps additional arguments could be passed to the change handler, in addition to the date value? Something that refers to the validity of the current state of the input?

abidvf:
+1

JohnnyBDude:
+1 this is driving me insane:)

Its probably a bug though, as onDateChanged should not be triggered at all.
See README:

> The SingleDatePicker also manages internal state for partial dates entered by typing (although onDateChange will not trigger until a date has been entered completely in that case).

thornebrandt:
+1

mnikolaus:
+1

JohnnyBDude:
I fixed this using this strategy:

- onDateChange callback does not change state of wrapping component if it recieves null as an argument.
- Put a "clear" button next to the datepicker that sets the state to null.

Not pretty but works....

I would also consider contributing a bugfix, but not sure where to start - i dont really have any experience in contribution to foss - i would love to get some though.

Flui:
+1  Same problem here

thornebrandt:
Looks like my +1 comment was marked as spam for some reason, but I am also experiencing the same problem.

ljharb:
All +1 comments are spam; github etiquette is to use an emoji reaction on the OP.

t83714:
We probably should make `onDateChange` fired with either a `Date` object or a string (contains whatever user's original input) to fix this issue

renaudl:
Hi all, huge issue for us. As we don't know if the date is being edited but invalid or just empy. Maybe use null / undefined.

KrisB1022:
RE: https://github.com/airbnb/react-dates/pull/2015

pranjaljately:
Any updates on this? Hoping to find a solution on this as well...

explodingdoggos:
really late i guess but i need this too

[...]

PR #2015, opened 2020-06-30
Issue 1023 pass invalid date to onchange

KrisB1022:
Resolve issue [1023](https://github.com/airbnb/react-dates/issues/1023)

When a date is invalid, it is hard to tell if it is invalid due to an empty string, where the user has not entered a date, or if the user entered an invalid date.

Currently the only way to determine if the date is invalid or just empty, is to query the DOM for the specific <input/> tags to read their values.

This change will pass a new argument to the onChangeHandler that will include the date string that was entered into the input.

* NOTE
There were some unrelated files that had linting errors related to having too many spaces. This PR also helps the team out in fixing those issues.


KrisB1022:
handled by https://github.com/airbnb/react-dates/pull/1941

PR #2021, opened 2020-07-14
Clear renderCalendarInfo timeout before assigning a new timeout (#1323)

smit1625:
This is a do-over of @mstanaland's work on https://github.com/airbnb/react-dates/pull/1736 to fix a memory leak issue where the timeout is not cleared before being set again. I am doing this so that we can be current with master, since the original PR went unmerged for six months due to lack of a unit test. This branch does contain a test displaying that `setState` is called by the timeout only once.

Original PR description by @mstanaland:

> Clears the setTimeout assigned to this.setCalendarInfoWidthTimeout before a new setTimeout is assigned/reassigned to prevent a possible memory leak.
  Fixes issue #1323

### DIFF
--- a/src/components/DayPicker.jsx
+++ b/src/components/DayPicker.jsx
@@ -367,6 +367,8 @@ class DayPicker extends React.PureComponent {
     // breaks the CSS transition.
     // The setTimeout will wait until the transition ends.
     if (this.calendarInfo) {
+      clearTimeout(this.setCalendarInfoWidthTimeout);
+
       this.setCalendarInfoWidthTimeout = setTimeout(() => {
         const { calendarInfoWidth } = this.state;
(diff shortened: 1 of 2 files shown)


smit1625:
Looks like the failures in Travis CI are not related to the code changed by this PR. @ljharb is this normal for this project?

ljharb:
In the future, please do not create a second, duplicate PR - post a link to your branch and I can update the original PR. Now both of these need to stay open and kept in sync for the duration.

ljharb:
If that's the only failure, it's something I can try to look into in the next week or so.

smit1625:
Okay sure, would you like me to close this PR?

ljharb:
@smit1625 no, please don't close it :-) i'll need to keep it in sync with #1736 until both are landed.

PR #2057, opened 2020-10-26
Add preventDefault for non-Tab, const keys in onKeyDown

### LINKED ISSUE #2055: Multiple arrow key presses in calendar causes page/scrollable element to scroll.
**react-dates version**
react-dates@21.8.0

**Describe the bug**
With the calendar days in focus, pressing up/down arrow keys quickly will cause the entire page to scroll.

1. go to: 
http://airbnb.io/react-dates/?path=/story/daypickersingledatecontroller--with-custom-input
This page has lengthy content in the example div, so it should be in a state where scrolling is required.
2 Tab into the calendar with keyboard and then rapidly press up/down arrow keys as if you're trying to navigate quickly through the days... the page/div will scroll, perhaps every 3rd or 4th keypress.

Most of the key presses are caught, but (perhaps in the throttle?) some slip through and cause the default scrolling behavior to occur.

If you have custom methods that you are passing into a `react-dates` component, e.g. `onDatesChange`, `onFocusChange`, `renderMonth`, `isDayBlocked`, etc., please include the source for those as well.

**Screenshots/Gifs**
If applicable, add screenshots or gifs to help explain your problem.

 - OS: Windows 10
 - Browser Chrome
 - Version 86.0.4240.111

Yes... http://airbnb.io/react-dates/?path=/story/daypickersingledatecontroller--with-custom-input

Additional - 
Might be a simple fix...
```
if (e.key !== 'Tab') e.preventDefault();
```
within the condional block at https://github.com/airbnb/react-dates/blob/master/src/components/DayPicker.jsx#L429-L431 before to the `throttledKeyDown` call...
I haven't fully tested it, but this seemed to fix the issue in a local modification.
Might need expansion if other keys should be allowed.

JohnC-80:
Should resolve #2055


ljharb (changes requested):
Thanks!

Let's also add a test to https://github.com/airbnb/react-dates/blob/master/test/components/DayPicker_spec.jsx#L240-L273 for this.

JohnC-80:
Will do! Sorry I'm hitting a bit of a first-time-contributor bump... I develop on Windows so the script commands are giving me a tough time actually running tests/building locally... putting this PR up I honestly left it up to the CI to get it there... I typically would employ something like `cross-env` to get things working - and I might just do enough there to try and get it to spun up without making too many changes... just hasn't been automatic yet and I've had to move into some other things dayjob-wise... will keep after it, thanks @ljharb ! Any suggestions you have would be great - currently I'm hung on the build CSS step after clearing the `BABEL ENV` hurdle.. any input would be great. Thanks for a stellar library.
image

ljharb:
No worries at all; indeed that command is not Windows-compatible. I'll fix it in master shortly, and then you can rebase and should be unblocked.

ljharb:
Instead of pushing it to master, I prepended them to this PR, so you should be set (reset your local branch to the pushed changes)

JohnC-80:
Thanks... sheesh, now it's stumbling on installing the peers for enzyme-adapter-react... 
image

ljharb:
@JohnC-80 what versions of node/npm are you using?

JohnC-80:
node = v12.13.0
npm = v6.13.0

ljharb:
hmm - that should be using `^16.0.0-0`, not `16.0.0-0`. if you do `npm install -g npm@6`, does it still fail?

JohnC-80:
@ljharb 
I'm upgraded to npm@6.14.8 and it's still bombing out with the same error installing those peerDeps for the enzyme-adapter.

ljharb:
@JohnC-80 hmm, I'm not actually sure :-/ it seems like it's worth doing a lot of manual testing here.

PR #2058, opened 2020-10-30
[Fix] #1513 Replace sass-specific  with runtime color calculation

11bit:
Fixes #1513

1. In generated css it replaces sass specific `darken` function with actual color.
2. Also works in css-in-js mode by calculating colors in runtime


ljharb (approved):
Thanks, this seems great

11bit:
Great, thanks!

ljharb:
It’s very confusing how tests passed here in node 6, since color2k doesn’t even parse in node 6 :-/

Issue #1748, opened 2019-08-09
React 16.9 warning


maxmoeschinger:
I am getting this warning when running react-dates with react 16.9:

<img width="524" alt="Skjermbilde 2019-08-09 kl  13 24 59" src="https://user-images.githubusercontent.com/15786838/62775842-27b62a80-baa9-11e9-976d-d6ed1cc6269f.png">

I am using react-dates version 20.2.5.

ljharb:
Since we support versions of react older than 16.9, and older than the unsafe variants, that warning is unavoidable.

NMinhNguyen:
> that warning is unavoidable.

I don’t think that’s necessarily true. We should be able to use https://github.com/reactjs/react-lifecycles-compat which even has tests for React 0.14.9 and has been mentioned in an older issue https://github.com/airbnb/react-dates/issues/1179

ljharb:
That’s fair; we’d have to make a breaking change to require the polyfill.

NMinhNguyen:
I don’t think we need to put the requirement on users and can just apply the polyfill to all `react-dates` components ourselves.

ljharb:
hmm, I do wonder why 0.14.9 would be required instead of 0.14.0 - but a breaking change to hoist the minimum to 0.14.9 is something i'd happily do, since i doubt any 0.14 users are on "not the latest 0.14" at this point.

AmberWilliams:
Is this going to be implemented?

ranneyd:
Upvote

ranneyd:
Part 1 of this fix is fixing it in one of this lib's dependencies.

https://github.com/airbnb/react-outside-click-handler/pull/36

jeremy-clearlabs:
@ranneyd Looks like that was fixed in `v1.3.0`. I tried to update `package.json`, but my tests weren't passing. Hoping someone else could raise a PR.

Kivylius:
Is there any progress on this? Can we just release v22 which doesn't support the older versions? or implement the required fix thats above? If not this library is good as dead to me and I can no longer use it moving forward.

ranneyd:
@ljharb because the fix has been around since react 16.3 (a year ago) so highly used component libraries (this) could fix it before this warning happens (one year later).

If this lib hasn't fixed this by now, will it be fixed before React 17 comes out?

Also a warning like this not only pollutes the console but would be totally unacceptable at my company if it was from in-house code.

It's also annoying as all hell.

[...]

PR #2059, opened 2020-10-30
changing name of deprecated react lifecycle methods

ranneyd:
https://github.com/airbnb/react-dates/issues/1748

- Instead of switching to `getDerivedStateFromProps`, I figured using the `UNSAFE` versions of the functions was less disruptive. Perhaps the lib will move to a functional/hooks-based implementation some day.
- The `UNSAFE` prefix on the function names upset eslint's `camelcase` rule. I copied the config from [here](https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb-base/rules/style.js#L24) but added the `allow` property


package.json
```diff
@@ -100,8 +100,8 @@
     "node-sass": "^4.12.0",
     "nyc": "^14.1.1",
     "raw-loader": "^0.5.1",
-    "react": "^0.14 || ^15.5.4 || ^16.1.1",
-    "react-dom": "^0.14 || ^15.5.4 || ^16.1.1",
+    "react": "^16.3.0 || ^17.0.0",
+    "react-dom": "^16.3.0 || ^17.0.0",
```
ljharb, line 104: just calling out that this makes it a breaking change.

src/components/CalendarMonth.jsx
```diff
@@ -109,7 +109,7 @@ class CalendarMonth extends React.PureComponent {
     this.setMonthTitleHeightTimeout = setTimeout(this.setMonthTitleHeight, 0);
   }
 
-  componentWillReceiveProps(nextProps) {
+  UNSAFE_componentWillReceiveProps(nextProps) {
```
ljharb, line 112: if we're doing the breaking change, we might as well migrate to getDerivedStateFromProps where we can? It's ok to keep the unsafe versions where that's not easy tho.

ljharb (changes requested):
You'll also want to update `.travis.yml` so it stops testing on React 14 and 15, and start testing on React 17.

The latter means this is blocked on https://github.com/enzymejs/enzyme/pull/2430 - so another alternative is requiring `^16.3` and waiting until a future PR to add (in a semver-minor) v17 support.

ranneyd:
@ljharb just removed < 16 from `.travis.yml`. I also removed react 17 from the dependencies; we can add that in later when enzyme et al. are updated.

src/components/CalendarMonth.jsx
ranneyd: I have a few thoughts on this:
- I want to rock the boat with this PR as little as possible. If people are upgrading from an older react to 16.3+, then upgrading this, if they have a bug it would be nice to say fairly confidently that it wasn't this and probably part of some other part of their migration.
- Related, a _ton_ of the logic in these components is in these lifecycle methods, so I see any substantive change to them as a larger-scope change.
- From my understanding, these methods aren't going anywhere anytime soon. My understanding is the react team has great interest in _discouraging_ their use rather than removing them.
- If you think this is the time/PR to do this lmk and I can see what I can do. Might be more appropriate to do it in two parts.
- Depending on what the future holds, this lib might go the way of many of the other libs out there and get a functional, hooks-based refactor. Maybe instead of migrating all this huge lifecycle method logic twice we just wait until that day comes (though I believe the migration to `getDerivedStateFromProps` is not nearly on the same scale as a migration to hooks, which would be..._non-trivial_.)

I will say in general I'm all about the "while we're at it" mentality. If it were my own repo or wasn't 427,576 weekly downloads then I'd feel more comfortable just doing it 😂 .

My vote is to do it this dumb, rename-only way first to get the warnings out and get the compatibility there, then either have a follow-up PR or include this in future refactoring conversations/planning.

ljharb: If we're going to migrate to hooks any time soon, we'd need to require React v16.9 now, and not just v16.3, even if we do this PR as-is first, and change other things later.

ranneyd: Yeah definitely not advocating hooks migration in the near-term.

andrew-pyle:
Is this still blocked? Any way a contributor could help? react-dates is preventing migration to React 17 for me.

ljharb:
Yes; this still needs rebasing, but also would still be a breaking change.

alex-pex:
Hello! We still need this issue to be fixed to migrate to React 17. Any news?

Issue #1592, opened 2019-03-21
DateRangePicker noBorder prop only applies to inputs


gidztech:
**react-dates version**
react-dates@19.0.3

**Describe the bug**
I'm using the `DateRangePicker` component and when I pass `noBorder` as a prop, the input fields no longer have borders, but the picker has a box shadow still.

**Source code (including props configuration)**
```
<DateRangePicker ...otherProps noBorder />
```

**Screenshots/Gifs**
![noborder](https://user-images.githubusercontent.com/15942538/54764408-a4766400-4bef-11e9-9e10-51dded2709c7.png)

`DayPicker` seems to have `DayPicker__withBorder` class, which contains the border. 

My expectation was that `DateRangePicker` would pass `noBorder` down to both components, not just the inputs.

**Desktop (please complete the following information):**
 - OS: Windows 10
 - Browser Chrome
 - Version 72.0.3626.121

**Is the issue reproducible in Storybook?**
No because theres no story with `noBorder` variation for `DateRangePicker`, but I can confirm that both individual components respect the prop. That leads me to think that the prop isn't being passed down properly from the parent.

PR #2065, opened 2020-11-09
[New] `DateRangePicker`: pass noBorder as noBorder to `DayPickerRangeController`

fedetaglia:
Hi

I'm following up on an issue raised https://github.com/airbnb/react-dates/issues/1592 about the `noBorder` props not passed to the `DayPickerRangeController`. As result the `noBorder` is only applied to the inputs and not to the day range panel.

With this PR also the date range will be rendered without border when `noBorder` is applied.

I‘ve also added a quick spec and story.


ljharb (approved):
Seems reasonable, thanks

PR #2066, opened 2020-11-11
[New] `DayPickerSingleDateController`: add `allowUnselect` prop option to allow unselecting dates

### LINKED ISSUE #2061: Feature Request: Support unselecting a selected date by clicking on it
**react-dates version**
react-dates@15.5.3

**Describe the bug**
This is mostly a feature request. When using the single date picker, it is not possible to unselect a currently selected date by clicking on the already-selected date. This FR was initially [filed in AMP's repo](https://github.com/ampproject/amphtml/issues/25609). [AMP uses the `DayPickerSingleDateController`](https://github.com/ampproject/amphtml/blob/b6313e372fdd1298928e2417dcc616b03288e051/extensions/amp-date-picker/0.1/single-date-picker.js#L33) for building the single date component. 

We would like to support the use-case of unselecting a date by clicking it again. However, instead of building this on our side, we thought it would be a good idea to fully support it on the `react-dates` side since that would be more sustainable and would potentially benefit others.

I've already explored a possible solution for this (see my comments in the [original FR](https://github.com/ampproject/amphtml/issues/25609)). I'd be happy to contribute with a PR for this.

Select a date on the [demo DayPickerSingleDateController](http://airbnb.io/react-dates/?path=/story/daypickersingledatecontroller--default) and then clicking that same date again. The date is not unselected on the second click.

joaopsrleal:
Adds a new property in DayPickerSingleDateController which allows users to unselect a previously selected date by clicking on it.
This PR fixes #2061.

I'm tentatively increasing the minor version given that this PR effectively adds new functionality in a backward compatible way. I'm happy to simply increase the patch version as well, even though this is not necessarily a bug fix.

**Alternative considered in the implementation:**
I've also considered making this functionality enabled by default, however there would be the risk that it would break users who would not handle undefined (or eventually null) in the onDateChange callback.

### DIFF
--- a/examples/DayPickerSingleDateControllerWrapper.jsx
+++ b/examples/DayPickerSingleDateControllerWrapper.jsx
@@ -19,6 +19,7 @@ const propTypes = forbidExtraProps({
   initialDate: momentPropTypes.momentObj,
   showInput: PropTypes.bool,
 
+  allowUnselect: PropTypes.bool,
   keepOpenOnDateSelect: PropTypes.bool,
   isOutsideRange: PropTypes.func,
   isDayBlocked: PropTypes.func,
@@ -56,6 +57,7 @@ const defaultProps = {
   showInput: false,
 
   // day presentation and interaction related props
+  allowUnselect: false,
   renderCalendarDay: undefined,
   renderDayContents: null,
   isDayBlocked: () => false,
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-dates",
-  "version": "21.8.0",
+  "version": "21.9.0",
```
ljharb, line 3: PRs shouldn't ever bump versions; I'll revert this when rebasing.

ljharb (approved):
LGTM modulo comment

package.json
joaopsrleal: Sorry about that. I didn't know. Thanks for the feedback!

Issue #1809, opened 2019-09-27
DRP: drop down not closing when (forward) tabbing out of focus


antjernberg:
**react-dates version**
react-dates@21.1.0

**Describe the bug**
I believe the calendar should close when you move the focus outside of the components within the calendar component. This is when happens when you back-tab, but not when you forward tab. Meaning, if you forward tab to another element which also uses some keyboard interaction and absolute positioned dropdowns, you could end up in an awkward position.

Problem: switching focus away from calendar using Tab does not close the Calendar dropdown. 

**Source code (including props configuration)**
This bug is found in any setup of the component that I've tried, also in the publicly available storybook here:
https://airbnb.io/react-dates/?path=/story/daterangepicker-drp--default
(Use the same prop setup as the storybook basic example)

**Screenshots/Gifs**
In the gif, you can see me tabbing to select the calendar. After tabbing through all the elements and focusing on *other* elements on the site, the dropdown remains open.
https://gyazo.com/fb17ed0af5d125c7566c0352a0cf6c28

Backwards tabbing works correctly however. It closes the dropdown when currently focusing the first element and pressing shift-tab.

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser: Chrome 
 - Version: 77.0.3865.90

**Is the issue reproducible in Storybook?**
Yes.
https://airbnb.io/react-dates/?path=/story/daterangepicker-drp--default

mmarkelov:
@antjernberg yeah that's true. I made PR about two month ago. And this fix issues like this. But I'm not sure that it will be merged(

PR #2084, opened 2021-01-27
[Fix] `DateRangePicker`, `DateRangePickerInput`: Fix keyboard navigation for second input

greglo:
Keyboard navigation is broken for the date range picker. The following bugs exist:
* When passing through the component with tab-navigation, after exiting the end-date input the dropdown does not close [1]
* When attempting to enter values into the end-date dropdown, users would have to tab backwards back to the dropdown to enter values [2]

The simplest and clearest solution that came to be is simply to put the day picker after the currently active input. In this way, the dropdown always appears next in the tab-order to the input the user has selected. The existing blur handler in the dropdown means that tabbing through the component will leave the dropdown closed at the end.

The react keys ensure that the element is just moved over in the DOM and so there is no flickering.

Fixes https://github.com/airbnb/react-dates/issues/1809

@majapw 

[1] Before, passing through the component using tab
jan27-before-pass-through

[2] Before, enter values with keyboard
jan27-before-enter

[3] After, passing through the component using tab
jan27-after-pass-through

[4] After, enter values with keyboard
jan27-after-enter

### DIFF
--- a/src/components/DateRangePicker.jsx
+++ b/src/components/DateRangePicker.jsx
@@ -481,6 +481,7 @@ class DateRangePicker extends React.PureComponent {
     /* eslint-disable jsx-a11y/click-events-have-key-events */
     return (
       <div
+        key="day-picker"
         ref={this.setDayPickerContainerRef}
         {...css(
           styles.DateRangePicker_picker,
--- a/src/components/DateRangePickerInput.jsx
+++ b/src/components/DateRangePickerInput.jsx
@@ -253,7 +253,7 @@ function DateRangePickerInput({
         regular={regular}
       />
 
-      {children}
+      {!isEndDateFocused && children}
 
       {
         <div
@@ -288,6 +288,8 @@ function DateRangePickerInput({
         small={small}
         regular={regular}
       />
+
+      {isEndDateFocused && children}
 
       {showClearDates && (
         <button
--- a/test/components/DateRangePicker_spec.jsx
+++ b/test/components/DateRangePicker_spec.jsx
@@ -9,6 +9,7 @@ import DateRangePicker, { PureDateRangePicker } from '../../src/components/DateR
 
 import DateRangePickerInputController from '../../src/components/DateRangePickerInputController';
 import DayPickerRangeController from '../../src/components/DayPickerRangeController';
+import DayPicker from '../../src/components/DayPicker';
 
 import {
   HORIZONTAL_ORIENTATION,
@@ -16,6 +17,49 @@ import {
 } from '../../src/constants';
 
 const describeIfWindow = typeof document === 'undefined' ? describe.skip : describe;
+
+class DateRangePickerWrapper extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      focusedInput: null,
+      startDate: null,
+      endDate: null,
+    };
+
+    this.onDatesChange = this.onDatesChange.bind(this);
+    this.onFocusChange = this.onFocusChange.bind(this);
+  }
+
+  onDatesChange({ startDate, endDate }) {
+    this.setState({ startDate, endDate });
+  }
+
+  onFocusChange(focusedInput) {
+    this.setState({ focusedInput });
+  }
+
+  render() {
+    const { focusedInput, startDate, endDate } = this.state;
+
+    return (
+      <div>
+        <DateRangePicker
+          {...this.props}
+          onDatesChange={this.onDatesChange}
+          onFocusChange={this.onFocusChange}
+          focusedInput={focusedInput}
+          startDate={startDate}
+          endDate={endDate}
+        />
+        <button type="button">
+          Dummy button
+        </button>
+      </div>
+    );
+  }
+}
 
 const requiredProps = {
   onDatesChange: () => {},
@@ -536,6 +580,23 @@ describe('DateRangePicker', () => {
         wrapper.instance().onDayPickerFocus();
         expect(onFocusChangeStub.getCall(0).args[0]).to.equal(START_DATE);
       });
+    });
+  });
+
+  describeIfWindow('day picker position', () => {
+    it('day picker is opened after the end date input when end date input is focused', () => {
+      const wrapper = mount((
+        <DateRangePickerWrapper
+          startDateId="startDate"
+          endDateId="endDate"
+        />
+      ));
+      expect(wrapper.find(DayPicker)).to.have.length(0);
+      wrapper.find('input').at(0).simulate('focus'); // when focusing on start date the day picker is rendered after the start date input
+      expect(wrapper.find('DateRangePickerInput').children().childAt(1).find(DayPicker)).to.have.length(1);
+      wrapper.find('input').at(1).simulate('focus'); // when focusing on end date the day picker is rendered after the end date input
+      expect(wrapper.find('DateRangePickerInput').children().childAt(1).find(DayPicker)).to.have.length(0);
+      expect(wrapper.find('DateRangePickerInput').children().childAt(3).find(DayPicker)).to.have.length(1);
     });
   });
 


majapw (approved):
Seems simple enough to me!

greglo:
@ljharb It seems like you've had some luck contributing recently. Do you have any idea why CI doesn't appear to be running for this PR?

.gitignore
```diff
@@ -36,6 +36,7 @@ node_modules
 lib
 esm
 .idea
+*.iml
```
ljharb, line 39: i'm not sure what this is, but generally things like this (non-project-specific files) go in your global gitignore file rather than every project you touch.

src/components/DateRangePicker.jsx
```diff
@@ -391,7 +391,7 @@ class DateRangePicker extends React.PureComponent {
 
     if (withPortal || withFullScreenPortal || appendToBody) {
       return (
-        <Portal>
+        <Portal key="day-picker-portal">
```
ljharb, line 394: this isn't rendered in an array, so the key has no effect. why is this here?

src/components/DateRangePicker.jsx
```diff
@@ -481,6 +481,7 @@ class DateRangePicker extends React.PureComponent {
     /* eslint-disable jsx-a11y/click-events-have-key-events */
     return (
       <div
+        key="day-picker"
```
ljharb, line 484: similarly, this is never rendered into an array, so the key here has no effect. why is this here?

ljharb (changes requested):
Is there any kind of regression test we could add for the DateRangePickerInput changes?

I've rebased this; that should fix CI.

src/components/DateRangePicker.jsx
greglo: Keys can let us control how React create new instances of components. This is most commonly used for preventing unnecessary component mount/unmounts when rendering arrays of components, but is not _only_ for arrays. 

Ken Dodds has a clear example of why a key might be useful in non-array components to force a new instance [here](https://kentcdodds.com/blog/understanding-reacts-key-prop).

In this case I am doing the precise opposite: by specifying a key I give React the information it needs to not create a new instance even though I am doing something which means it normally would (move a React element to a new position in the tree). If you check out my branch and play around with it in storybook with and without either `key` prop you'll see that it flickers without the `key` prop on reverse-tabbing (caused by un/re-mount). If it isn't apparent you can add logging to the existing `ref` callback and see the absence of a `key` prop causes additional DOM changes.

src/components/DateRangePicker.jsx
greglo: See below

src/components/DateRangePicker.jsx
ljharb: This is a single hardcoded string though - what happens if I have multiple DateRangePicker instances on the page?

greglo: There's no issue here across multiple instances. Keys only have to be unique across siblings within a given element but not globally either across elements or instances of those elements. They key is applied to an element which

Just to be sure I tested this with multiple instances on the same page and they behave completely independently as expected. Do you have a specific concern for unwanted behavior that we can check for?

Here is what I am trying to achieve with the key:
```tsx
// Before tab
<DateInput ... />
<DatePicker ... />
<DateInput ... />

// After tab
<DateInput ... />
<DateInput ... />
<DatePicker ... />
```

Without a key on `DatePicker` React is going to create a new instance which causes a flicker. With the key React will not create a new instance and there won't be a flicker (aside: during forward navigation the blur handler will still cause a flicker, which is a separate issue. Back-tab-navigation is the better demonstration here).

greglo: The element with a key is within the implementation of this component. So while it would be possible for a future editor of this library to encounter this if they rendered two drop-downs (albeit unlikely since a single DateRangePicker shouldn't want multiple drop-downs) it is not possible for a user of the library.

exviva:
@ljharb if you have no further objections, could we get this approved and merged? It's a blocker for one of our Accessibility tickets.

ljharb:
@exviva @greglo i'm still waiting for an answer on "Is there any kind of regression test we could add for the DateRangePickerInput changes?"

danielmag:
> @exviva @greglo i'm still waiting for an answer on "Is there any kind of regression test we could add for the DateRangePickerInput changes?"

@ljharb I was trying to add a regression test but didn't find any good way to test this using enzyme. Should we merge as is?

danielmag:
@ljharb What do you think?

mikepspc:
Also waiting on this.

ljharb:
Nothing should ever be merged without a regression test - if there's nothing to prevent it breaking in the future, then that means nobody will care if it's broken later, which means nobody cares it's broken now.

enzyme has both shallow and mount tests; if it's actually a technical impossibility to test this absent a full browser then a storybook story will suffice, but I'm not convinced that's the case.

danielmag:
@ljharb It seems most of the tests are broken when I try running them. Any idea why this could happen? I'm thinking it might be some library upgrade breaking the tests since we don't have a package.lock file

danielmag:
> @ljharb It seems most of the tests are broken when I try running them. Any idea why this could happen? I'm thinking it might be some library upgrade breaking the tests since we don't have a package.lock file

I found the culprit. It's version 4.2.0 of `react-with-styles`. Locking in ~4.1.0 solves the issue.

danielmag:
@ljharb Greg is no longer working on this issue. How should I contribute here without opening a new PR?
In the meanwhile I have tried adding a regression test but without success so far. The tab onKeyDown event does not seem to be working as expected. See the code here:

`test/components/DateRangePicker_spec.jsx`:
```javascript
...

class DateRangePickerWrapper extends React.Component {
  constructor(props) {
    super(props);

    this.state = {
      focusedInput: null,
      startDate: null,
      endDate: null,
    };

    this.onDatesChange = this.onDatesChange.bind(this);
  }

    this.setState({ startDate, endDate });
  }

  onFocusChange(focusedInput) {
    this.setState({ focusedInput });
  }

  render() {
    const { focusedInput, startDate, endDate } = this.state;

    return (
      <div>
        <DateRangePicker
          {...this.props}
          onDatesChange={this.onDatesChange}
          focusedInput={focusedInput}
          startDate={startDate}
          endDate={endDate}
        />
        <button type="button">
          Dummy button
        </button>
      </div>
    );
  }
}

...

    it('tabbing out with keyboard will close the day picker', () => {
      const wrapper = mount((
        <DateRangePickerWrapper
          startDateId="startDate"
          endDateId="endDate"
        />
      ));
      expect(wrapper.find(DayPicker)).to.have.length(0);
      wrapper.find('input').at(0).simulate('focus');
      expect(wrapper.find(DayPicker)).to.have.length(1);
      // Tab through 10 times to eventually go through all tabbable elements and
      // close the picker
      wrapper.simulate('keydown', { key: 'Tab', keyCode: 9, which: 9 });
      wrapper.simulate('keydown', { key: 'Tab', keyCode: 9, which: 9 });
      wrapper.simulate('keydown', { key: 'Tab', keyCode: 9, which: 9 });
      wrapper.simulate('keydown', { key: 'Tab', keyCode: 9, which: 9 });
      wrapper.simulate('keydown', { key: 'Tab', keyCode: 9, which: 9 });
      wrapper.simulate('keydown', { key: 'Tab', keyCode: 9, which: 9 });
      wrapper.simulate('keydown', { key: 'Tab', keyCode: 9, which: 9 });
      wrapper.simulate('keydown', { key: 'Tab', keyCode: 9, which: 9 });
      wrapper.simulate('keydown', { key: 'Tab', keyCode: 9, which: 9 });
      wrapper.simulate('keydown', { key: 'Tab', keyCode: 9, which: 9 });
      expect(wrapper.find(DayPicker)).to.have.length(0);
    });
```

danielmag:
@ljharb Would really appreciate your input here. Also, if we cannot manage to test this with enzyme, how would you suggest testing it with a storybook story since we need to tab through the calendar picker to see the issue.

juliaevk:
> @ljharb here is the branch with my changes: https://github.com/danielmag/react-dates/tree/dm--fix-keyboard-navigation. Let me know if there is a better way to test this with enzyme. I'm not sure how we could test this using storybook.

I've encountered this same issue, and this PR would resolve it nicely for me. +1 on this!

danielmag:
@ljharb Did get a chance to take a look at my branch? Also contains the fix for the multiple failing tests

danielmag:
@ljharb I found another way to test this and tests are now passing. Let me know if it's good to merge

ljharb:
@danielmag @greglo @sharmilajesupaul this looks great, but there's still no tests or storybook story in the PR. Is this bug evident in existing stories?

danielmag:
Hey @ljharb, thanks for the reply. I added the test in my branch here: https://github.com/danielmag/react-dates/tree/dm--fix-keyboard-navigation

ljharb (approved):

ljharb:
@danielmag what was the breaking change in a minor version of `react-with-styles`? Has an issue been filed on that repo?

danielmag:
> @danielmag what was the breaking change in a minor version of `react-with-styles`? Has an issue been filed on that repo?

I don't remember the issue but the newer version was breaking our tests.

ljharb:
Great catch of course, but whatever the breakage is needs to be fixed in `react-with-styles` :-) would you mind filing an issue there about it?

danielmag:
Sure 👍

ljharb:
Some of the tests are failing because nvm isn't up to date on travis, so `nvm install-latest-npm` isn't working properly. I'll ping travis folks about it.

ljharb:
Looks like this will in fact have to wait a bit longer; the branch protections need to be updated in #2160 first.

PR #2093, opened 2021-02-17
Update storybook docs with default prop openDirection

KevenMax:
The openDirection property is not currently set as the default property in the storybook examples.

This PR sends the addition of that property.


ljharb (approved):
Thanks!

PR #2095, opened 2021-02-22
[Fix] `DayPicker`: update calendarMonthWeeks if numberOfMonths is changed

### LINKED ISSUE #1394: adjustDayPickerHeight not calculating predictably when setting numberOfMonths prop dynamically

haojonat:
Fixes #1394.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ script:
   - 'if [ -n "${KARMA-}" ]; then npm run tests-karma ; fi'
   - 'if [ -n "${COVERAGE-}" ] && [ "${TRAVIS_BRANCH-}" = "master" ]; then npm run cover; cat ./coverage/lcov.info | ./node_modules/.bin/coveralls ; fi'
 install:
-  - 'if [ -n "${LINT-}" ]; then npm install --legacy-bundling ; else npm install ; fi'
+  - 'if [ -n "${LINT-}" ]; then NPM_CONFIG_LEGACY_PEER_DEPS=true npm install --legacy-bundling ; else npm install ; fi'
 env:
   global:
     - TEST=true
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "lint": "eslint --ext .js,.jsx src test",
     "mocha": "mocha ./test/_helpers",
     "storybook:uninstall": "npm uninstall --no-save @storybook/react && rimraf node_modules/@storybook node_modules/react-modal node_modules/react-dom-factories",
-    "react": "enzyme-adapter-react-install 16",
+    "react": "NPM_CONFIG_LEGACY_PEER_DEPS=true enzyme-adapter-react-install 16",
     "pretest": "npm run --silent lint",
     "pretests-only": "npm run react",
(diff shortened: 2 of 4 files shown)


haojonat:
This is a rebase of this pr:

https://github.com/airbnb/react-dates/pull/1395

haojonat:
@LMulvey @ljharb - any ideas on why the builds are failing?

ljharb:
@haojonat there's no need to close and reopen a PR to "kick" the tests; if they don't pass immediately, and should, then there's a problem with the repo and it's not something that can be fixed by a rerun.

The tests are failing right now due to the release of npm 7.

haojonat:
@ljharb ah gotcha, in that case, should i wait for the tests to get fixed?  or are we comfortable enough to get this overridden and merged?

ljharb:
Please do wait; I’ll rebase this PR once master is passing again.

haojonat:
Sounds good, thanks @ljharb

ljharb (approved):

ljharb:
It looks like the new tests aren't actually covering the new lines, so the coverage requirement isn't being met.

haojonat:
@ljharb, i think i might have force pushed over one of your changes to run the LEGACY_PEER_DEPENDENCIES or something.  My bad.  Could you repush that?

haojonat:
@ljharb - noticed a case where the bug persists.  will push another commit to address that as well as take another look at the tests

ljharb:
@haojonat sounds great, thanks.

ljharb:
Why do the storybook knobs need to be removed?

haojonat:
@ljharb - it was causing the REACT=15 builds to fail for some reason.  /shrug

ljharb:
found another way to fix that.

haocow:
hi @ljharb, I just wanted to see if you were planning on updating the 21.8.0 release commit or creating a new 21.8.1 release based on latest?

ljharb:
Releases can't ever be updated; they're immutable. This PR alone would constitute a patch release, but I'm not sure when the next release will be, nor what its version number.

haocow:
ah, darn, okay

abenitesDC:
Hi @ljharb, do you know what does releasing a patch (or minor) depends on?

PR #2112, opened 2021-05-04
Add name proptype, close #2110

### LINKED ISSUE #2110: Pass props like required, name, placeholder, pattern to the input element
I think we should be able to pass those props down to the input element with

```jsx
      <SingleDatePicker
        onDateChange={/*...*/}
        date={/*...*/}
        focused={/*...*/}
        onFocusChange={/*...*/}
        {...props} // contains required, name, ... other input props
      />
```

else it's necessary to use a ref and a useEffect, and set attributes manually. It's really not practical

Can I open a PR for that?

caub:


ljharb (changes requested):
Just adding the propType wouldn’t do anything - the component would have to use it, and we’d need tests and a storybook story.

caub:
`name` is already passed into the component, I can see date in input's attributes, it's just that proptypes was out of sync, and was warning about name as extra prop

ljharb:
`SingleDatePickerInput` never references a `name` prop, so all this PR currently achieves is removes the _correct_ warning that `name` is an extra, unused prop.

In order to remove the warning, we'd have to actually *use the prop* inside `SingleDatePickerInput`.

caub:
Oh correct, sorry, wrong component, fixed

ljharb:
and which component is using the `name` prop but lacks the propType? https://github.com/airbnb/react-dates/blob/6c2cb61c5e876fbdac72e6efed49c0bd64f06d6b/src/components/SingleDatePicker.jsx, for example, uses the shape but doesn't use the `name` prop.

caub:
yes SingleDatePicker, the one component I use in https://codesandbox.io/s/react-datepickers-o9fbh?file=/src/Datepicker.js

you can see the warning: 
```
Warning: Failed prop type: withStyles(SingleDatePicker): unknown props found: name, pattern
    at withStyles(SingleDatePicker)
```

And you can also see that date attribute is passed to the input, so I guess that prop is spread to other components

```html
<input class="css-1fqugce" aria-label="YYYY-MM-DD" type="text" id="date" name="date" placeholder="YYYY-MM-DD" autocomplete="off" required="" aria-describedby="DateInput__screen-reader-message-date" value="2021-05-05">
```

ljharb:
It's not, in fact - the name is always the same as the ID: https://github.com/airbnb/react-dates/blob/master/src/components/DateInput.jsx#L229-L230

caub:
@ljharb Oh thanks, good to know. I'm not really a fan of those 'opinionated' choices, but I'll do with it. It'd be better I think to allow to pass independently name and id

Let's deal with this choice for now, thanks

PR #2128, opened 2021-06-18
Replaced moment library with date-fns

Najd-Feki:
.


ljharb (changes requested):
Temporal is stage 3, and the instant it has a production-ready polyfill, date-fns along with every other date library will be obsolete. It's not worth migrating to date-fns (especially a prerelease version) when it's about to be dead.

ljharb:
@LeandroPS it's stage 3 already, it's just that changes are still being made.

Either way, it would be worth it to stay on moment for years more, rather than incur the cost of *two* migrations.

moment is sufficiently performant; that's not a compelling argument.

PR #2146, opened 2021-09-19
2140 time at noon and cache

### LINKED ISSUE #2140: Why does isOutsideRange get called multiple times for each date, and at different times (midnight vs noon)?
**react-dates version**
react-dates@21.8.0

**Describe the bug**
isOutsideRange prop callback is passing each calendar day multiple times for some reason

specifically, each date can be called at local time midnight or local time noon, which makes it hard for me to know which one I'm supposed to compare for to return a correct value to isOutSideRange

Load and see strange output
```
<DayPickerRangeController
  isOutsideRange={(date) => console.log('huh', date.format())}
/>
```

image
image

https://user-images.githubusercontent.com/5566843/128084014-da550974-cb35-4e27-9d29-a19a6013ba90.mp4

 - OS: mac
 - Browser chrome
 - Version 92.0.4515.107 (Official Build) (x86_64)

glintik:
Two commits that fix #2140:
1) Set time to noon(12:00) for cloned Moment objects with startOf() calls in DayPickerRangeController. 
2) Added cache for isDayBlocked and isOutsideRange.


src/components/DayPickerRangeController.jsx
```diff
@@ -213,11 +213,12 @@ export default class DayPickerRangeController extends React.PureComponent {
 
     this.isTouchDevice = isTouchDevice();
     this.today = moment();
+    this.cachedProps = {};
```
ljharb, line 216: This will break use cases; it’s not guaranteed that these callbacks’ return values are memoizable.

ljharb (changes requested):

ljharb:
For the first commit, can we add regression tests?

src/components/DayPickerRangeController.jsx
glintik: I think, these callbacks return values can be memoizable during render time at least, but it's up to you to decide.

glintik:
> For the first commit, can we add regression tests?

Yes, I can add some. You mean, to check time in changed methods or something more sophisticated?

src/components/DayPickerRangeController.jsx
ljharb: During a single render, that seems reasonable, but also I’m not sure what the value is of adding this optimization.

ljharb:
Something that replicates the bug, and that your changes fix, without directly testing midnight vs noon.

src/components/DayPickerRangeController.jsx
glintik: Value is performance improvement. 3x less calls of isDayBlocked() and isOutsideRange(). If user has DayPickerRangeController with both callbacks used, then 90(3 months)*2(methods)*3(not cached) = 540 calls performed. After the improvement with return values cache, calls count decreases to 180.

ljharb: hmm. this seems like a bit more complexity than should be necessary for per-render caching. could we split that out into a separate PR, and keep this one just for the bug fix and regression tests?

glintik: Moved cache commit to PR #2147

glintik:
@ljharb, please take a look again. I reverted cache changes and added some tests.
P.S. I wasn't able to run tests from DayPicker_spec.jsx, looks like tests are broken?
Screenshot 2021-09-21 at 14 29 10

test/components/DayPickerSingleDateController_spec.jsx
```diff
@@ -974,6 +974,19 @@ describe('DayPickerSingleDateController', () => {
...
+      const wrapper = shallow((
+        <DayPickerSingleDateController
+          date={null}
+          onDateChange={sinon.stub()}
+          onFocusChange={sinon.stub()}
+        />
+      ));
+      const firstFocusableDay = wrapper.instance().getFirstFocusableDay(today);
+      expect(firstFocusableDay.hours()).to.equal(12);
+    });
```
ljharb, line 988: this seems like it's testing the implementation, rather than the bug reported in #2140.

glintik: @ljharb, not sure I follow what you mean.
From bug:
> expected behavior: returns consistent time, either always midnight local time or always noon local time
Test ensures that time is a noon. Bug closed.

ljharb: right, but the bug is "it's called twice and the time isn't the same" - so shouldn't the test call it twice and guarantee the time is the same?

(it's fine to *also* test it's at noon, because there's other reasons it should be noon)

glintik: "called twice" resolved in another commit, I sent you another PR with cache solution to resolve it. This PR fixes time issue only.

ljharb: The issue was that *when* it's called multiple times, even across renders, the *time* is different. I don't think the caching would address that.

glintik: @ljharb As I said, no caching involved in my solution in current PR. I reverted it. The only changes - to set up hours to 12:00.

ljharb (changes requested):
Thanks! I've rebased this.

There's still changes in `utils/getNumberOfCalendarMonthWeeks` and `utils/isDayVisible` and `components/DayPicker` that don't have associated tests, though.

glintik:
> Thanks! I've rebased this.
> 
> There's still changes in `utils/getNumberOfCalendarMonthWeeks` and `utils/isDayVisible` and `components/DayPicker` that don't have associated tests, though.

hi @ljharb,

May you refer to an example, what test needed for these changes?

src/utils/getNumberOfCalendarMonthWeeks.js
```diff
@@ -9,7 +9,7 @@ export default function getNumberOfCalendarMonthWeeks(
   month,
   firstDayOfWeek = moment.localeData().firstDayOfWeek(),
 ) {
-  const firstDayOfMonth = month.clone().startOf('month');
+  const firstDayOfMonth = month.clone().startOf('month').hour(12);
```
ljharb, line 12: i assume that `getNumberOfCalendarMonthWeeks` has a bug in some timezones if the hour is set to 0 - can we have a unit test that triggers that bug, such that this line's change fixes it?

src/utils/isDayVisible.js
```diff
@@ -21,7 +21,7 @@ export default function isDayVisible(day, month, numberOfMonths, enableOutsideDa
 
   if (enableOutsideDays) {
     if (!startCacheOutsideDays.has(startKey)) {
-      startCacheOutsideDays.set(startKey, month.clone().startOf('month').startOf('week'));
+      startCacheOutsideDays.set(startKey, month.clone().startOf('month').startOf('week').hour(12));
```
ljharb, line 24: same question here

src/components/DayPicker.jsx
```diff
@@ -695,7 +695,7 @@ class DayPicker extends React.PureComponent {
     }
 
     if (newMonth && (!focusedDate || !isDayVisible(focusedDate, newMonth, numberOfMonths))) {
-      focusedDate = newMonth.clone().startOf('month');
+      focusedDate = newMonth.clone().startOf('month').hour(12);
```
ljharb, line 698: similarly, something that unit tests or exercises getFocusedDay, and shows the bug that this change fixes, would cover here

src/utils/getNumberOfCalendarMonthWeeks.js
glintik: To be honest, getNumberOfCalendarMonthWeeks() is not affected by the issue, I added "hour(12)" as unified approach to get firstDayOfMonth. Probably it's a good idea to have special util methods like :
```js
const INITIAL_HOUR = 12;

getFirstDayOfMonth = (month) => month.clone().startOf('month').hour(INITIAL_HOUR);
getFirstDayOfFirstWeekOfMonth = (month) => getFirstDayOfMonth(month).startOf('week').hour(INITIAL_HOUR);
```

src/components/DayPicker.jsx
glintik: Bug is not related to timezone, it's related to initial time for newly created date objects because in some code lines hours are set while in others are not. So I'm in stuck what test is needed here.

src/utils/isDayVisible.js
glintik: Same as for ```getNumberOfCalendarMonthWeeks```, just to honor unified approach to get first day of week/month.

src/utils/getNumberOfCalendarMonthWeeks.js
ljharb: Is there no timezone/date combination where the first day of the month, or the first day of the first week of a month, doesn't have a midnight?

src/components/DayPicker.jsx
ljharb: If this change isn't fixing anything, maybe it should be reverted?

glintik: > If this change isn't fixing anything, maybe it should be reverted?

Sure, but there is a good reason to have unified approach everywhere: it minimizes possible FUTURE bugs. All developers will see how to get first day of month and in all future cases will get a date with proper initial time.

Only "hour(12)" in components really matter for the issue.

src/utils/getNumberOfCalendarMonthWeeks.js
glintik: Midnight is 00:00, but I set hour(12) that is a noon to mitigate summer/winter time changes, so noon is a good choice for date calculations. Or you mean something other, sorry?

ljharb: Right - I’m envisioning a bug caused by midnight that noon fixes.

glintik: I did quick research and didn't find any other issues related to midnight, but react-dates codebase is large and no warranty here without full code audit.

ljharb: What I mean is, 2013.10.20 was the start of the week in Brazil where midnight doesn't exist, but it's not the start of the month or the first day of the week when the week ends in a different month. Maybe there's another date in a specific timezone where either of those is true?

ljharb:
I've played with a few of the dates here: https://stackoverflow.com/a/61328380/632724 and i can't come up with any test cases where the hour matters. I'll merge this as-is.

ljharb (approved):

src/utils/getNumberOfCalendarMonthWeeks.js
glintik: > where midnight doesn't exist
I don't see that lack of midnight causes any error:
```js
const moment = require('moment-timezone');
moment.tz("2013-10-20 00:00:00", "Brazil/East").toString();
```
Output:
```Sun Oct 20 2013 01:00:00 GMT-0200```
Moment handles this flawlessly as I can see(+1 hour added).

Thank you for merging!

PR #2147, opened 2021-09-21
Add cache for isDayBlocked and isOutsideRange in DayPickerRangeContro…

glintik:
…ller.

Proposed solution to cache modifiers result from #2140 to improve rendering performance.
@ljharb please take a look.

### DIFF
--- a/src/components/DayPickerRangeController.jsx
+++ b/src/components/DayPickerRangeController.jsx
@@ -213,11 +213,12 @@ export default class DayPickerRangeController extends React.PureComponent {
 
     this.isTouchDevice = isTouchDevice();
     this.today = moment();
+    this.cachedProps = {};
     this.modifiers = {
       today: (day) => this.isToday(day),
       blocked: (day) => this.isBlocked(day),
-      'blocked-calendar': (day) => props.isDayBlocked(day),
-      'blocked-out-of-range': (day) => props.isOutsideRange(day),
+      'blocked-calendar': (day) => this.getCachedPropByDay(day, 'isDayBlocked'),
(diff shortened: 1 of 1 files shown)


src/components/DayPickerRangeController.jsx
```diff
@@ -1111,6 +1110,42 @@ export default class DayPickerRangeController extends React.PureComponent {
     return { currentMonth, visibleDays };
   }
 
+  getCachedPropByDay(day, prop, props, forceUpdate = false) {
+    const key = day.toString();
```
ljharb, line 1114:
```suggestion
    const key = String(day);
```

src/components/DayPickerRangeController.jsx
```diff
@@ -1111,6 +1110,42 @@ export default class DayPickerRangeController extends React.PureComponent {
     return { currentMonth, visibleDays };
   }
 
+  getCachedPropByDay(day, prop, props, forceUpdate = false) {
```
ljharb, line 1113: is there a reason this is cached on the instance, instead of being done in react state?

src/components/DayPickerRangeController.jsx
```diff
@@ -1111,6 +1110,42 @@ export default class DayPickerRangeController extends React.PureComponent {
...
+    return value;
+  }
+
+  removeCachedProps(day, prop) {
+    if (!day) {
+      // Remove all
+      this.cachedProps = {};
+    }
+
+    const key = day.toString();
```
ljharb, line 1137:
```suggestion
    const key = String(day);
```

ljharb (changes requested):
Are there any tests that could illustrate how this helps?

PR #2178, opened 2022-01-25
Upgrade storybook version to the newest

hugobessa:
Upgrades the storybook version and its configurations.

I also took the liberty of fixing some extra commas and missing quotes in some configuration files (unrelated to storybook)


.babelrc
```diff
@@ -1,59 +1,59 @@
...
-      "presets": [["airbnb", { looseClasses: true }]],
+      "presets": [["airbnb", { "looseClasses": true }]],
       "plugins": [
         ["inline-react-svg", {
           "svgo": false
         }],
         ["transform-replace-object-assign", { "moduleSpecifier": "object.assign" }],
         "./scripts/pure-component-fallback.js",
-        "istanbul",
+        "istanbul"
```
ljharb, line 11: babelrc (and eslintrc) allow trailing commas, so they're preferred to reduce git diff churn.

.storybook-css/.eslintrc
```diff
@@ -3,6 +3,6 @@
     "import/no-extraneous-dependencies": [2, {
       "devDependencies": true
     }],
-    "global-require": 2,
```
ljharb, line 6: same here

.storybook/.eslintrc
```diff
@@ -3,6 +3,6 @@
     "import/no-extraneous-dependencies": [2, {
       "devDependencies": true
     }],
-    "global-require": 2,
+    "global-require": 2
```
ljharb, line 6: and here

ljharb (commented):
Thanks, overall this looks great! I'll test out the storybook locally before landing, as well.

.babelrc
hugobessa: will adjust then. My linter was complaining about these

ljharb: The only linter that should be ran in a project is that project’s, and ours doesn’t :-)

PR #2206, opened 2022-06-27
[Fix] DateInput: Address accessibility issues

henten20:


ljharb:
Why did you duplicate this with #2207?

LauRoxx:
This looks to be an accidental PR that was only meant to be a PR on our forked repo (see https://github.com/Availity/react-dates/pull/13 ). Should be good to close again.

ljharb:
@LauRoxx thanks for following up. is there a reason the changes shouldn't be upstreamed?

LauRoxx:
The last commit wasn't a correct solution to the aria-controls issue and was not accepted on our forked repo.

PR #2212, opened 2022-07-25
Aria control add

mtramantT:
Added 'aria-control' and 'id' attributes to DatePicker class


ljharb:
Was this intended for your company’s fork?

PR #2215, opened 2022-08-22
fix code for react 18

golee:


ljharb:
It's definitely not this simple, unfortunately.

PR #2219, opened 2022-10-10
AX-351 update moment

APshenkin:


samuxyz:
@APshenkin Hey, out of curiosity, any reason why this was closed? Moment > 2.29.2 has some important security patches.

ljharb:
@samuxyz we depend on moment using `^`, so there's no reason to update react-dates - just install the latest moment.

PR #2229, opened 2023-02-17
[New] Add 'fixed' option to allow position-fixed appendToBody pickers

hsource:
## Motivation

We are trying to use the date-picker in a small `position: fixed` container with `overflow: hidden`. To make it escape its container using a portal, we use `appendToBody={true}`

This mostly works, but because the container is appended to the body in a static layout, the date picker becomes invisible once we scroll down the page.

## Fix

1. Add `appendToBody="fixed"` as an additional option, in addition to the normal booleans. Update all PropTypes

## Testing

Added new automated Mocha tests that verify that `position: 'fixed'` gets added when `appendToBody="fixed"` is set


hsource:
@ljharb Is there any chance you can review/merge this in? I'd appreciate it!

Issue #2204, opened 2022-06-20
Error: Cannot find module 'node-sass'


bzappasodi:
I'm running:
node -v
v17.4.0
npm -v
8.3.1

I'm Seeing the following:
ERROR in ./css/storybook.scss (./node_modules/raw-loader!./node_modules/sass-loader/dist/cjs.js!./css/storybook.scss)
Module build failed (from ./node_modules/sass-loader/dist/cjs.js):
Error: Cannot find module 'node-sass'

razlani:
> I'm running: node -v v17.4.0 npm -v 8.3.1
> 
> I'm Seeing the following: ERROR in ./css/storybook.scss (./node_modules/raw-loader!./node_modules/sass-loader/dist/cjs.js!./css/storybook.scss) Module build failed (from ./node_modules/sass-loader/dist/cjs.js): Error: Cannot find module 'node-sass'

Seconded, trying to contribute via a feature request and who knew the first boss would be making it compile/run storybook.

-Did you figure out all the steps for windows dev?

razlani:
After some laborious setup via installing visual studio with msbuild tools which is the dependency to node-gyp (among other deps), I'm getting an error as described in the log file below, upon attemping an `npm install node-sass@4.0.0` 

Time to just build components from scratch I reckon!
[error.txt](https://github.com/react-dates/react-dates/files/9985138/error.txt)

Pedro-Porfirio:
Same problem, it seems that node-sass is not included in the initial package.json. I read that node-sass is deprecated, not sure exactly what is happening here.

onlywei:
Same problem here as well. Cannot start storybook.

ghost:
node-sass is deprecated ( more [info ](https://github.com/sass/node-sass#node-sass) ), thus we need to use sass package. it's a seamless installation no changes are needed.

PR: https://github.com/react-dates/react-dates/pull/2231

PR #2231, opened 2023-04-18
build: add sass package for storybook

ghost:
sass package is needed to run storybook. while it's not listed in package.json we need to install it manually. this commit adds sass package in package.json

#2204

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,7 +129,8 @@
     "react-portal": "^4.2.1",
     "react-with-direction": "^1.4.0",
     "react-with-styles": "~4.1.0",
-    "react-with-styles-interface-css": "^6.0.0"
+    "react-with-styles-interface-css": "^6.0.0",
+    "sass": "^1.62.0"
   },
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -129,7 +129,8 @@
     "react-portal": "^4.2.1",
     "react-with-direction": "^1.4.0",
     "react-with-styles": "~4.1.0",
-    "react-with-styles-interface-css": "^6.0.0"
+    "react-with-styles-interface-css": "^6.0.0",
+    "sass": "^1.62.0"
```
ledenis, line 133: Shouldn't it be in `devDependencies`? Consumers of `react-dates` don't need `sass`, do they?

ljharb (changes requested):
please move sass to dev deps.

also, does `sass` require compilation? it'd be ideal to only depend on pure JS deps.

ljharb:
I guess @lukaonik deleted their account.

PR #2244, opened 2023-11-07
Accept React 17.x

florentdestremau:
Hi, seing this repo has no recent activity, we are still using it daily with React 17 with no issues, but I'm trying to remove our warnings.


ljharb (changes requested):
The repo is still maintained, and adding support for a react version we can’t test on is unacceptable.

florentdestremau:
hi @ljharb 

The last commit is almost 2 years old, I don't see a CONTRIBUTING file either, and many of the "recent" PRs are being closed with a single comment like this. On NPM the [last package version](https://www.npmjs.com/package/react-dates?activeTab=versions) is 4 years old.

I'm very thankful for what react-dates it has brought to the open source, but React 16 development ended 3 years ago and React 17 was released about the same date (and is also not actively maintained either way), so if it's unacceptable to support it, this package definitely feels abandoned to me.

I see no docs on testing, and I am not allowed to run the tests on the CI anyway. If you point me in the right direction I might try to add React 17 to the testing workflow and see how it goes :slightly_smiling_face: As I said, we've been using it in production for a while with React 17.x and encounter no issue.

ljharb:
Recency is irrelevant; this package is maintained.

we use enzyme for tests; enzyme doesn’t support react > 16 yet, so we’re blocked from doing so. You’re welcome to take a look at enzyme and see if you can get the open PR for react 17 support to have passing tests.

florentdestremau:
I've seen the discussions on the enzyme repository about React 17+ compat, I understand that sadly this isn't going anywhere. Thank you for your time and dedication to be responsive on these subjects nevertheless!

May I suggest indicating in the README of this repo that React 17 & up won't be supported unless https://github.com/enzymejs/enzyme/issues/2429 is merged AND tests are updated here to verify the compatibility? Again, my only complaint here is that _de facto_ this works great on my React 17 project, I just want to avoid the `npm i -f` all the time because of the peerDependecy error.

ljharb:
Tests won’t need to be updated here beyond just adding the version to peer deps, once enzyme supports it.

I’m very happy to hear it works; that suggests tests will Just Work at that future time.

